PR #8, opened 2019-03-16
Add logs to SQL Rest Handler for better debugging.

jinsoor-amzn:
*Description of changes:*
Following try catch will cause NPE without any logs. This commit helps to better debug NPE issue until we add more meaningful exception back to customer.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
@@ -15,6 +15,8 @@
 
 package com.amazon.opendistroforelasticsearch.sql.plugin;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.elasticsearch.client.node.NodeClient;
 import org.elasticsearch.common.settings.Settings;
 import com.amazon.opendistroforelasticsearch.sql.executor.ActionRequestRestExecutorFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -33,6 +35,7 @@
 import java.util.Set;
 
 public class RestSqlAction extends BaseRestHandler {
+    public static Logger logger = LogManager.getLogger(RestSqlAction.class);
```
qreshi: Might be better to declare this Logger ```private static final```

jinsoor-amzn: Oops. Good catch. Updated.

dai-chen: Thanks for the change! Mohammad was referring to the logger convention like this: https://github.com/opendistro-for-elasticsearch/sql/blob/c126e456538167283df92f49850aa93e0307e69a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java#L44

I will try to fix the NPE issue this week.

jinsoor-amzn:
Please squash and merge when approved.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
jinsoor-amzn: I have updated with package convention in third commit.

Issue #9, opened 2019-03-20
Problems for selecting on index patterns in JDBC


HernanMora:
When I try the plugin on REST it works correctly

```
curl -H "Content-Type: application/json" -XPOST http://localhost:9200/_opendistro/_sql -d '{"query":"select * from my-index* limit 1","parameters":[]}'
```
But when I perform the same test through the JDBC connector I get the following error:

```
curl -H "Content-Type: application/json" -XPOST http://localhost:9200/_opendistro/_sql?format=jdbc -d '{"query":"select * from my-index* limit 1","parameters":[]}'
{
  "error": {
    "reason": "Index type null does not exist",
    "details": "java.lang.IllegalArgumentException: Index type null does not exist\n\tat com.amazon.opendistroforelasticsearch.sql.executor.format.SelectResultSet.loadFromEsState(SelectResultSet.java:109)\n\tat com.amazon.opendistroforelasticsearch.sql.executor.format.SelectResultSet.<init>(SelectResultSet.java:71)\n\tat com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.loadResultSet(Protocol.java:61)\n\tat com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.<init>(Protocol.java:48)\n\tat com.amazon.opendistroforelasticsearch.sql.executor.format.PrettyFormatRestExecutor.execute(PrettyFormatRestExecutor.java:65)\n\tat com.amazon.opendistroforelasticsearch.sql.executor.format.PrettyFormatRestExecutor.execute(PrettyFormatRestExecutor.java:42)\n\tat com.amazon.opendistroforelasticsearch.sql.executor.AsyncRestExecutor.lambda$async$1(AsyncRestExecutor.java:104)\n\tat org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:660)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)\n\tat java.base/java.lang.Thread.run(Thread.java:834)\n",
    "type": "java.lang.IllegalArgumentException"
  },
  "status": 500
}
```
It would be very useful to support index patterns

Regards!

dai-chen:
Thanks for reporting this issue. We're looking into it.

[...]

PR #11, opened 2019-03-21
Fix index pattern issue for JDBC format

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/9

*Description of changes:* Use the first one in index mapping rather than look up by index name/pattern which caused the mismatch. This works because all mappings found are exactly the same and there is validation for this assumption in rewriter.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/PrettyFormatResponseTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/PrettyFormatResponseTest.java
@@ -98,7 +98,7 @@ public class PrettyFormatResponseTest {
     public void indexPattern() {
         String query = "SELECT * FROM " + TestsConstants.TEST_INDEX + "_loc*";
         Protocol protocol = CheckPrettyFormatContents.execute(query, "jdbc");
-        assertThat(4L, equalTo(protocol.getResultSet().getDataRows().getSize()));
+        assertThat(protocol.getResultSet().getDataRows().getSize(), equalTo(4L));
     }
 
     /** Test Protocol as a whole */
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/PrettyFormatResponseTest.java
```diff
@@ -90,10 +90,17 @@ public void wrongIndexType() {
...
-            assertThat(e.getMessage(), is(String.format("Index type %s does not exist", type)));
+            assertThat(e.getMessage(), is(String.format("Index type [%s/%s] does not exist", TestsConstants.TEST_INDEX_ACCOUNT, type)));
         }
     }
 
+    @Test
+    public void indexPattern() {
+        String query = "SELECT * FROM " + TestsConstants.TEST_INDEX + "_loc*";
+        Protocol protocol = CheckPrettyFormatContents.execute(query, "jdbc");
+        assertThat(4L, equalTo(protocol.getResultSet().getDataRows().getSize()));
```
qreshi: NP: I believe for ```assertThat()``` the format goes ```assertThat(T actual, Matcher<> matcher)``` so this assertion would be ```assertThat(protocol.getResultSet().getDataRows().getSize(), equalTo(4L))```. This way it almost reads like a sentence, "Assert that dataRows size is equal to 4".

dai-chen: > NP: I believe for `assertThat()` the format goes `assertThat(T actual, Matcher<> matcher)` so this assertion would be `assertThat(protocol.getResultSet().getDataRows().getSize(), equalTo(4L))`. This way it almost reads like a sentence, "Assert that dataRows size is equal to 4".

Thanks! Fixing it. I think I confused the assertion of Hamcrest and JUnit. :)

Issue #10, opened 2019-03-20
NPE thrown for illegal sql query


dai-chen:
In some case, NullPointerException or full stack trace will be returned which is not friendly to user. Need to improve error handling code to return clear error message in uniform format.

PR #13, opened 2019-03-25
Better error handling

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/10

*Description of changes:* Return useful error message rather than null to ES. Hide detailed stack trace for security concern. Also skip rewriter in the case of FROM not present.

Sample output: 

```
{
  "error": {
    "reason": "no such index",
    "details": "[account11] IndexNotFoundException[no such index]",
    "type": "IndexNotFoundException"
  },
  "status": 400
}
```


src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -56,43 +67,30 @@ public String getName() {
...
                     if (request.hasParam(paramName)) {
                         additionalParams.put(paramName, request.param(paramName));
                     }
                 }
-                return channel -> restExecutor.execute(client, additionalParams, finalQueryAction, channel);
+                return channel -> restExecutor.execute(client, additionalParams, queryAction, channel);
             }
-        } catch (SqlParseException | SQLFeatureNotSupportedException e) {
+        } catch (Exception e) {
             // FIXME: need to catch all exceptions to avoid ES process from crashing
```
qreshi: This ```FIXME``` can be removed if it's addressed by the change to ```catch (Exception e)``` above.

dai-chen: Yes, removing it. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -56,43 +67,29 @@ public String getName() {
...
-            queryAction = searchDao.explain(sqlRequest.getSql());
+            final SqlRequest sqlRequest = SqlRequestFactory.getSqlRequest(request);
+            final QueryAction queryAction = new SearchDao(client).explain(sqlRequest.getSql());
             queryAction.setSqlRequest(sqlRequest);
 
-            // TODO add unittests to explain. (rest level?)
             if (request.path().endsWith("/_explain")) {
                 final String jsonExplanation = queryAction.explain().explain();
-                return channel -> channel.sendResponse(new BytesRestResponse(RestStatus.OK, jsonExplanation));
+                return channel -> channel.sendResponse(new BytesRestResponse(OK, jsonExplanation));
```
jngz-es:
```suggestion
                return RestChannelConsumer sendResponse(jsonExplanation, OK);
```

dai-chen: > Suggested change 
>                 return channel -> channel.sendResponse(new BytesRestResponse(OK, jsonExplanation));
>                 return RestChannelConsumer sendResponse(jsonExplanation, OK);

Thanks! Addressing it.

Issue #19, opened 2019-03-28
Add Useful Logging for Troubleshoot


dai-chen:
Currently there are very few logging in the plugin, except those new added in query plan executor. Adding more useful and general logging for troubleshoot.

PR #33, opened 2019-04-17
Add useful logging

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/19

*Description of changes:* The following changes are being introduced in this PR:

 1. Generate unique ID for each request for tracking
 2. Prepend request ID to the logging along the path
 3. Extra log for slow query with configurable threshold (add generic code for using ES cluster settings)
 4. Return only general message in details in error message for security concern.

All unit test and legacy/new integration tests passed. Checked added log in elasticsearch.log and did some manually testing in Kibana, primarily for slow log threshold:

```
POST _opendistro/_sql
{
  "query": "SELECT * FROM accounts WHERE age = 30"
}

PUT /_cluster/settings
{
  "transient": {
    "opendistro.sql.query.slowlog": "0"
  }
}

PUT /_cluster/settings
{
  "transient": {
    "opendistro.sql.query.slowlog": null
  }
}

PUT /_cluster/settings
{
  "persistent": {
    "opendistro.sql.query.slowlog": "0"
  }
}
```

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
@@ -71,7 +71,7 @@ public class LocalClusterState {
     /** Current cluster state on local node */
     private ClusterService clusterService;
 
-    /** Sql specific setting in ES cluster setting */
+    /** Sql specific settings in ES cluster settings */
     private SqlSettings sqlSettings;
 
     /** Index name expression resolver to get concrete index name */
@@ -83,7 +83,7 @@ public class LocalClusterState {
      */
     private final Cache<Tuple<List<String>, List<String>>, IndexMappings> cache;
 
-    /** Latest setting value for each registered key. Thread-safe is required here as well. */
+    /** Latest setting value for each registered key. Thread-safe is required. */
     private final Map<String, Object> settingMap = new ConcurrentHashMap<>();
 
 
@@ -120,7 +120,7 @@ public class LocalClusterState {
                 setting,
                 newVal -> {
                     if (LOG.isDebugEnabled()) {
-                        LOG.debug("Setting [{}] changed to [{}]", setting.getKey(), newVal);
+                        LOG.debug("Setting value of [{}] changed to [{}]", setting.getKey(), newVal);
                     }
                     settingMap.put(setting.getKey(), newVal);
                 });
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
@@ -84,7 +84,7 @@ public class LocalClusterState {
     private final Cache<Tuple<List<String>, List<String>>, IndexMappings> cache;
 
     /** Latest setting value for each registered key. Thread-safe is required. */
-    private final Map<String, Object> settingMap = new ConcurrentHashMap<>();
+    private final Map<String, Object> latestSettings = new ConcurrentHashMap<>();
 
 
     public static synchronized LocalClusterState state() {
@@ -120,9 +120,9 @@ public class LocalClusterState {
                 setting,
                 newVal -> {
                     if (LOG.isDebugEnabled()) {
-                        LOG.debug("Setting value of [{}] changed to [{}]", setting.getKey(), newVal);
+                        LOG.debug("The value of setting [{}] changed to [{}]", setting.getKey(), newVal);
                     }
-                    settingMap.put(setting.getKey(), newVal);
+                    latestSettings.put(setting.getKey(), newVal);
                 });
         }
     }
@@ -144,7 +144,7 @@ public class LocalClusterState {
     @SuppressWarnings("unchecked")
     public <T> T getSettingValue(String key) {
         Objects.requireNonNull(sqlSettings, "SQL setting is null");
-        return (T) settingMap.getOrDefault(key, sqlSettings.getSetting(key).getDefault(EMPTY));
+        return (T) latestSettings.getOrDefault(key, sqlSettings.getSetting(key).getDefault(EMPTY));
     }
 
     /** Get field mappings by index expressions. All types and fields are included in response. */
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistroforelasticsearch.sql.esdomain;
 
+import com.amazon.opendistroforelasticsearch.sql.plugin.SqlSettings;
 import com.carrotsearch.hppc.cursors.ObjectObjectCursor;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
@@ -29,6 +30,7 @@ import org.elasticsearch.cluster.metadata.MetaData;
 import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.collect.ImmutableOpenMap;
 import org.elasticsearch.common.collect.Tuple;
+import org.elasticsearch.common.settings.Setting;
 import org.elasticsearch.index.IndexNotFoundException;
 import org.json.JSONObject;
 
@@ -38,11 +40,13 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static java.util.Collections.emptyMap;
+import static org.elasticsearch.common.settings.Settings.EMPTY;
 
 /**
  * Local cluster state information which may be stale but help avoid blocking operation in NIO thread.
@@ -67,6 +71,9 @@ public class LocalClusterState {
     /** Current cluster state on local node */
     private ClusterService clusterService;
 
+    /** Sql specific setting in ES cluster setting */
+    private SqlSettings sqlSettings;
+
     /** Index name expression resolver to get concrete index name */
     private IndexNameExpressionResolver resolver;
 
@@ -75,6 +82,9 @@ public class LocalClusterState {
      * Array cannot be used as key because hashCode() always return reference address, so either use wrapper or List.
      */
     private final Cache<Tuple<List<String>, List<String>>, IndexMappings> cache;
+
+    /** Latest setting value for each registered key. Thread-safe is required here as well. */
+    private final Map<String, Object> settingMap = new ConcurrentHashMap<>();
 
 
     public static synchronized LocalClusterState state() {
@@ -103,12 +113,38 @@ public class LocalClusterState {
         });
     }
 
+    public void setSqlSettings(SqlSettings sqlSettings) {
+        this.sqlSettings = sqlSettings;
+        for (Setting<?> setting : sqlSettings.getSettings()) {
+            clusterService.getClusterSettings().addSettingsUpdateConsumer(
+                setting,
+                newVal -> {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Setting [{}] changed to [{}]", setting.getKey(), newVal);
+                    }
+                    settingMap.put(setting.getKey(), newVal);
+                });
+        }
+    }
+
     public void setResolver(IndexNameExpressionResolver resolver) {
         this.resolver = resolver;
     }
 
     private LocalClusterState() {
         cache = CacheBuilder.newBuilder().maximumSize(100).build();
+    }
+
+    /**
+     * Get setting value by key. Return default value if not configured explicitly.
+     *
+     * @param key   setting key registered during plugin launch.
+     * @return      setting value or default
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T getSettingValue(String key) {
+        Objects.requireNonNull(sqlSettings, "SQL setting is null");
+        return (T) settingMap.getOrDefault(key, sqlSettings.getSetting(key).getDefault(EMPTY));
     }
 
     /** Get field mappings by index expressions. All types and fields are included in response. */
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
@@ -135,7 +135,7 @@ public class AsyncRestExecutor implements RestExecutor {
     }
 
     private String requestId(QueryAction action) {
-        return action.getSqlRequest() == null ? "Unknown" : action.getSqlRequest().getId();
+        return action.getSqlRequest().getId();
     }
 
     private String name(Object object) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
@@ -53,11 +53,11 @@ public class AsyncRestExecutor implements RestExecutor {
     private final Predicate<QueryAction> isBlocking;
 
 
-    public AsyncRestExecutor(RestExecutor executor) {
+    AsyncRestExecutor(RestExecutor executor) {
         this(executor, ALL_ACTION_IS_BLOCKING);
     }
 
-    public AsyncRestExecutor(RestExecutor executor, Predicate<QueryAction> isBlocking) {
+    AsyncRestExecutor(RestExecutor executor, Predicate<QueryAction> isBlocking) {
         this.executor = executor;
         this.isBlocking = isBlocking;
     }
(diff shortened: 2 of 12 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
```diff
@@ -103,6 +113,20 @@ public void setClusterService(ClusterService clusterService) {
...
     }
 
+    public void setSqlSettings(SqlSettings sqlSettings) {
+        this.sqlSettings = sqlSettings;
+        for (Setting<?> setting : sqlSettings.getSettings()) {
+            clusterService.getClusterSettings().addSettingsUpdateConsumer(
+                setting,
+                newVal -> {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Setting value of [{}] changed to [{}]", setting.getKey(), newVal);
```
arsen-es: `"The value of setting [{}] changed to [{}]"`

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
```diff
@@ -76,6 +83,9 @@
      */
     private final Cache<Tuple<List<String>, List<String>>, IndexMappings> cache;
 
+    /** Latest setting value for each registered key. Thread-safe is required. */
+    private final Map<String, Object> settingMap = new ConcurrentHashMap<>();
```
arsen-es: I prefer not including type information in the identifier, I think it's better to call this `latestSettings`

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
dai-chen: Done!

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
```diff
@@ -66,17 +71,17 @@
     public void execute(Client client, Map<String, String> params, QueryAction queryAction, RestChannel channel) throws Exception {
         if (isBlockingAction(queryAction) && isRunningInTransportThread()) {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Async blocking query action [{}] for executor [{}] in current thread [{}]",
-                    name(executor), name(queryAction), Thread.currentThread().getName());
+                LOG.debug("[{}] Async blocking query action [{}] for executor [{}] in current thread [{}]",
```
arsen-es: I think this approach of pre-pending request id to log messages does not scale + is error prone. So far I've seen this type of change in log4j configuration itself, which includes the request id in all log messages for a given appender and then you don't have to worry about it in the code.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
```diff
@@ -97,24 +102,56 @@ private boolean isRunningInTransportThread() {
...
+            channel.sendResponse(new BytesRestResponse(SERVICE_UNAVAILABLE, new ErrorMessage(e, SERVICE_UNAVAILABLE.getStatus()).toString()));
+        }
+    }
+
+    /** Time the real execution of Executor and log slow query for troubleshooting */
+    private void doExecuteWithTimeMeasured(Client client,
+                                           Map<String, String> params,
+                                           QueryAction action,
+                                           RestChannel channel) throws Exception {
+        Instant startTime = Instant.now();
```
arsen-es: Instant.now() can have pretty big errors depending on time drift/skew/last time of sync of system time/etc. Please use `System.nanoTime()` for anything where you need to measure a period and not get a specific point in time (https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--).

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: > I think this approach of pre-pending request id to log messages does not scale + is error prone. So far I've seen this type of change in log4j configuration itself, which includes the request id in all log messages for a given appender and then you don't have to worry about it in the code.

I'm not sure how log4j could be aware of the relation between logging entries. Because here basically we want to track all logs related to a request by an unique ID. Could you give me a sample?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
```diff
@@ -66,17 +71,17 @@
...
         }
         else {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Continue running query action [{}] for executor [{}] in current thread [{}]",
-                    name(executor), name(queryAction), Thread.currentThread().getName());
+                LOG.debug("[{}] Continue running query action [{}] for executor [{}] in current thread [{}]",
+                    requestId(queryAction), name(executor), name(queryAction), Thread.currentThread().getName());
             }
-            executor.execute(client, params, queryAction, channel);
+            sync(client, params, queryAction, channel);
```
arsen-es: the name `sync` is a little misleading, I read it as *synchronize* where I assume you meant *executeSynchronously*.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/ErrorMessage.java
```diff
@@ -37,10 +38,16 @@ public ErrorMessage(Exception exception, int status) {
 
     private String fetchType() { return exception.getClass().getSimpleName(); }
 
-    private String fetchReason() { return emptyStringIfNull(exception.getLocalizedMessage()); }
+    private String fetchReason() {
+        return status == RestStatus.BAD_REQUEST.getStatus()
+                ? "SQL in the query request is not valid"
```
arsen-es: np: `"Invalid SQL query"`

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
dai-chen: Done!

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/ErrorMessage.java
arsen-es: How about other client-side errors? Should we at least print out a generic message for everything else which is 4xx and then another one for 5xx?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -67,8 +67,11 @@ public String getName() {
 
     @Override
     protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+        SqlRequest sqlRequest = SqlRequest.NULL;
```
arsen-es: seems this line is redundant now.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: > Instant.now() can have pretty big errors depending on time drift/skew/last time of sync of system time/etc. Please use `System.nanoTime()` for anything where you need to measure a period and not get a specific point in time (https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--).

Thanks for the suggestion! I wasn't aware of this. Could you pass me some reference for the problem you mentioned? Because we only care about the delta value returned from `Duration.between`, is it possible to be wrong too in some case?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -67,8 +67,11 @@ public String getName() {
...
         try {
-            final SqlRequest sqlRequest = SqlRequestFactory.getSqlRequest(request);
+            sqlRequest = SqlRequestFactory.getSqlRequest(request);
```
arsen-es: why not `final` anymore?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: Yeah, probably, I use it to be consistent with the method `async`. So do you think it's better to change both to their full name `executeSynchronously` and `executeAsynchronously`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/ErrorMessage.java
dai-chen: Done!

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
dai-chen: > seems this line is redundant now.

Why?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlPlug.java
```diff
@@ -47,6 +48,9 @@
 
 public class SqlPlug extends Plugin implements ActionPlugin {
```
arsen-es: seems there is inconsistent indentation in this file

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
dai-chen: > why not `final` anymore?

To make it `final`, it can be set only once (in `try` block) which is a compiling error because it's being read in `catch` block too.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlPlug.java
dai-chen: > seems there is inconsistent indentation in this file

Yes, it's a long pending issue...

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
dai-chen: I see. Please read my comments below.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
```diff
@@ -29,17 +29,29 @@
...
 import java.io.IOException;
 import java.util.Collections;
+import java.util.UUID;
 
 public class SqlRequest {
 
+    public static final SqlRequest NULL = new SqlRequest("Unassigned", "", null);
+
+    /** Unique request ID for tracking */
+    private final String id;
```
arsen-es: I am thinking if it would be more useful to keep the id attached to the RestRequest context, not sqlRequest. That would make sure we have end-to-end (from the point of entering the plugin) tracking of requests. This is just a food for thought, no immediate action required.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
```diff
@@ -97,24 +102,56 @@ private boolean isRunningInTransportThread() {
...
+            },
+            new TimeValue(0L),
+            SQL_WORKER_THREAD_POOL_NAME);
+    }
+
+    private void sync(Client client, Map<String, String> params, QueryAction queryAction, RestChannel channel) {
+        try {
+            doExecuteWithTimeMeasured(client, params, queryAction, channel);
+        }
+        catch (Exception e) {
```
jngz-es: Do we need to remove this catch to throw exception outside? There is reportError outside to differentiate clientError and serverError.

dai-chen: Make sense. I think I was confused and thought this runs outside that try-catch.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
arsen-es: it's a little too long, but I think it's still better to have more explicit/descriptive naming, so yes.

dai-chen: As Jing's comments, will remove this wrapper method. Resolving this conversation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
arsen-es: I couldn't find a good example right now, but I remember I did that previously with slf4j (using log4j implementation); in a nutshell it has some logging context in the thread context, and you just write the requestId into that as soon as you get the request, and the logger knows how to pull values from that context, you don't need to specify the value yourself in each log call. This will be a bigger change, so let's create an issue to track it and we can decide how to fix it later. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: Done! Will publish new revision soon.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: Sure. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
dai-chen: Sure. Not sure how to do this for now. Maybe we can incorporate it into your log4j proposal too. Resolving the conversation..

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlPlug.java
dai-chen: Will fix this later to avoid big PR with unrelated changes. AFAIK, we do have a few files with this problem.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
arsen-es: This one touches upon the problem a little bit: https://www.baeldung.com/java-measure-elapsed-time. Most of the sources say that resolution of nanoTime is better, but that's not the issue. The real difference is that `System.currentTimeMillis()` and `Instant.now()` (which relies on the former) both use wall-clock time from the JVM, which in its turn gets it from OS. Depending on what protocol the machine is using to synchronize the time, it can lag behind or go far ahead and then during sync get a huge change (either negative or positive). The `System.nanoTime()` is based on software-based counter (not related to wall clock), which only increases, and does not need to be synced with any other machine/server/clock, so the diffs of values from nanoTime will always be pretty accurate (I guess as long as you are using it on the same thread).

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
arsen-es: ok, assuming the `SqlRequestFactory.getSqlRequest` can throw, I see the problem now.

dai-chen: Thanks! Resolving this if it makes sense.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: > This one touches upon the problem a little bit: https://www.baeldung.com/java-measure-elapsed-time. Most of the sources say that resolution of nanoTime is better, but that's not the issue. The real difference is that `System.currentTimeMillis()` and `Instant.now()` (which relies on the former) both use wall-clock time from the JVM, which in its turn gets it from OS. Depending on what protocol the machine is using to synchronize the time, it can lag behind or go far ahead and then during sync get a huge change (either negative or positive). The `System.nanoTime()` is based on software-based counter (not related to wall clock), which only increases, and does not need to be synced with any other machine/server/clock, so the diffs of values from nanoTime will always be pretty accurate (I guess as long as you are using it on the same thread).

Thanks! Good to know the subtle difference. I think `Instant` and `Duration` is sufficient in our case. Because we are not benchmarking, we just need a rough elapsed time in second(s). `nanoTime()` may be more accurate in some edge case, though it's hassle to convert it to sec or millisec for logging or comparison. Hopefully this makes sense to you.

arsen-es: It's just two lines that need to be changed:
#140 should be `final long startNanos = System.nanoTime();`
#145 should be `final Duration elapsed = Duration.ofNanos(System.nanoTime() - startNanos);

It's up to you if you want to keep it this way, not a blocker.

dai-chen: > It's just two lines that need to be changed:
> #140 should be `final long startNanos = System.nanoTime();`
> #145 should be `final Duration elapsed = Duration.ofNanos(System.nanoTime() - startNanos);
> 
> It's up to you if you want to keep it this way, not a blocker.

Easier than I thought. Testing it. Will push very soon. Thanks!

dai-chen:
> Please create the issue for updating the request id logging approach and include in the response to the relevant comment before merging and closing this PR.

Done!

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/AsyncRestExecutor.java
dai-chen: Created issue for tracking: https://github.com/opendistro-for-elasticsearch/sql/issues/34

PR #36, opened 2019-04-21
Deb build

allenyin96:
Created buildPackages task in build.gradle that runs all the package builds per suggestion here https://github.com/opendistro-for-elasticsearch/opendistro-infra/pull/27#discussion_r277120968


allenyin96:
Tested by running 
```
./gradlew buildPackages --console=plain -Dbuild.snapshot=false
```
in the project root directory and verifying that the correct artifacts (jar, zip, deb & rpm) are built.

Issue #23, opened 2019-04-03
Add Node Level Metrics API


jngz-es:
Add RESTful API to collect metrics, such as memory circuit breaker etc.

PR #39, opened 2019-04-24
Metrics in node level

jngz-es:
*Issue #23 , if available:*

*Description of changes:*
Add metrics in SQL plugin.
Currently ignore metrics integration test case, because
1) ES integration testing framework doesn't support singleton/static using.
2) nodeSettings can't add parameters into the initial settings of ES cluster.


src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Counter.java
```diff
@@ -0,0 +1,27 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+public interface Counter<T> {
+
+    void increment();
+
+    void increment(long n);
```
arsen-es: Should we stick to java's naming convention (see `LongAdder::add(long)`) and call this method `add` instead?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
```diff
@@ -0,0 +1,37 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import java.util.function.Supplier;
+
+public class GaugeMetric<T> extends Metric<T> {
+
+    private Supplier<T> LOAD_VALUE;
```
arsen-es: why is the field name in all caps?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
```diff
@@ -0,0 +1,37 @@
...
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import java.util.function.Supplier;
+
+public class GaugeMetric<T> extends Metric<T> {
```
arsen-es: can you add documentation for all classes (at least non-trivial ones) to make the purpose of the class clear.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
```diff
@@ -0,0 +1,55 @@
...
+    REQ_TOTAL("request_total", 0),
+    REQ_COUNT_TOTAL("request_count", 1),
+    FAILED_REQ_COUNT_SYS("failed_request_count_syserr", 1),
+    FAILED_REQ_COUNT_CUS("failed_request_count_cuserr", 1),
+    FAILED_REQ_COUNT_CB("failed_request_count_cb", 1),
+    CIRCUIT_BREAKER("circuit_breaker", 2),
+    DEFAULT("default", 0);
+
+    private String name;
+    private int type;
```
arsen-es: np: `typeId` ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
```diff
@@ -0,0 +1,55 @@
...
+    public int getType() {
+        return type;
+    }
+
+    public static List<String> getNames() {
+        return Arrays.stream(MetricType.values()).map(v -> v.name).collect(Collectors.toList());
+    }
+
+    public boolean isNumerical() {
+        return type < 2;
```
arsen-es: This looks really neat, but I think it's also not very extensible. First, it requires maintaining the order of the values, and second, if one adds a new value, s/he might have to update this number as well, so it's extra maintenance, plus implicit requirement. I suggest we explicitly whitelist the types here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
```diff
@@ -0,0 +1,33 @@
...
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import com.amazon.opendistroforelasticsearch.sql.query.join.BackOffRetryStrategy;
+
+public class MetricFactory {
+
+    public static Metric createMetric(MetricType type) {
+        if (type.getType() == 1) {
+            return new NumericMetric<>(type.getName(), new RollingCounter());
+        } else if (type == MetricType.CIRCUIT_BREAKER) {
```
arsen-es: it's a little confusing to have different cases based on different values - some for `type.getType()` and others for the `type` itself. It might introduce overlaps in the future, and generally it's hard to follow, please either use `type` everywhere, or `getType`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
```diff
@@ -0,0 +1,55 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public enum MetricType {
```
arsen-es: so far my impression is you combined the name of the metric and its type in a this enum, which is contradictory to its name. And one consequence of that choice is having the second type field in it, which in my opinion *is* the real type. Another bad side effect is when you instantiate metrics, you get the name of the metric from the type. So I suggest replacing the integer values (for `type` field) with actual enum values, in other words, getting rid of the type field, and making this enum less granular. Have a different utility class where you have constant names for metrics as well as the mapping between the name and the type (many to one is fine).

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import com.amazon.opendistroforelasticsearch.sql.query.join.BackOffRetryStrategy;
+
+public class MetricFactory {
+
+    public static Metric createMetric(MetricType type) {
```
arsen-es: if you agree to my overall suggestions about refactoring the MetricType class, this method will then accept two parameters: `metricName` and `metricType`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
arsen-es: Another bad side effect of adding that subtype (`type` field) : no one will understand what those numbers mean, and when one adds a new type with the existing approach s/he won't know which number to assign to it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+            registerMetric(MetricFactory.createMetric(metricType));
+        }
+    }
+
+    public void registerMetric(Metric metric) {
+        metricMap.put(metric.getName(), metric);
+    }
+
+    public void unRegisterMetric(String name) {
+        if (name == null || !metricMap.containsKey(name)) {
```
arsen-es: `Map::remove(key)` checks if the key is present, so this check is redundant here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+            return;
+        }
+
+        metricMap.remove(name);
+    }
+
+    public Metric getMetric(String name) {
+        if (name == null || !metricMap.containsKey(name)) {
```
arsen-es: Same here, if the key is not present `Map::get` will return `null`, so no need to do the check here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/RollingCounter.java
```diff
@@ -0,0 +1,102 @@
...
+public class RollingCounter implements Counter<Long> {
+
+    private final long capacity;
+    private final long window;
+    private final long interval;
+    private final Clock clock;
+    private final ConcurrentSkipListMap<Long, Long> time2CountWin;
+    private final LongAdder count;
+
+    public RollingCounter() {
```
arsen-es: please avoid duplicating code:
```java
public RollingCounter() {

  final long window = LocalClusterState.state().getSettingValue(SqlSettings.METRICS_ROLLING_WINDOW);
  final long interval = LocalClusterState.state().getSettingValue(SqlSettings.METRICS_ROLLING_INTERVAL);

  this(window, interval);
}
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/RollingCounter.java
```diff
@@ -0,0 +1,102 @@
...
+        this.interval = interval;
+        clock = Clock.systemDefaultZone();
+        time2CountWin = new ConcurrentSkipListMap<>();
+        count = new LongAdder();
+        capacity = window / interval * 2;
+    }
+
+    @Override
+    public void increment() {
+        trim();
```
arsen-es: you can just call `increment(1L)`

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -91,6 +95,11 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                 return channel -> restExecutor.execute(client, additionalParams, queryAction, channel);
             }
         } catch (Exception e) {
```
arsen-es: ideally this should be
```java
catch (final NullPointerException | SqlParseException | ParserException | SQLFeatureNotSupportedException | IllegalArgumentException | IndexNotFoundException e) {
  // do whatever
} catch (final Exception e) {
  // handle generic case
}
```

but even if you want to keep it this way please update to this:
```java
Metrics.getInstance().getNumericMetric(isClientError(e) ? MetricType.FAILED_REQ_COUNT_CUS : MetricType.FAILED_REQ_COUNT_SYS).increment();
```

arsen-es: Also, I know it's not your code, but can you fix `isClientError` method to use `||` instead of `|` .

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class Metrics {
```
abbashus: Shouldn't this be a singleton?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlStatsAction.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.elasticsearch.rest.RestStatus.BAD_REQUEST;
+import static org.elasticsearch.rest.RestStatus.SERVICE_UNAVAILABLE;
+
+public class RestSqlStatsAction extends BaseRestHandler {
```
arsen-es: please do not duplicate code (e.g. methods `responseParams`, `isClientError`, etc.), if needed create another base class with all the shared code, or move the shared code into a util class and include it through containment, anyway you want, but have the common methods in one place only.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/join/BackOffRetryStrategy.java
```diff
@@ -51,6 +54,8 @@
 
     private final static Object obj = new Object();
 
+    public final static Supplier<Integer> GET_CB_STATE = () -> isMemoryHealthy() ? 0 : 1;
```
arsen-es: what does **CB** stand for?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/join/BackOffRetryStrategy.java
```diff
@@ -74,12 +79,17 @@ public static boolean isHealthy() {
...
             if (ThreadLocalRandom.current().nextBoolean()) {
+                Metrics.getInstance().getNumericMetric(MetricType.FAILED_REQ_COUNT_CB).increment();
                 LOG.warn("[MCB1] Directly abort on idx {}.", i);
                 return false;
             }
             backOffSleep(intervals[i]);
         }
-        return isMemoryHealthy();
+
+        boolean health = isMemoryHealthy();
```
arsen-es: `boolean isHealthy`

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public Metric getMetric(String name) {
+        if (name == null || !metricMap.containsKey(name)) {
+            return null;
+        }
+
+        return metricMap.get(name);
+    }
+
+    public NumericMetric getNumericMetric(MetricType metricType) {
```
arsen-es: np: `getNumericalMetric`

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Counter.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+public interface Counter<T> {
```
arsen-es: Why is this class generic? What types other than `long` do you expect to have in the future? It would be a little odd to have a method `increment` for non-integral types, and all integral types can fit under `long`. Do I miss something?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/NumericMetric.java
```diff
@@ -0,0 +1,51 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+public class NumericMetric<T> extends Metric<T> {
+
+    private Counter<T> counter;
+
+    public NumericMetric(String name, Counter counter) {
```
arsen-es: I think it's not a good idea to "outsource" a concrete type's behavior. What's the use case where we would need to change how we count? And the second comment, even if you want to support this, it should be an override to the default, in other words there should be a constructor which accepts just the name and uses `BasicCounter` as the `Counter` implementation.

arsen-es: Ok, never mind, I guess this is fine. However, I am thinking if it would be better to derive the Counter implementation from the `MetricType`, I got the impression that those are somehow connected. Anyway, this is an `NP` comment, just out of my curiousity.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
dai-chen: > Also, I know it's not your code, but can you fix `isClientError` method to use `||` instead of `|` .

Good catch. That was my mistake. :)

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Counter.java
jngz-es: Sure, will change it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
jngz-es: I consider it as a function, so use all caps.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
jngz-es: Sure, will add some comments.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
jngz-es: I am fine with either.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
jngz-es: Good suggestion, will update.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
jngz-es: OK, will do it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/RollingCounter.java
arsen-es: I think this won't work, because the constructor call has to be the first, which means you can inline the window and interval calculations.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
arsen-es: Do we have any other code where we use all caps for function names? From what I've seen and used so far, only the names of constants should be in all caps.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/RollingCounter.java
jngz-es: Good suggestions, will do it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricType.java
jngz-es: Good suggestion, will do it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
jngz-es: Sure, will do it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class Metrics {
+
+    private static Metrics metrics = new Metrics();
+    private ConcurrentHashMap<String, Metric> metricMap = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
jngz-es: OK, will change it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
jngz-es: OK, will keep only null check.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
jngz-es: OK, will keep only null check.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/RollingCounter.java
jngz-es: Good suggestion!

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/RollingCounter.java
jngz-es: Good suggestion!

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
jngz-es: Good catch! Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlStatsAction.java
jngz-es: Will do it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/join/BackOffRetryStrategy.java
jngz-es: circuit breaker

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/join/BackOffRetryStrategy.java
jngz-es: Will change it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
jngz-es: OK, will change it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Counter.java
jngz-es: Counter only support integer/long increment, but the value could be float etc., for example, error rate.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/NumericMetric.java
jngz-es: Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
jngz-es: Not sure. I can change it to loadValue, is that OK for you?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Counter.java
arsen-es: It seems counter-intuitive to me to have a `float` counter, firstly because I wouldn't know what the increment for `float` is. Seems like it should always be `long`, and then the users can always convert it to whatever numerical type.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/GaugeMetric.java
arsen-es: That is fine, thank you.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
```diff
@@ -0,0 +1,39 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.metrics;
+
+import com.amazon.opendistroforelasticsearch.sql.query.join.BackOffRetryStrategy;
+
+public class MetricFactory {
+
+    public static Metric createMetric(MetricName name) {
```
arsen-es: Should I be able to create more than one `Metric` instance for each `MetricName` ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+        for (MetricName metricName : MetricName.values()) {
+            registerMetric(MetricFactory.createMetric(metricName));
+        }
+    }
+
+    public void registerMetric(Metric metric) {
+        registeredMetricsByName.put(metric.getName(), metric);
+    }
+
+    public void unRegisterMetric(String name) {
```
arsen-es: np: `unregisterMetric`

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/join/BackOffRetryStrategy.java
```diff
@@ -74,12 +79,17 @@ public static boolean isHealthy() {
...
+                Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_CB).increment();
                 LOG.warn("[MCB1] Directly abort on idx {}.", i);
                 return false;
             }
             backOffSleep(intervals[i]);
         }
-        return isMemoryHealthy();
+
+        boolean isHealthy = isMemoryHealthy();
+        if (!isHealthy) Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_CB).increment();
```
arsen-es: np: please use proper if block `{ }`

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricName.java
```diff
@@ -0,0 +1,51 @@
...
+
+public enum MetricName {
+
+    REQ_TOTAL("request_total"),
+    REQ_COUNT_TOTAL("request_count"),
+    FAILED_REQ_COUNT_SYS("failed_request_count_syserr"),
+    FAILED_REQ_COUNT_CUS("failed_request_count_cuserr"),
+    FAILED_REQ_COUNT_CB("failed_request_count_cb"),
+    CIRCUIT_BREAKER("circuit_breaker"),
+    DEFAULT("default");
```
arsen-es: what is this one for?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
```diff
@@ -0,0 +1,88 @@
...
+    private static Metrics metrics = new Metrics();
+    private ConcurrentHashMap<String, Metric> registeredMetricsByName = new ConcurrentHashMap<>();
+
+    public static Metrics getInstance() {
+        return metrics;
+    }
+
+    private Metrics() {}
+
+    public void registerDefaultMetrics() {
```
arsen-es: how are default metrics different from non-default? If I want to add my own metric, should I add it into the set of default or not?

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
jngz-es: No, Metric is singleton for SQL plugin.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Counter.java
jngz-es: Not increment for float. I means float for getValue. Anyway, we can make more discussion about it in next release. Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
jngz-es: Ah, nice, will change it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/join/BackOffRetryStrategy.java
jngz-es: OK, will change it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricName.java
jngz-es: FAILED_REQ_COUNT_CUS is for counting request failures by customer error, like SQL syntax error etc.
FAILED_REQ_COUNT_CB is for counting request failures by circuit breaker.
CIRCUIT_BREAKER is current status of circuit breaker, i.e. circuit is broken or not.
DEFAULT is a default metric, could be used in some scenarios in the future.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/Metrics.java
jngz-es: Either is OK, I think. You could add it in this default if you think this metric is should be initialized from beginning. I not, you could choose when and where to register/unregister your own.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
arsen-es: `Metrics` is, but `Metric` and its subclasses are not.

jngz-es: Oh, I got what you mean. Metric is not singleton, but we only hold one metric per MetricName.

abbashus:
Since the stats will be for a particular node and not whole cluster, we should also add the node-name/node-id to the response. Ex:

```
{
  "node_name" :  "somenode" ,
  "request_count" : 600,
  "failed_request_count_syserr" : 0,
  "failed_request_count_cuserr" : 0,
  "circuit_breaker" : 0 
}
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MetricsIT.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.concurrent.TimeUnit;
+
+import static com.amazon.opendistroforelasticsearch.sql.intgtest.TestsConstants.TEST_INDEX_DOG;
```
qreshi: NP: Since we're using ```esintgtest.TestConstants``` which is in the same directory, this static import isn't necessary. The IDE probably chose the wrong one and added the import.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
arsen-es: for now yes, but you don't restrict it. I think we should not allow registering metrics with the same name, moreover, not even creating if there is already a registered metric.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MetricsIT.java
jngz-es: Make sense, thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricName.java
arsen-es: I think it's better to get rid of default, especially that you have a default case in your `MetricFactory::createMetric` anyway.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricFactory.java
jngz-es: We could override the metric, just one metric for one name is OK.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlStatsAction.java
```diff
@@ -0,0 +1,72 @@
...
+import org.elasticsearch.rest.RestRequest;
+import org.elasticsearch.rest.RestStatus;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.elasticsearch.rest.RestStatus.SERVICE_UNAVAILABLE;
+
+public class RestSqlStatsAction extends BaseRestHandler {
```
arsen-es: Let's explicitly document that this API will return the stats for the node that received the request, and that the second change (coming soon) will add support to get cluster level stats.

src/main/java/com/amazon/opendistroforelasticsearch/sql/metrics/MetricName.java
jngz-es: I think these two default are two different concepts. One is for user to use intentionally another one is for some unexpected scenarios.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlStatsAction.java
jngz-es: Sure, will add it.

Issue #47, opened 2019-05-10
Support for Elasticsearch 7.x


penghuo:
Enable support for Elasticsearch 7.0.1

PR #48, opened 2019-05-10
Support for Elasticsearch 7.0.1 (#47)

penghuo:
*Issue #, if available: #47 

*Description of changes:*
This PR enable support for Elasticsearch 7.0.1

- Breaking changes for [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html)


src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/SearchResult.java
```diff
@@ -49,7 +50,7 @@
 
 	public SearchResult(SearchResponse resp) {
 		SearchHits hits = resp.getHits();
-		this.total = hits.getTotalHits();
+		this.total = Optional.ofNullable(hits.getTotalHits()).map(totalHits -> totalHits.value).orElse(0L);
```
qreshi: NP: Could perhaps move this to a utility method since it's been called a few times throughout the code and may be used more in the future.

penghuo: In my opinion, it is the methodology for handling the nullable value and return default value. It is possible to build the common static method and put it into util class, but I don't think it could help more and eventually will make the utils class out of control. any thoughts?

qreshi: Okay, if you don't find it all that helpful then it's fine to leave it as is.

PR #42, opened 2019-04-29
Csv fix

arsen-es:
*Issue #, if available:* #40 

*Description of changes:* Fixed the csv output of queries (#40), updated integration tests, plus a
couple of minor fixes to make maven build pass.

*Tested:* Manually tested on local cluster, plus added integ tests.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #20, opened 2019-03-29
Enable PainlessPlugin for integ tests


arsen-es:
`ESIntegTestCase` base class starts an in-memory elasticsearch cluster with no plugins attached. We need the PainlessPlugin to enable lang-painless for translation of queries requiring scripts, e.g. queries containing date or math functions (YEAR, FLOOR, etc.).

arsen-es:
Fixed by #55

PR #49, opened 2019-05-13
Fixing failing integ tests for mvn build

arsen-es:
*Issue #, if available:* N/A

*Description of changes:* After introducing new data in test indexes (see #42 ), some of the integ tests were broken. This change fixes those.

We didn't notice this, since we rely on gradle build mostly, which does not run the mvn integ tests. Once #20 is fixed all the remaining mvn tests can be migrated to the new ESIntegTest subclasses and we can deprecate mvn. 

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/NestedFieldQueryTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/NestedFieldQueryTest.java
@@ -15,8 +15,8 @@
 
 package com.amazon.opendistroforelasticsearch.sql.intgtest;
 
-import com.amazon.opendistroforelasticsearch.sql.plugin.SearchDao;
 import com.amazon.opendistroforelasticsearch.sql.exception.SqlParseException;
+import com.amazon.opendistroforelasticsearch.sql.plugin.SearchDao;
 import com.amazon.opendistroforelasticsearch.sql.query.SqlElasticRequestBuilder;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.search.SearchHit;
@@ -27,11 +27,15 @@ import org.elasticsearch.search.aggregations.bucket.terms.Terms;
 import org.elasticsearch.search.aggregations.metrics.Avg;
 import org.elasticsearch.search.aggregations.metrics.Sum;
 import org.elasticsearch.search.aggregations.metrics.ValueCount;
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
 import org.hamcrest.FeatureMatcher;
 import org.hamcrest.Matcher;
 import org.junit.Test;
 
 import java.sql.SQLFeatureNotSupportedException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.Function;
 
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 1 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/NestedFieldQueryTest.java
```diff
@@ -341,6 +355,40 @@ public void groupByNestedAndRegularField() {
...
+    private Matcher<SearchHit> myNum(int[] values) {
+
+        return new BaseMatcher<SearchHit>() {
+
+            @Override
+            public boolean matches(Object item) {
+
+                if (item instanceof SearchHit) {
+                    final SearchHit hit = (SearchHit)item;
+                    List<Integer> actualValues = (ArrayList<Integer>)hit.getSourceAsMap().get("myNum");
```
qreshi: NP: Typically a space follows after a cast. Same with ```(SearchHit)``` in the line above.

src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/NestedFieldQueryTest.java
```diff
@@ -341,6 +355,40 @@ public void groupByNestedAndRegularField() {
...
+
+                    if (actualValues.size() != values.length) {
+                        return false;
+                    }
+
```
abbashus: LGTM. np: I see a lot of extra empty lines in this function. See if it can be removed without disturbing the overall structure of the class.

arsen-es: Removed some new lines.

src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/NestedFieldQueryTest.java
```diff
@@ -341,6 +355,37 @@ public void groupByNestedAndRegularField() {
...
+        return new BaseMatcher<SearchHit>() {
+
+            @Override
+            public boolean matches(Object item) {
+
+                if (item instanceof SearchHit) {
+                    final SearchHit hit = (SearchHit) item;
+                    List<Integer> actualValues = (ArrayList<Integer>) hit.getSourceAsMap().get("myNum");
+
+                    if (actualValues.size() != values.length) {
```
jngz-es: Should we validate if the values is null?

arsen-es: ideally yes, but since this is to-be-deprecated test code, and NPE will fail the test anyway, I don't see a need for that.

src/test/java/com/amazon/opendistroforelasticsearch/sql/intgtest/NestedFieldQueryTest.java
```diff
@@ -341,6 +355,37 @@ public void groupByNestedAndRegularField() {
...
+                        return false;
+                    }
+                    for (int value : values) {
+                        if (!actualValues.contains(value)) {
```
jngz-es: I don't think "contains" works here, it is not equal. For example, actualValues is [3, 3, 4] and values is [3, 4, 4]. The "if" condition will meet but they are different values.

arsen-es: Agree, changing to 1-by-1 comparison of elements.

Issue #17, opened 2019-03-27
Migrate Legacy Integration Tests to ESTest in Gradle


dai-chen:
There are still quite a few integration tests in Maven which perform "fake" integration testing (create core classes of plugin in memory with connection to an external ES cluster). We need to migrate all of them to Gradle to do it in standard ES Test manner.

arsen-es:
All test classes are now migrated, there are some test cases which are blocked by different reasons, but those will be tracked in their own issues, resolving this.

PR #56, opened 2019-05-20
Adding coverage report

arsen-es:
*Issue #, if available:* #17 

*Description of changes:* Migrating Show clause tests, plus, adding jacoco coverage report.

### DIFF
--- a/build-tools/sqlplugin-coverage.gradle
+++ b/build-tools/sqlplugin-coverage.gradle
@@ -64,9 +64,10 @@ integTestCluster {
 
 jacocoTestReport {
     dependsOn integTest, test
-    executionData dummyTest.jacoco.destinationFile, dummyIntegTest.jacoco.destinationFile
-    sourceDirectories = sourceSets.main.allSource
-    classDirectories = sourceSets.main.output
+    executionData.from = [dummyTest.jacoco.destinationFile, dummyIntegTest.jacoco.destinationFile]
+    sourceDirectories.from = sourceSets.main.java.sourceDirectories
+    classDirectories.from = files(sourceSets.main.java.outputDir)
+
     reports {
         html.enabled = true // human readable
         csv.enabled = true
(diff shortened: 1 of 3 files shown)


arsen-es:
The coverage change is mostly copied from Alerting plugin, you can check their code for reference if interested.

build-tools/sqlplugin-coverage.gradle
```diff
@@ -0,0 +1,96 @@
...
+ *
+ *  To workaround these we start the cluster with jmx enabled and then use Jacoco's JMX MBean to get the execution data before the
+ *  cluster is stopped and dump it to a file. Luckily our current security policy seems to allow this. This will also probably
+ *  break if there are multiple nodes in the integTestCluster. But for now... it sorta works.
+ */
+apply plugin: 'jacoco'
+
+// Get gradle to generate the required jvm agent arg for us using a dummy tasks of type Test. Unfortunately Elastic's
+// testing tasks don't derive from Test so the jacoco plugin can't do this automatically.
+task dummyTest(type: Test) {
```
penghuo: why this task is needed?

build-tools/sqlplugin-coverage.gradle
```diff
@@ -0,0 +1,96 @@
...
+    enabled = false
+    workingDir = file("/") // Force absolute path to jacoco agent jar
+    jacoco {
+        destinationFile = file("${buildDir}/jacoco/test.exec")
+        destinationFile.parentFile.mkdirs()
+        jmx = true
+    }
+}
+
+task dummyIntegTest(type: Test) {
```
penghuo: ditto

build-tools/sqlplugin-coverage.gradle
arsen-es: This one is only used to store the destinationFile setting, we can get rid of this completely and inline that value at the place of usage.

build-tools/sqlplugin-coverage.gradle
arsen-es: This one is used for the jacaco nested object which is converted to jvmArgs in the `integTestCluster`. As for why exactly it is passed in, I think to account for *IT classes' contribution in coverage. I tried building the argument manually, but did not find a good way.

PR #63, opened 2019-05-25
Moving the execution of subqueries from a transport to a worker thread.

arsen-es:
*Issue #, if available:* N/A

*Description of changes:*
If a query has a subquery, the replacement of the subquery
with actual values takes place before the main query execution and on
the same thread that handles the request - the transport thread, which
is not to be used for blocking actions. This change moves the execution
of the subquery to the sql worker threadpool.

*How Tested:* if '-ea' (enable assertions) jvm option is enabled, any
query with a subquery would fail because of this issue. Verified that
with this change the query succeeds. Sample query used:
'SELECT firstname FROM account WHERE firstname IN (SELECT firstname FROM
people)'. Also added integ tests for the same query.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubQueryIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubQueryIT.java
@@ -1,0 +1,78 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubQueryIT.java
```diff
@@ -0,0 +1,78 @@
...
+        final JSONObject result = executeQuery(query);
+        final JSONArray hits = getHits(result);
+
+        final Set<String> expectedNames = new HashSet<>(Arrays.asList("Hattie", "Nanette", "Dale", "Elinor", "Virginia",
+                "Dillard", "Mcgee", "Aurelia", "Fulton", "Burton", "Josie"));
+        Assert.assertThat(hits.length(), equalTo(expectedNames.size()));
+        Set<String> actualNames = new HashSet<>();
+        hits.forEach(obj -> {
+            JSONObject hit = (JSONObject)obj;
+            actualNames.add(hit.getJSONObject("_source").getString("firstname"));
```
qreshi: NP: Could possibly call ```getSource()``` from ```SQLIntegTestCase``` class here or even extend it to take field names for cases like this.

arsen-es: Agree, but don't want to increase the scope of this change. I might address that when I do some more integ tests.

qreshi: Sounds good.

arsen-es:
I am working to completely make this async instead of just moving part of the execution to a worker thread. Not pushing this change for now.

Issue #50, opened 2019-05-13
Add support to enable/disable SQL feature


abbashus:
Add new dynamic cluster setting to enable/disable SQL feature

abbashus:
Resolved by #65

PR #65, opened 2019-05-28
Support for enabling/disabling SQL feature

abbashus:
*Issue #50*

*Description of changes:* This change allows enabling/disabling SQL feature by two methods
- Persistent and transient dynamic cluster setting `opendistro.sql.enabled`
- Node level unmodifiable setting `rest.action.multi.allow_explicit_index`

*Testing:* Existing unit and integ tests passing, Added new integration tests for the above change. 


src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
```diff
@@ -0,0 +1,26 @@
...
+package com.amazon.opendistroforelasticsearch.sql.exception;
+
+public class SQLFeatureDisabledException extends Exception {
+
+    private static final long serialVersionUID = 1L;
+
+    public SQLFeatureDisabledException(String message) {
+        super(message);
+    }
+
```
arsen-es: np: redundant new line

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.exception;
+
+public class SQLFeatureDisabledException extends Exception {
```
arsen-es: I think this class represents a very concrete exception, so one should be able to use a default message, instead of always passing it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -73,6 +79,11 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
         Metrics.getInstance().getNumericalMetric(MetricName.REQ_COUNT_TOTAL).increment();
         SqlRequest sqlRequest = SqlRequest.NULL;
         try {
+
+            if (!isSQLFeatureEnabled()) {
+                throw new SQLFeatureDisabledException("Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false");
```
arsen-es: np: seems like longer than 120 characters, please break the line if it is. We will eventually enable checkstyle to catch these errors, but for now let's at least not create more.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    protected void init() throws Exception {
+        loadIndex(Index.ACCOUNT);
+    }
+
+    @Test
+    public void sqlEnableSettingsTest() throws IOException {
+        updateClusterSettings(SettingType.PERSISTENT, "opendistro.sql.enabled", "true");
+        JSONObject queryResult = executeQuery(String.format("SELECT firstname FROM %s/account WHERE account_number=1", TEST_INDEX_ACCOUNT));
```
arsen-es: please use `Locale.ROOT` with `String.format`

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
```diff
@@ -0,0 +1,95 @@
...
+     * java.lang.AssertionError: test leaves persistent cluster metadata behind
+     *
+     * TODO: Look for a way to reset all the persistent and transient settings instead of
+     *       individually resetting of each such setting
+     */
+    private JSONObject resetClusterSettings(SettingType settingType, String setting) throws IOException {
+        return updateClusterSettings(settingType, setting, "null");
+    }
+
+    private String executeRequest(final Request request) throws IOException {
```
arsen-es: you could take the existing `SqlIntegTestCase::executeRequest(Request)` method (see [here](https://github.com/opendistro-for-elasticsearch/sql/blob/d0f2e0b5a72df8094a57af2250b45460ed92ba3a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java#L157-L175)), and just change the assert line to compare against a provided status code instead of hardcoded `200`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
```diff
@@ -0,0 +1,95 @@
...
+        JSONObject error = queryResult.getJSONObject("error");
+        Assert.assertThat(error.getString("reason"), equalTo("Invalid SQL query"));
+        Assert.assertThat(error.getString("details"), equalTo("Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false"));
+        Assert.assertThat(error.getString("type"), equalTo("SQLFeatureDisabledException"));
+        resetClusterSettings(SettingType.PERSISTENT, "opendistro.sql.enabled");
+    }
+
+    private JSONObject updateClusterSettings(SettingType settingType, String setting , String value) throws IOException {
+        Request request = new Request("PUT", "/_cluster/settings?pretty");
+        String persistentSetting = "{\""+ settingType.val() +"\": {\"" + setting + "\": " + value + "}}";
```
arsen-es: np: if there are more than 1 substitutions, `String.format` makes it more readable.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
```diff
@@ -0,0 +1,95 @@
...
+        request.setJsonEntity(persistentSetting);
+        RequestOptions.Builder restOptionsBuilder = RequestOptions.DEFAULT.toBuilder();
+        restOptionsBuilder.addHeader("Content-Type", "application/json");
+        request.setOptions(restOptionsBuilder);
+        JSONObject result = new JSONObject(executeRequest(request));
+        System.out.println(result);
```
arsen-es: remove

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
```diff
@@ -0,0 +1,95 @@
...
+            response = ex.getResponse();
+        }
+
+        queryResult = new JSONObject(TestUtils.getResponseBody(response));
+        Assert.assertThat(queryResult.getInt("status"), equalTo(400));
+        JSONObject error = queryResult.getJSONObject("error");
+        Assert.assertThat(error.getString("reason"), equalTo("Invalid SQL query"));
+        Assert.assertThat(error.getString("details"), equalTo("Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false"));
+        Assert.assertThat(error.getString("type"), equalTo("SQLFeatureDisabledException"));
+        resetClusterSettings(SettingType.PERSISTENT, "opendistro.sql.enabled");
```
arsen-es: do you also need to reset transient settings? If no, I suggest getting rid of that settings type and just using one constant String literal instead of enum.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
abbashus: This is needed for debugging. We do have  `System.out.println();` in other parts of the test codebase as well.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
abbashus: Will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
abbashus: Will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
abbashus: This has been already implemented like this internally, so its good to keep it like that. Do not want to divert much.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
abbashus: Will remove.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
abbashus: Will do.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
arsen-es: We shouldn't. We can add it locally when needed, but we shouldn't check in. We could have a logger configured for tests that will output to console, but I think it would be more useful if it weren't easy to change the code while debugging, whereas in this case you can always update the code and run locally, so I don't think there's a good reason to keep any of the `System.out.print*` calls.

abbashus: Will remove.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
abbashus: Okay, will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
abbashus: I think this allows us to reuse this exception if we want to disable SQL feature for a different reason.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
abbashus: We are not asserting anything here. Also its a private method that I can't use directly in `PluginIT`. I could make it protected and change the signature to `SqlIntegTestCase::executeRequest(Request, statusCode)` and change the all the usages, if that's what intended.

arsen-es: ok, up to you, I just wanted to avoid much duplicate code.

abbashus: Fortunately , I was able to reuse `SqlIntegTestCase::executeRequest(Request)` without changing signature. But still needed `TestUtils.getResponseBody` to parse response from exception.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
arsen-es: np: I was suggesting to have a default constructor along with the one accepting a message, not critical though.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SQLFeatureDisabledException.java
arsen-es: I meant the line before the last closing brace, not after.

Issue #64, opened 2019-05-26
support "SELECT 1"


dictcp:
Why
----
there are BI tool (eg. metabase) using simple dummy query "SELECT 1" to detect if the connection works.
But it is not supported according to
https://opendistro.github.io/for-elasticsearch-docs/docs/sql/operations/

How
----
Can consider to hardcode a response "1"

penghuo:
Thanks for bringing this up. We are looking into this now.

jordanw-bq:
We are currently encountering this issue when connecting the ODBC driver to Tableau. Looking at the linked PR (#68), the issue was not fully addressed.

The particular case that needs to be supported is for the exact query `SELECT 1`, without a FROM clause. The expected result set should be a single column (column name does not matter, can use the default approach for situations like unaliased functions in the SELECT list), and a single row containing the integer `1`.

Example from Presto:
```
 _col0
-------
     1
```

This seems to be necessary for using the ODBC driver with Tableau, as we have so far not found a way to disable this type of discovery query.

dai-chen:
We have the issue for tracking: https://github.com/opendistro-for-elasticsearch/sql/issues/264

PR #68, opened 2019-06-03
Support Select 1

penghuo:
*Issue #, if available:* #64

*Description of changes:*
Support numeric literals.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
@@ -85,6 +85,10 @@ public class FieldMaker {
             methodParameters.add(new KVValue(alias));
             methodParameters.add(new KVValue(scriptCode));
             return new MethodField("script", methodParameters, null, alias);
+        } else if (expr instanceof SQLIntegerExpr) {
+            SQLMethodInvokeExpr methodInvokeExpr = new SQLMethodInvokeExpr("assign", null);
+            methodInvokeExpr.addParameter(expr);
+            return makeMethodField(methodInvokeExpr.getMethodName(), methodInvokeExpr.getParameters(), null, alias, tableAlias, true);
         } else {
             throw new SqlParseException("unknown field name : " + expr);
         }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
@@ -85,7 +85,7 @@ public class FieldMaker {
             methodParameters.add(new KVValue(alias));
             methodParameters.add(new KVValue(scriptCode));
             return new MethodField("script", methodParameters, null, alias);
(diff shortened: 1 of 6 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -85,6 +85,10 @@ public static Field makeField(SQLExpr expr, String alias, String tableAlias) thr
             methodParameters.add(new KVValue(alias));
             methodParameters.add(new KVValue(scriptCode));
             return new MethodField("script", methodParameters, null, alias);
+        } else if (expr instanceof SQLIntegerExpr) {
```
arsen-es: is it only for integers? how about `SELECT 1.0 FROM myData`?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -191,4 +192,14 @@ public void split_field() throws Exception {
...
+    @Test
+    public void functionAssign() throws Exception {
+        String query = "SELECT 10 as key "+
+                "from " + TEST_INDEX_ACCOUNT + "/account limit 1";
+
+        assertThat(
+                executeQuery(query),
+                hitAny(kvInt("/fields/key/0", equalTo(10))));
+    }
 }
```
arsen-es: please add another test without assigning an alias to the field.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
penghuo: Good point! Add the support for numeric literals.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
penghuo: Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/FieldMakerTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.junit.Assert.assertTrue;
+
+public class FieldMakerTest {
+
+    private static final String ALIAS = "a";
+
+    private static final String TABLE_ALIAS = "t";
+
+    @Test
+    public void makeFiled_assign() throws SqlParseException {
```
qreshi: Typo: I believe you meant ```makeField_assign()```, same for the test below

abbashus: Just to add, we should be consistent with Java naming convention. It should be `makeFieldAssign()` here.

penghuo: Done.

PR #71, opened 2019-06-05
Fix #46

### LINKED ISSUE #46: VerificationException: When using multiple indices, the mappings must be identical.
I'm trying to run SQL queries on an index alias (e.g. _select field from myindex-*_) and my indexes are daily (so I have myindex-2019-04-01, etc.).

For some of the indexes, additional fields have been added and dynamic field mapping has updated the mapping, resulting on myindex-2019-04-01 having less fields compared to myindex-2019-04-02 for example.

When running SQL queries, elasticsearch is complaining with:

> VerificationException: When using multiple indices, the mappings must be identical.

even though the SQL query is not using the fields that are different between the indexes.
Is there any way to relax the strictness of all mappings having to be identical?

if I run:

> select myfield from myindex*

and the mapping for _myfield_ is the same on all indexes matching myindex*, do we really want to enforce the whole mapping is identical? I'm not seeing the benefit of that.
even if some indexes do not contain _myfield_, wouldn't it be better to just return an empty value for such column instead of not allowing the request at all?

zhongnansu:
*Issue #46 :*
*Testing:* Passed gradle build
*Description of changes:*
Made minimal change to get rid of identical mapping check when query from multi-index, but still keeping the unknown index check. Updated corresponding tests.


src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,12 +210,8 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
             );
     }
 
-    public boolean hasIdenticalMappings(TermFieldScope scope, Map<String, String> indexToType) {
+    public boolean hasUnknownIndex(TermFieldScope scope) {
```
arsen-es: What is the definition of "unknown index"? Based on the code in this method, as well as on lines 76-77, I think the logic should be inverted.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
```diff
@@ -102,6 +102,7 @@ public void testNonResolvingIndexPatternWithNonExistingIndex() throws IOExceptio
     }
 
     @Test
+    @Ignore("allow non-identical mapping when query multi-index")
```
arsen-es: I prefer deleting the tests that are not asserting any truth anymore.

zhongnansu: Deleted

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: You're right. "unknown index" refers to index not existed, or can not be matched by wildcard(in this case). I modified the logic to make more sense.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +210,15 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
             );
     }
 
-    public boolean hasIdenticalMappings(TermFieldScope scope, Map<String, String> indexToType) {
```
abbashus: I realized that relaxing the mapping check completely will break the functionality where mappings are used to generate Elasticsearch DSL such as `WHERE =` clause, `GROUP BY`, `ORDER By ` etc. This is because it will make it ambiguous as to mapping from which index to be applied. Lets  keep this function, but only check for mappings present as part of `SELECT` clause.

zhongnansu:
Switch to another approach.

PR #80, opened 2019-06-11
Converting HashJoinMoreTest and HavingTest

qreshi:
*Issue #, if available:* #17 

*Description of changes:* 
Renamed HashJoinBasicIT to HashJoinIT and migrated HashJoinMoreTest into HashJoinIT.
Migrated HavingTest.
Updated test index mappings.

*Testing method:* Tested using gradle build


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
```diff
@@ -131,29 +131,45 @@ public static String getPeople2IndexMapping() {
...
+                "          },\n" +
+                "          \"ofHisName\": {\n" +
+                "            \"type\": \"integer\"\n" +
+                "          }\n" +
+                "        }\n" +
+                "      },\n" +
+                "      \"house\": {\n" +
+                "        \"type\": \"text\",\n" +
+                "        \"fields\": {\n" +
+                "          \"keyword\": {\n" +
```
arsen-es: Two questions here:
1. How does this field get populated if you didn't modify the data to be indexed? Does a field with name "keyword" get handled in a special way?
2. Wouldn't it be better to just change the the type of the fields `house` and `gender` to `keyword` instead of adding a nested field?

qreshi: 1. I'm not sure I understand the first part of this question but the [fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html) setting is used when you want to index the same field in a different way. So in this case we also index ```gender``` and ```house``` as type ```keyword``` and refer to this as ```keyword```. The actual name given to the additional field is arbitrary and could have been anything from my understanding.
2. Although in most cases we prefer using a field with type ```keyword```, I decided to leave the original type as text and include a multi-field of type keyword. That way we still have the originally defined type for those fields, should we need them. If you think this is unnecessary, I can remove it and just change the type from ```text``` -> ```keyword```.

arsen-es: Alright, sounds good, it doesn't really matter much for tests, I just thought it was unnecessary.

Issue #34, opened 2019-04-18
Improve Tracking Request ID in Logging


dai-chen:
Currently we prepended a generated request ID to logging related to a request. In the PR, @arsen-es proposed to investigate other approach provided by Log4j, such as maintain thread context and let Log4j take care of this.

dai-chen:
I think it would be great if we could do this programmatically without touching Log4j config file.

PR #81, opened 2019-06-12
Detaching request id from SqlRequest

arsen-es:
*Issue #, if available:* #34 

*Description of changes:* Removed `id` field from `SqlRequest` class. Using logging library's ThreadContext to store the request id as soon as we get it, and passing it around to make it available in case new threads are started from the main one. Provided a utility class to make the access to the ThreadContext simpler.

The current layout pattern of elasticsearch log messages is `[%d{ISO8601}][%-5p][%-25c{1.}] [%node_name]%marker %m%n`, if we pre-pend `[%X{request_id}]` all the messages will automatically be pre-pended with the request id. However, the plugin code does not manage the log configuration, and I didn't want to modify it at run-time.

*How tested:* Made some requests on local cluster, made sure request id is getting properly populated. Tested for tasks running on worker threads as well.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
@@ -28,6 +28,7 @@ import com.amazon.opendistroforelasticsearch.sql.query.QueryAction;
 import com.amazon.opendistroforelasticsearch.sql.request.SqlRequest;
 import com.amazon.opendistroforelasticsearch.sql.request.SqlRequestFactory;
 import com.amazon.opendistroforelasticsearch.sql.rewriter.matchtoterm.VerificationException;
+import com.amazon.opendistroforelasticsearch.sql.utils.LogUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.client.node.NodeClient;
@@ -43,6 +44,7 @@ import java.sql.SQLFeatureNotSupportedException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -52,14 +54,19 @@ import static org.elasticsearch.rest.RestStatus.OK;
 import static org.elasticsearch.rest.RestStatus.SERVICE_UNAVAILABLE;
 
 public class RestSqlAction extends BaseRestHandler {
+
     private static final Logger LOG = LogManager.getLogger(RestSqlAction.class);
+
     private final boolean allowExplicitIndex;
 
-    /** API endpoint path */
+    /**
+     * API endpoint path
+     */
     public static final String QUERY_API_ENDPOINT = "/_opendistro/_sql";
     public static final String EXPLAIN_API_ENDPOINT = QUERY_API_ENDPOINT + "/_explain";
 
-    public RestSqlAction(Settings settings, RestController restController) {
+    RestSqlAction(Settings settings, RestController restController) {
+
         super(settings);
         restController.registerHandler(RestRequest.Method.POST, QUERY_API_ENDPOINT, this);
         restController.registerHandler(RestRequest.Method.GET, QUERY_API_ENDPOINT, this);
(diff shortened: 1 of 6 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -78,17 +85,19 @@ public String getName() {
...
                 throw new SQLFeatureDisabledException(
-                    "Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false"
+                        "Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false"
                 );
             }
 
-            sqlRequest = SqlRequestFactory.getSqlRequest(request);
-            LOG.info("[{}] Incoming request {}: {}", sqlRequest.getId(), request.uri(), sqlRequest.getSql());
+            final SqlRequest sqlRequest = SqlRequestFactory.getSqlRequest(request);
+            LOG.info("[{}] Incoming request {}: {}", LogUtils.getRequestId(), request.uri(), sqlRequest.getSql());
```
abbashus: Though not related to original fix, but can you also remove "Incoming request" wording,  as logically all requests are incoming and it also does not add any value , except for maybe grepping logs.

arsen-es: I can remove the word _incoming_, but I think it's there to indicate this it the first log message about this request, so if you for instance scroll through a log file trying to see the end-to-end execution logs for a given request, this will indicate the start. Doesn't really matter much for me, though, I am fine with just keeping the word _request_.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/LogUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Utility class for generating/accessing the request id from logging context.
+ */
+public class LogUtils {
+
+    /**
+     * The key of the request id in the context map
+     */
+    public static final String REQUEST_ID_KEY = "request_id";
```
dai-chen: `private` may be better? Since this is more like your internal detail rather than a public constant.

arsen-es: Agree, no reason to keep it `public`. I was trying to do something fancier to basically check if the elastic's thread context has the id and get it from there, that's why made this `public`, but that would be a different change anyway (if we ever make it), changing this to `private`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/LogUtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+
+public class LogUtilsTest {
+
+    @After
+    public void cleanUpContext() {
+
+        ThreadContext.clearMap();
```
dai-chen: When this or cleanup of context happen in our plugin? Or we just generate an ID for each thread and keep using it forever?

arsen-es: this never happens, it's just a way to make the test cases independent.

arsen-es: also, we should generate only once for the main (accepting the request thread), and pass it over to other threads created from within that one. As long as the threads are working on the same request, they should share the same id.

dai-chen: > this never happens, it's just a way to make the test cases independent.

Does this mean for all the following request served by the thread, we keep using same ID?

dai-chen: > also, we should generate only once for the main (accepting the request thread), and pass it over to other threads created from within that one. As long as the threads are working on the same request, they should share the same id.

Need more context how this works. So you're saying we generate an ID for each transport thread only once. Then I saw you "preserve" the context when you pass it to worker thread. So for all the logging on the path (for a transport thread), same ID would be in use always between requests?

arsen-es: OK, I see your point now. You're right, I need to clean up the request id after the processing is done or overwrite it when the processing starts. It should be per request, not per thread, and with current code it won't work for re-used threads.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -114,7 +124,7 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
             } else {
                 Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
             }
-            LOG.error(String.format("[%s] Failed during query execution", sqlRequest.getId()), e);
+            LOG.error(String.format(Locale.ROOT, "[%s] Failed during query execution", LogUtils.getRequestId()), e);
```
dai-chen: Just want to learn, what's the difference between this `ROOT` locale and default locale used by `String.format`? Any concern here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
abbashus: Sounds plausible to me. Lets keep it then.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
arsen-es: ROOT is the location/language/culture-neutral locale, which means if we are using it, the code will behave the same were independent of environment. If you do not specify the `Locale`, it picks the default one, and default means JVM just gets it from the underlying OS settings. That in its turn means if I set my windows location to Armenia and language to Armenian, and you set to US and English, we might get different result from the `String.format` call. Of course we usually don't use special characters which depend on locale, but it's a good practice to use `Locale.ROOT` always (not for strings exposed on the UI) to be on the safe side.

dai-chen: > ROOT is the location/language/culture-neutral locale, which means if we are using it, the code will behave the same were independent of environment. If you do not specify the `Locale`, it picks the default one, and default means JVM just gets it from the underlying OS settings. That in its turn means if I set my windows location to Armenia and language to Armenian, and you set to US and English, we might get different result from the `String.format` call. Of course we usually don't use special characters which depend on locale, but it's a good practice to use `Locale.ROOT` always (not for strings exposed on the UI) to be on the safe side.

Good to know this. Thanks!

PR #83, opened 2019-06-14
Moving subquery execution to a worker thread

arsen-es:
*Issue #, if available:* N/A

*Description of changes:* Run queries including a subquery on a worker thread. This change does not affect the execution of other types of queries. This is just a first step to unblock the queries with `WHERE field IN (subquery)` syntax, but still needs improvement including but not limited to the following:
 1. the condition of whether the query has a subquery should ideally be derived from the Sql parser, not by string manipulation,
 1. the explain call should not execute the subquery, but should just explain how it will be executed.

*How tested:* All integ tests pass, manually tested on queries that failed previously (with jvm assertions enabled) and they pass with this change.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
@@ -17,9 +17,10 @@ package com.amazon.opendistroforelasticsearch.sql.plugin;
 
 import com.alibaba.druid.sql.parser.ParserException;
 import com.amazon.opendistroforelasticsearch.sql.esdomain.LocalClusterState;
-import com.amazon.opendistroforelasticsearch.sql.exception.SqlParseException;
 import com.amazon.opendistroforelasticsearch.sql.exception.SQLFeatureDisabledException;
+import com.amazon.opendistroforelasticsearch.sql.exception.SqlParseException;
 import com.amazon.opendistroforelasticsearch.sql.executor.ActionRequestRestExecutorFactory;
+import com.amazon.opendistroforelasticsearch.sql.executor.AsyncRestExecutor;
 import com.amazon.opendistroforelasticsearch.sql.executor.RestExecutor;
 import com.amazon.opendistroforelasticsearch.sql.executor.format.ErrorMessage;
 import com.amazon.opendistroforelasticsearch.sql.metrics.MetricName;
@@ -31,14 +32,18 @@ import com.amazon.opendistroforelasticsearch.sql.rewriter.matchtoterm.Verificati
 import com.amazon.opendistroforelasticsearch.sql.utils.LogUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.elasticsearch.client.Client;
 import org.elasticsearch.client.node.NodeClient;
 import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.index.IndexNotFoundException;
 import org.elasticsearch.rest.BaseRestHandler;
 import org.elasticsearch.rest.BytesRestResponse;
+import org.elasticsearch.rest.RestChannel;
 import org.elasticsearch.rest.RestController;
 import org.elasticsearch.rest.RestRequest;
 import org.elasticsearch.rest.RestStatus;
+import org.elasticsearch.threadpool.ThreadPool;
 
 import java.sql.SQLFeatureNotSupportedException;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


dai-chen:
IMHO, trying to avoid query execution in explain is the better way (or at least in this case of subquery). The code would be clean, same logic for different explain and no need for hack like in your `containsSubquery`. Unless customer really wants an executable DSL from explain anyway. I think explain of Join, especially Nested Loop, could be your reference.

arsen-es:
@dai-chen The intention was to unblock this use case before coming up with the more structured and generic solution, so I agree that this is not the best way. Considering we don't claim supporting subqueries yet, I am closing this PR, and will work on the better solution.

dai-chen:
> @dai-chen The intention was to unblock this use case before coming up with the more structured and generic solution, so I agree that this is not the best way. Considering we don't claim supporting subqueries yet, I am closing this PR, and will work on the better solution.

Yeah, my concern is we're introducing a major code change which is very likely to refactor or even revert later. Thanks for your effort on this anyway!

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -147,16 +205,16 @@ private boolean isClientError(Exception e) {
                 e instanceof VerificationException;
     }
 
-    private RestChannelConsumer reportError(Exception e, RestStatus status) {
-        return sendResponse(new ErrorMessage(e, status.getStatus()).toString(), status);
+    private void sendResponse(final RestChannel channel, final String message, final RestStatus status) {
```
penghuo: is it possible to avoid these unnecessary change.

arsen-es:
Re-opening, as we are considering having a quick fix before we roll out the better support for sub-queries, which might take longer time.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -136,7 +140,61 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+    }
+
+    private static boolean explainRequiresBlockingIOCall(final SqlRequest request) {
+
+        return containsSubquery(request);
+    }
+
+    private static boolean containsSubquery(final SqlRequest request) {
+
+        return request.getSql().toLowerCase(Locale.ROOT).contains("(select ");
```
penghuo: how to handle "( select"? Is there any preprocessing already handle this?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
arsen-es: the return type is changed, this is necessary.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
arsen-es: I can check for regex match, instead of hard-coded string, e.g.:
```
return request.getSql().toLowerCase(Locale.ROOT).matches("\\(\\s*select ");
```

arsen-es:
> 
> 
> Becase this PR is the quick fix for subquery, could you make it easy to revert in the future?
> e.g. remove unnecessary code change.

Seems like I can revert the change of (need to double check)
src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java

dai-chen:
> > Becase this PR is the quick fix for subquery, could you make it easy to revert in the future?
> > e.g. remove unnecessary code change.
> 
> Seems like I can revert the change of (need to double check)
> src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java

And it's saying "This branch has conflicts that must be resolved". Could you merge latest changes in master too in your new revision?

arsen-es:
> 
> 
> > > Becase this PR is the quick fix for subquery, could you make it easy to revert in the future?
> > > e.g. remove unnecessary code change.
> > 
> > 
> > Seems like I can revert the change of (need to double check)
> > src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
> 
> And it's saying "This branch has conflicts that must be resolved". Could you merge latest changes in master too in your new revision?

Sure @dai-chen, I should have done that in the first place. Working on it now...

arsen-es:
I have addressed all the comments in #115 . Couldn't update this PR with the changes, since it wanted me to rebase the branch, and that would include too many commits, that's why moved to a new PR. Please continue the review there, I am closing this one.

PR #84, opened 2019-06-18
[Fix #46] support multi-indices query with compatible mappings. Give …

### LINKED ISSUE #46: VerificationException: When using multiple indices, the mappings must be identical.

zhongnansu:
*Issue #, if available:*#46

*Description of changes:*
1. Previously we were throwing VerificationException if the mappings were non-identical for expanded indices. Even though the fields of the indices with extra mappings are not used we were throwing VerificationException.
2. For fields with same name but different mappings across indices, we will throw VerificationEXception highlighting the differences in mapping causing this exception. 

*Test:* 
Passes existing Gradle build and tests. Added new integration tests.


src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
```diff
@@ -35,11 +35,7 @@
 import org.json.JSONObject;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
arsen-es: Please do not use star imports, it makes it more difficult to quickly identify where each of the used types is coming from. You can update IntelliJ's setting to not auto-convert to star.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
```diff
@@ -399,6 +395,12 @@ public FieldMappings(MappingMetaData mappings) {
             fieldMappings = mappings.sourceAsMap();
         }
 
+        public FieldMappings(Map<String, Map<String, Object>> map) {
```
arsen-es: Please add documentation for public methods, especially for the cases like this one, where it's not obvious (e.g.  why we need the second constructor).

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+                                 flatMap(typeMappings -> typeMappings.allMappings().stream()).
+                                 collect(Collectors.toSet());
 
-        if (isMappingOfAllIndicesDifferent()) {
-            return false;
-        }
+        FieldMappings fieldMappings = null;
+
+        if (set.size() > 1) {
+            Map<String, Map<String, Object>> map = new LinkedHashMap<>();
```
arsen-es: please use more descriptive names for variables.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
     }
 
-    public boolean hasIdenticalMappings(TermFieldScope scope, Map<String, String> indexToType) {
+    public void checkMappingCompatibility(TermFieldScope scope, Map<String, String> indexToType) {
         if (scope.getMapper().isEmpty()) {
             throw new VerificationException("Unknown index " + indexToType.keySet());
         }
+        // Collect all FieldMappings into hash set and ignore index/type names. set.size > 1 means FieldMappings NOT unique,
+        // needs further compatibility check
+        Set<FieldMappings> set = curScope().getMapper().allMappings().stream().
```
arsen-es: use more descriptive names, the identifier should represent what the underlying object's functional purpose is, not what type it is of.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
-        }
+        FieldMappings fieldMappings = null;
+
+        if (set.size() > 1) {
+            Map<String, Map<String, Object>> map = new LinkedHashMap<>();
+
+            for (FieldMappings f : set) {
+                Map<String, Map<String, Object>> m = f.data();
+
+                for (Map.Entry<String, Map<String, Object>> e : m.entrySet()) {
```
arsen-es: ideally, we should not have variables with names `m`, `e`, unless it's a simple index in a `for` loop, or an exception in a `catch` statement, maybe a couple more exceptions.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+                    String fieldName = e.getKey();
+                    Map<String, Object> fieldMappingValue = e.getValue();
+
+                    if (!map.containsKey(fieldName)) {
+                        map.put(fieldName, fieldMappingValue);
+                    } else {
+                        // check if types are same
+                        if (!fieldMappingValue.equals(map.get(fieldName))) {
+                           // TODO: Merge mappings if they are compatible, like text and text/keyword to text/keyword.
```
arsen-es: I think merge should take the intersection, not the union. Because otherwise one can use properties of a mapping for an index that does not actually have it. Also, please open an issue to track it, I don't think we have a clearly defined process for resolving TODOs.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
 
+                            String firstFieldType= new JSONObject(fieldMappingValue).toString().replaceAll("\"", "");
+                            String secondFieldType = new JSONObject(map.get(fieldName)).toString().replaceAll("\"", "");
+
+                            String exceptionReason = String.format("Cannot use field [%s] " +
+                                            "due to ambiguities being mapped as incompatible types: [%s] and [%s] ",
```
arsen-es: not sure what 'ambiguities' refers to here, but I think the message itself is ambiguous. +Ideally we provide the index name also (but that's not a blocker). Maybe just say `"Field [%s] has different different mappings in the specified indices. Index [%s] - mapping: [%s], index [%s] - mapping [%s]"`. If you want to omit index names, maybe something like `"Different mappings are not allowed for the same field: found [%s] and [%s]"`. Finally, please use `Locale.ROOT` with `String.format`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
arsen-es: seems you are not using any of the LinkedList features of this map, so just make it a `HashMap`

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
```diff
@@ -24,11 +24,7 @@
 import java.io.IOException;
 import java.util.Locale;
 
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_ACCOUNT;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_BANK;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_BANK_TWO;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_DOG;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_ONLINE;
+import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.*;
```
arsen-es: no star imports please

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
```diff
@@ -102,19 +100,28 @@ public void testNonResolvingIndexPatternWithNonExistingIndex() throws IOExceptio
     }
 
     @Test
-    public void testNonIdenticalMappings() throws IOException {
+    public void testNonCompatibleMappings() throws IOException {
         try {
-            explainQuery(String.format(Locale.ROOT, "SELECT firstname, birthdate FROM %s, %s " +
-                            "WHERE firstname = 'Leo' OR male = 'true'",
-                    TEST_INDEX_BANK, TEST_INDEX_ONLINE));
+            explainQuery(String.format(Locale.ROOT, "SELECT * FROM %s, %s ",
```
arsen-es: please add `Assert.fail()` after this line (in the `try` block), otherwise if no exception is thrown your test will still pass.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
```diff
@@ -102,19 +100,28 @@ public void testNonResolvingIndexPatternWithNonExistingIndex() throws IOExceptio
...
-            assertThat(entity, containsString("When using multiple indices, the mappings must be identical"));
+            assertThat(entity, containsString("dog_name"));
+            assertThat(entity, containsString("{type:text,fields:{keyword:{type:keyword,ignore_above:256}}}"));
+            assertThat(entity, containsString("{type:text,fielddata:true}"));
             assertThat(entity, containsString("\"type\": \"VerificationException\""));
         }
     }
 
+    @Test
+    public void testCompatibleMappings() throws IOException {
```
arsen-es: the name is misleading, since we don't check compatibility but equality, I'd rename this to `testEqualFieldMappings` or something alike. When that TODO is resolved and we really check compatibility, then we'll need to add more tests for that.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
             );
     }
 
-    public boolean hasIdenticalMappings(TermFieldScope scope, Map<String, String> indexToType) {
+    public void checkMappingCompatibility(TermFieldScope scope, Map<String, String> indexToType) {
```
arsen-es: can this method be `private`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+        Set<FieldMappings> set = curScope().getMapper().allMappings().stream().
+                                 flatMap(typeMappings -> typeMappings.allMappings().stream()).
+                                 collect(Collectors.toSet());
 
-        if (isMappingOfAllIndicesDifferent()) {
-            return false;
-        }
+        FieldMappings fieldMappings = null;
+
+        if (set.size() > 1) {
```
arsen-es: NP: this `if` block is too large and too nested, I think extracting out a pieces/pieces into smaller private methods would help with readability.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
dai-chen: > not sure what 'ambiguities' refers to here, but I think the message itself is ambiguous. +Ideally we provide the index name also (but that's not a blocker). Maybe just say `"Field [%s] has different different mappings in the specified indices. Index [%s] - mapping: [%s], index [%s] - mapping [%s]"`. If you want to omit index names, maybe something like `"Different mappings are not allowed for the same field: found [%s] and [%s]"`. Finally, please use `Locale.ROOT` with `String.format`.

Agree. Maybe we can make this easier, just give index name, field name/type and expected type. I think these should be sufficient for customer to understand what's happening.

@arsen-es btw, we may need a utility method if you think passing `Locale.ROOT` is required for each call to `String.format`.

abbashus: @arsen-es The message structure is taken from existing ElasticSearch SQL Exception

```
{
  "error": {
    "root_cause": [
      {
        "type": "verification_exception",
        "reason": "Found 1 problem(s)\nline 1:8: Cannot use field [field_1] due to ambiguities being mapped as [2] incompatible types: [text] in [test_1], [long] in [test_2]"
      }
    ],
    "type": "verification_exception",
    "reason": "Found 1 problem(s)\nline 1:8: Cannot use field [field_1] due to ambiguities being mapped as [2] incompatible types: [text] in [test_1], [long] in [test_2]"
  },
  "status": 400
}
```

Since we have already stripped out index names, we can't add index name to exception message.

abbashus: @dai-chen Since we can have multiple indices wit same field name but different mappings, We cannot identify a expected type. Any of them could be expected depending on users case.

dai-chen: > @dai-chen Since we can have multiple indices wit same field name but different mappings, We cannot identify a expected type. Any of them could be expected depending on users case.

Sure. Sorry I meant the type in `fieldMappingValue` and `map.get(fieldName)` which we're comparing.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: > seems you are not using any of the LinkedList features of this map, so just make it a `HashMap`

I created this as LinkedHashMap because all the fieldMappings we generated elsewhere are created this way, and I think it would be better to keep it the same data type to avoid mismatch issue.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
zhongnansu: will fix this and the similar issue in other files. Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: Yes, it can. But I saw many methods that has only one usage within the same class are given public access identifier(even the one I deleted(hasIdenticalMappings), I wonder if you did this on purpose for future use.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
arsen-es: the fact is you use only `put` and `contains`, and for that you don't need any ordering (which is what `LinkedHashMap` adds to `HashMap`)

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
arsen-es: @abbashus The current message says we are "mapping the ambiguities", which is incorrect. So let's fix it (I didn't mean you introduced the bad message).

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
arsen-es: `public` methods are meant to be called from outside, if there is no need for that ,the method should not be `public`. In this case `private` is the best option, since this is just internal implementation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java
zhongnansu: The constructor signature is self-documenting. We are inititalizing a class member using that.

zhongnansu: will fix the name of parameter from map to mapping

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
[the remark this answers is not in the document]
zhongnansu: Good point.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: That makes sense. Fixed.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
zhongnansu: We are not checking equality here. We are checking if the mappings for the same field name are equal or one index has extra mapping not present in others, we should not throw an exception. Does `testMappingsNotDifferent` sounds good?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: changed to mappingSet

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
zhongnansu: Fixed.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
zhongnansu: Good point. Added that as you pointed out.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: That makes sense. Will change to hashMap

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
arsen-es: I think that's what the code here does - check that for the relevant fields the mappings are equal. As I mentioned, the //TODO comment you put on TermFieldRewriter.java#244 is to support the compatibility, but with current change it's just equality. I just want the test name to reflect that.

zhongnansu: I misunderstood you. I thought you want sth like `testIdenticalMapping` which we already have.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: Fixed by wrapping some logic into a method called getMergedMapping

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: Get rid of e using foreach with lambda. For m, I can't think of a better name, any suggestion?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -72,10 +75,11 @@ public boolean visit(MySqlSelectQueryBlock query) {
         Map<String, String> indexToType = new HashMap<>();
         collect(query.getFrom(), indexToType, curScope().getAliases());
         curScope().setMapper(getMappings(indexToType));
-        if ((this.filterType == TermRewriterFilter.COMMA || this.filterType == TermRewriterFilter.MULTI_QUERY)
-            && !hasIdenticalMappings(curScope(), indexToType)) {
-            throw new VerificationException("When using multiple indices, the mappings must be identical.");
+
+        if ((this.filterType == TermRewriterFilter.COMMA || this.filterType == TermRewriterFilter.MULTI_QUERY)) {
```
arsen-es: remove the extra pair of parentheses

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
             );
     }
 
-    public boolean hasIdenticalMappings(TermFieldScope scope, Map<String, String> indexToType) {
+    private void checkMappingCompatibility(TermFieldScope scope, Map<String, String> indexToType) {
         if (scope.getMapper().isEmpty()) {
             throw new VerificationException("Unknown index " + indexToType.keySet());
         }
+        // Collect all FieldMappings into hash set and ignore index/type names. set.size > 1 means FieldMappings NOT unique,
```
arsen-es: I think this comment is redundant, it just tells what the following code does, and that in its turn is pretty straightforward.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+    private void getMergedMapping(String fieldName, Map<String, Object> fieldMappingValue , Map<String, Map<String, Object>> mergedMapping){
+
+        if (!mergedMapping.containsKey(fieldName)) {
+            mergedMapping.put(fieldName, fieldMappingValue);
+        } else {
+            // check if types are same
+            if (!fieldMappingValue.equals(mergedMapping.get(fieldName))) {
+                // TODO: Merge mappings if they are compatible, like text and text/keyword to text/keyword.
+
+                String firstFieldType= new JSONObject(fieldMappingValue).toString().replaceAll("\"", "");
```
arsen-es: missing space before `=`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,53 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+                String secondFieldType = new JSONObject(mergedMapping.get(fieldName)).toString().replaceAll("\"", "");
+
+                String exceptionReason = String.format(Locale.ROOT,"Cannot use field [%s] " +
+                                "due to ambiguities being mapped as incompatible types: [%s] and [%s] ",
```
arsen-es: message still not fixed

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: will fix that, thx.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: will get rid of it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: will fix that, thx.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
abbashus: I don't see any problem with the message. The message is in alignment with xpack exception. What is that you don't understand or seems incorrect to you. If you have good suggestion please provide it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: The todo here is only dealing with one case I met while testing, which is the text(or other type) and text(or other type).keyword. It doesn't affect solving this issue. Sure, will create another issue to track it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
zhongnansu: > message still not fixed

Okay, we will use the followoing meassage as suggested `Different mappings are not allowed for the same field[%s]: found [%s] and [%s]

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -210,19 +214,52 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+        } else {
+            fieldMappings = curScope().getMapper().firstMapping().firstMapping();
+        }
         // We need finalMapping to lookup for rewriting
-        FieldMappings fieldMappings = curScope().getMapper().firstMapping().firstMapping();
         curScope().setFinalMapping(fieldMappings);
-        return true;
+    }
+
+    private void getMergedMapping(String fieldName, Map<String, Object> fieldMappingValue , Map<String, Map<String, Object>> mergedMapping){
```
arsen-es: Sorry, missed this one, if the function name is `getSomething`, then it should return something, maybe `verifySingleFieldMapping` or whatever better name you think of.

zhongnansu: I was also thinking about this. The method is actually constructing the mapping we want to use. It's adding content to the Map `mergedMapping`, which is passed as a parameter.  Does `mergeMapping` sound like a good method name to you?

arsen-es: I think adding a new mapping to that merged map is just a side effect of this method, not the main function, so I am more inclined to assert the verification part in the name.

zhongnansu: Got you. Fixed as `verifySingleFieldMapping`

PR #85, opened 2019-06-19
Getting rid of maven build

arsen-es:
*Issue #, if available:* #17 

*Description of changes:* Migrated the last remaining batch of tests (CSVResultExtractorTest). Got rid of old intgtest classes. Removed the maven build configuration. And some minor code style fixes.

*How tested:* `./gradlew clean build` succeeds


arsen-es:
Apologies for such a big change, but it's mostly deletions.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
```diff
@@ -582,6 +582,69 @@ public void topHitTest_WithExclude() throws IOException {
 
     //region not migrated
 
+    // script on metric aggregation tests. uncomment if your elastic has scripts enable (disabled by default)
```
arsen-es: This batch was not migrated because we didn't have painless module enabled then, I just moved them here so that we don't lose any tests. Me or someone else can update/uncomment later in a separate PR.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -130,10 +142,27 @@ public void selectFieldWithSpace() throws IOException {
         }
     }
 
+    @Ignore("field aliases are not supported currently")
+    // it might be possible to change field names after the query already executed.
+	@Test
```
qreshi: NP: Extra indentation

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -879,10 +922,26 @@ public void complexObjectReturnField() throws IOException {
...
     }
 
+    /**
+     *  TODO: Fields prefixed with @ gets converted to SQLVariantRefExpr instead of SQLIdentifierExpr
+     *  Either change SQLVariantRefExpr to SQLIdentifierExpr
+     *  Or handle the special case for SQLVariantRefExpr
+     */
+    @Ignore
+    @Test
+    public void queryWithATfieldOnWhere() throws IOException {
```
qreshi: NP: queryWith```AtF```ieldOnWhere

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -971,16 +1032,16 @@ public void useScrollWithOrderByAndParams() throws IOException {
         }
     }
 
-    @Ignore // Subquery is still running in ES transport thread which fails the thread assertion as well as tests followed.
+    @Ignore // Subquery still runs in ES transport thread which fails the thread assertion as well as tests followed.
```
qreshi: NP: Should the ending of this comment be reworded to something like ```as well as the following tests```?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -142,6 +156,460 @@ public void fieldOrderWithScriptFields() throws IOException {
...
+        CSVResult csvResult = executeCsvRequest(query, false);
+
+        List<String> headers = csvResult.getHeaders();
+        Assert.assertEquals(2, headers.size());
+        Assert.assertTrue("name should be on headers", headers.contains("name"));
+        Assert.assertTrue("house should be on headers", headers.contains("house"));
+
+        List<String> lines = csvResult.getLines();
+        Assert.assertEquals(7, lines.size());
+        Assert.assertTrue(lines.contains("\"{firstname=Daenerys, lastname=Targaryen, ofHerName=1}\",Targaryen") ||
```
dai-chen: Are we trying to permute them here? If so, I think we better use Guava (`Collections2.permutations()`) to generate permutation. And adding a helper method would be useful. For example,

```
private boolean containsAny(String lines, Collection<String> strs) {} // You can come up with better names than strs :)
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -142,6 +156,460 @@ public void fieldOrderWithScriptFields() throws IOException {
...
+        CSVResult csvResult = executeCsvRequest(query, false);
+
+        List<String> headers = csvResult.getHeaders();
+        Assert.assertEquals(2, headers.size());
+        Assert.assertTrue("name should be on headers", headers.contains("name"));
+        Assert.assertTrue("house should be on headers", headers.contains("house"));
+
+        List<String> lines = csvResult.getLines();
+        Assert.assertEquals(7, lines.size());
+        Assert.assertTrue(lines.contains("{firstname=Daenerys},Targaryen"));
```
dai-chen: Maybe another one `containsAll` for here if agree with my previous comment.

dai-chen: Probably Hamcrest matcher is more readable and can help combine both for your use.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -142,6 +156,460 @@ public void fieldOrderWithScriptFields() throws IOException {
         verifyFieldOrder(expectedFields, query);
     }
 
+    //region Tests migrated from CSVResultsExtractorTests
+
+    @Test
+    public void simpleSearchResultNotNestedNotFlatNoAggs() throws Exception {
+        String query = String.format(Locale.ROOT, "select dog_name,age from %s/dog order by age", TEST_INDEX_DOG);
```
dai-chen: A helper method `format` would be helpful so that we don't have to pass `Locale.ROOT` everywhere and also in case of future change on this.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -142,6 +156,460 @@ public void fieldOrderWithScriptFields() throws IOException {
...
+        final CSVResult csvResult = executeCsvRequest(query, false);
+
+        List<String> headers = csvResult.getHeaders();
+        Assert.assertEquals(2, headers.size());
+        Assert.assertTrue("name should be on headers", headers.contains("dog_name"));
+        Assert.assertTrue("age should be on headers", headers.contains("age"));
+
+        List<String> lines = csvResult.getLines();
+        Assert.assertEquals(2, lines.size());
+        Assert.assertTrue("rex,2".equals(lines.get(0)) || "2,rex".equals(lines.get(0)));
```
dai-chen: I wasn't aware of this. So we may generate columns in random order?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -142,6 +156,460 @@ public void fieldOrderWithScriptFields() throws IOException {
...
+
+    }
+
+    @Test
+    public void extendedStatsAggregationTest() throws Exception {
+        String query = String.format(Locale.ROOT, "SELECT EXTENDED_STATS(age) FROM %s/account", TEST_INDEX_ACCOUNT);
+        CSVResult csvResult = executeCsvRequest(query, false);
+        List<String> headers = csvResult.getHeaders();
+        Assert.assertEquals(8, headers.size());
+        Assert.assertEquals("EXTENDED_STATS(age).count", headers.get(0));
```
dai-chen: Assert equals between 2 Lists is a little cleaner.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -163,4 +631,55 @@ private void setFlatOption(boolean flat) {
...
+
+    private CSVResult executeCsvRequest(final String query, boolean flat) throws IOException {
+
+        return executeCsvRequest(query, flat, false, false, false);
+    }
+
+    private CSVResult executeCsvRequest(final String query, boolean flat, boolean includeScore,
+                                        boolean includeType, boolean includeId) throws IOException {
+
+        final String requestBody = super.makeRequest(query);
```
dai-chen: Try to reuse `SQLIntegTestCase.executeQuery()`? I assume the only difference is extra params in request URL. We can add optional parameters in executeQuery for that. If not true, at lease reuse `SQLIntegTestCase.executeRequest()` for the last few lines.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
```diff
@@ -394,6 +394,11 @@ public static String getResourceFilePath(String relPath) {
     }
 
     public static String getResponseBody(Response response) throws IOException {
```
dai-chen: How do you like the idea that we change all methods in this class to `default` method in an interface? Then we could make `SQLIntegTestCase` implement it ("mixin"). Although mixin in Java is not as powerful as other language, I think we can start using it if you agree this is a good case.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
arsen-es: There were tab (`\t`) characters. Replaced them all with 4 spaces.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
arsen-es: Renamed.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
arsen-es: reworded

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: The method you mentioned is marked as beta and is unstable, we should not use it. I wrote my own getPermutations, will update the tests with that one.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: I think we fixed that bug, but I don't want to run the queries one by one and update the code for those. I know it makes sense, but this is more a migration work, not improving the tests (though I am doing some of that), so if you want we can create an issue and track that separately.

dai-chen: Sounds good!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: makes sense, updated

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: I started with using those two, then realized it's not possible without changing the interfaces of those, and since they are used in too many places, I didn't want to touch them. That would make this PR touch too many files.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
arsen-es: I think I never did that (default implementation in interface in java), so I am hesitant to use it now without properly understanding the intended usage. I will leave it as is at this time, but definitely will learn about that feature and how to properly use it.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: missed this one, will update.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: that will include pretty large number of files. Do you mind if I do it in a separate PR ?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
dai-chen: > I started with using those two, then realized it's not possible without changing the interfaces of those, and since they are used in too many places, I didn't want to touch them. That would make this PR touch too many files.

I don't think you need to "change" the interfaces. Basically you should override it to give the caller option to pass in URL parameters if they want.

PR #88, opened 2019-06-19
Empty line or query without spaces now throws error with more informative message

galkk:
*Issue #, if available:*
*Description of changes: Previously empty string or string without spaces could cause IndexOutOfBoundsException, now the exception and the message will state clearly what was the problem.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
@@ -60,7 +60,9 @@ public class ESActionFactory {
      * @return Query object.
      */
     public static QueryAction create(Client client, String sql) throws SqlParseException, SQLFeatureNotSupportedException {
-        sql = sql.replaceAll(System.lineSeparator()," ").trim();
+
+        // Linebreak matcher
+        sql = sql.replaceAll("\\R"," ").trim();
 
         switch (getFirstWord(sql)) {
             case "SELECT":
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
@@ -60,9 +60,9 @@ public class ESActionFactory {
      * @return Query object.
      */
     public static QueryAction create(Client client, String sql) throws SqlParseException, SQLFeatureNotSupportedException {
-        sql = sql.replaceAll("\n"," ");
-        String firstWord = sql.substring(0, sql.indexOf(' '));
-        switch (firstWord.toUpperCase()) {
+        sql = sql.replaceAll(System.lineSeparator()," ").trim();
+
+        switch (getFirstWord(sql)) {
             case "SELECT":
                 SQLQueryExpr sqlExpr = (SQLQueryExpr) toSqlExpr(sql);
                 sqlExpr.accept(new NestedFieldRewriter());
@@ -98,8 +98,16 @@ public class ESActionFactory {
                 IndexStatement describeStatement = new IndexStatement(StatementType.DESCRIBE, sql);
                 return new DescribeQueryAction(client, describeStatement);
             default:
-                throw new SQLFeatureNotSupportedException(String.format("Unsupported query: %s", sql));
+                throw new SQLFeatureNotSupportedException(
+                        String.format("Query must start with SELECT, DELETE, SHOW or DESCRIBE: %s", sql));
         }
+    }
+
+    private static String getFirstWord(String sql) {
+        int endOfFirstWord = sql.indexOf(' ');
+        return sql
+                .substring(0, endOfFirstWord > 0 ? endOfFirstWord : sql.length())
+                .toUpperCase();
     }
 
     private static boolean isMulti(SQLQueryExpr sqlExpr) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -60,9 +60,9 @@
      * @return Query object.
      */
     public static QueryAction create(Client client, String sql) throws SqlParseException, SQLFeatureNotSupportedException {
-        sql = sql.replaceAll("\n"," ");
-        String firstWord = sql.substring(0, sql.indexOf(' '));
-        switch (firstWord.toUpperCase()) {
+        sql = sql.replaceAll("\n"," ").trim();
```
arsen-es: I think if `sql.isEmpty()` after this statement, there's no need to go into that switch, and we need to throw a different exception, most likely something like invalid query/request (I am sure we already have something like that). `SQLFeatureNotSupportedException` indicates a case where you might have a valid sql statement, but we can't handle it. I think it's better to put a clear distinction between those two cases, at least start with some case.

arsen-es: another note, should we replace `String.format(Locale.ROOT, "%n")` instead of `"\n"`? What if the user sends `"\n\r"`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -98,10 +98,18 @@ else if(isJoin(sqlExpr,sql)){
                 IndexStatement describeStatement = new IndexStatement(StatementType.DESCRIBE, sql);
                 return new DescribeQueryAction(client, describeStatement);
             default:
-                throw new SQLFeatureNotSupportedException(String.format("Unsupported query: %s", sql));
+                throw new SQLFeatureNotSupportedException(
+                        String.format("Query must start with SELECT, DELETE, SHOW or DESCRIBE: %s", sql));
         }
     }
 
+    private static String getFirstWord(String sql) {
```
arsen-es: would be good to somehow document (either a comment, or an assert that the first character is not a space) that there's a precondition to this method : the string is already trimmed. Otherwise calling with a string `" SELECT f1, f2 FROM ..."` will return the whole string.

galkk: This is a private method in a class that has only one entry point and the method is called right on top of it. I feel that your suggestion might be an overkill

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
galkk: 1st comment: I agree with the comment that chosen exception is misleading, I wanted to limit the scope of change. (SqlParseException might be a better choice here).

2nd comment: agree, will change to System.lineSeparator()

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
arsen-es: Sounds good, no objections.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
dai-chen: > 1st comment: I agree with the comment that chosen exception is misleading, I wanted to limit the scope of change. (SqlParseException might be a better choice here).
> 
> 2nd comment: agree, will change to System.lineSeparator()

For 2nd comment, just curious, where/when this separator added to the original query? I assume changing it to `System.lineSeparator()` means use separator of platform our plugin is running on right? I wonder is it necessarily same as the separator in query we want to get rid of?

arsen-es: @dai-chen , that's a good point, those are not necessarily the same, so I think we should just replace all types of new lines.

dai-chen: > @dai-chen , that's a good point, those are not necessarily the same, so I think we should just replace all types of new lines.

Yeah, that was my concern.

dai-chen: @galkk does this make sense to you?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -60,9 +60,9 @@
      * @return Query object.
      */
     public static QueryAction create(Client client, String sql) throws SqlParseException, SQLFeatureNotSupportedException {
-        sql = sql.replaceAll("\n"," ");
-        String firstWord = sql.substring(0, sql.indexOf(' '));
-        switch (firstWord.toUpperCase()) {
+        sql = sql.replaceAll(System.lineSeparator()," ").trim();
```
arsen-es: as remarked by @dai-chen , the user input might not necessarily happen on the same system. Even though this is not ideal, but I suggest doing this to account for both Windows- and Unix-style new lines:
```
sql = sql.replaceAll("\r\n", " ").replaceAll("\n", " ").trim();
```

galkk: There's a regex that is called "linebreak matcher", I'll use it

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
galkk: Done

arsen-es:
This PR was to fix issue #93

PR #89, opened 2019-06-20
Fix improper handling of the table expression

galkk:
*Description of changes:* 
The wrong query, such as "select * from abc like 'zzz'" was throwing server error instead of ParserError, fixing that.

Introduced variables and simplified some very long casts for easier debugging

Testing: unit tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
@@ -167,8 +167,8 @@ public class TermFieldRewriter extends MySqlASTVisitorAdapter {
                     SQLIdentifierExpr right = (SQLIdentifierExpr) rightSideOfExpression;
                     indexToType.put(table, right.getName());
                 } else {
-                    throw new ParserException("Right side of the expression  " + rightSideOfExpression.toString() +
-                            " is expected to be an identifier");
+                    throw new ParserException("Right side of the expression [" + rightSideOfExpression.toString() +
+                            "] is expected to be an identifier");
                 }
             }
             if (tableSource.getAlias() != null) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
@@ -150,31 +150,31 @@ public class TermFieldRewriter extends MySqlASTVisitorAdapter {
     public void collect(SQLTableSource tableSource, Map<String, String> indexToType,  Map<String, String> aliases) {
         if (tableSource instanceof SQLExprTableSource) {
 
-            String table = null;
+            String tableName = null;
             SQLExprTableSource sqlExprTableSource = (SQLExprTableSource) tableSource;
 
             if (sqlExprTableSource.getExpr() instanceof SQLIdentifierExpr) {
                 SQLIdentifierExpr sqlIdentifier = (SQLIdentifierExpr) sqlExprTableSource.getExpr();
-                table = sqlIdentifier.getName();
-                indexToType.put(table, null);
+                tableName = sqlIdentifier.getName();
+                indexToType.put(tableName, null);
             } else if (sqlExprTableSource.getExpr() instanceof SQLBinaryOpExpr) {
                 SQLBinaryOpExpr sqlBinaryOpExpr = (SQLBinaryOpExpr) sqlExprTableSource.getExpr();
-                table = ((SQLIdentifierExpr) sqlBinaryOpExpr.getLeft()).getName();
+                tableName = ((SQLIdentifierExpr) sqlBinaryOpExpr.getLeft()).getName();
                 SQLExpr rightSideOfExpression = sqlBinaryOpExpr.getRight();
 
                 // This assumes that right side of the expression is different name in query
                 if (rightSideOfExpression instanceof SQLIdentifierExpr) {
                     SQLIdentifierExpr right = (SQLIdentifierExpr) rightSideOfExpression;
-                    indexToType.put(table, right.getName());
+                    indexToType.put(tableName, right.getName());
                 } else {
                     throw new ParserException("Right side of the expression [" + rightSideOfExpression.toString() +
                             "] is expected to be an identifier");
                 }
             }
             if (tableSource.getAlias() != null) {
-                aliases.put(tableSource.getAlias(), table);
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -147,19 +149,27 @@ public boolean visit(SQLIdentifierExpr expr) {
...
+            SQLExprTableSource sqlExprTableSource = (SQLExprTableSource) tableSource;
+
+            if (sqlExprTableSource.getExpr() instanceof SQLIdentifierExpr) {
+                SQLIdentifierExpr sqlIdentifier = (SQLIdentifierExpr) sqlExprTableSource.getExpr();
+                table = sqlIdentifier.getName();
+                indexToType.put(table, null);
+            } else if (sqlExprTableSource.getExpr() instanceof SQLBinaryOpExpr) {
+                SQLBinaryOpExpr sqlBinaryOpExpr = (SQLBinaryOpExpr) sqlExprTableSource.getExpr();
+                table = ((SQLIdentifierExpr) sqlBinaryOpExpr.getLeft()).getName();
+                SQLExpr rightSideOfExpression = sqlBinaryOpExpr.getRight();
```
arsen-es: np: `rightTerm`

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -147,19 +149,27 @@ public boolean visit(SQLIdentifierExpr expr) {
...
+
+                // This assumes that right side of the expression is different name in query
+                if (rightSideOfExpression instanceof SQLIdentifierExpr) {
+                    SQLIdentifierExpr right = (SQLIdentifierExpr) rightSideOfExpression;
+                    indexToType.put(table, right.getName());
+                } else {
+                    throw new ParserException("Right side of the expression  " + rightSideOfExpression.toString() +
```
arsen-es: add some delimiters to distinguish between the exception message and the actual expression value, e.g. `"... expression [" + value + "] is expected ..."`. Also I am a fan of using the [rfc2119](https://tools.ietf.org/html/rfc2119) terms, so in this case **must**.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/bugs/BuggedQueries.java
```diff
@@ -0,0 +1,36 @@
...
+import org.elasticsearch.client.Client;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.sql.SQLFeatureNotSupportedException;
+
+/**
+ * Special cases for queries that had broken our parser in past.
+ */
+public class BuggedQueries {
```
arsen-es: `bugged` means something unrelated to software bugs. Also , I think it might be confusing to have `bugs` in the package name. How about calling this class `RegressionTest` and the package name `*.unittests.regression`

galkk: Good suggestion, done

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/bugs/BuggedQueries.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.unittest.bugs;
```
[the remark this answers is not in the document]
galkk: The purpose of this package is regression tests, so I'll rename the package to "regressions"

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
galkk: As we can see, on the right side there can be lots of different things (SQLIdentifier, SQLCharExpression), so I'd like to keep it as is. If you insist, though, I'll rename

arsen-es: no objections, that's why it's `np`

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
galkk: We are not sure if there are no other cases, so as of now I want to avoid "must"

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -147,19 +149,27 @@ public boolean visit(SQLIdentifierExpr expr) {
 
     public void collect(SQLTableSource tableSource, Map<String, String> indexToType,  Map<String, String> aliases) {
         if (tableSource instanceof SQLExprTableSource) {
+
             String table = null;
```
abbashus: np: Should this be tableName?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
arsen-es: ok, makes sense.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
abbashus: Apologies. Lets keep it `table` as before

arsen-es: I think your comment was pretty reasonable, since it's not the table object, but the name of the table. Naming is very important since it's one of the first if not the first thing that affects readability.

arsen-es:
This PR was to fix issue #92

Issue #95, opened 2019-06-24
NPE thrown when selecting all in query with nested fields involved


dai-chen:
When trying to select all fields, no matter what syntax in use, NullPointerException thrown somewhere (no error stack logged) and ends up responding 500 server error to customer.

Related issue raised in JDBC client repo: https://github.com/opendistro-for-elasticsearch/sql-jdbc/issues/15. Please refer to the mapping and query there to reproduce the issue.

dai-chen:
It seems that JDBC format works well for nested field query except nested field has type of text with keyword. For example, it is working in the following scenario:

```
PUT /nested_obj
{
  "mappings": {
      "properties": {
        "someField": {
          "type": "keyword"
        },
        "myNum": {
          "type": "long"
        },
        "message": {
          "type": "nested",
          "properties": {
            "info": {
              "type": "keyword"
            },
            "author": {
              "type": "keyword"
            },
            "dayOfWeek": {
              "type": "long"
            }
          }
        },
        "comment": {
          "type": "nested",
          "properties": {
            "data": {
              "type": "keyword"
            },
            "likes": {
              "type": "long"
            }
          }
        }        
      }
  },
  "settings":{
      "index":{
          "number_of_shards":1,
          "number_of_replicas":1
      }
  }
}

PUT nested_obj/_doc/1
{"message": [{"info":"zz","author":"zz","dayOfWeek":6}],"comment":{"data":["aa","bb"],"likes":10},"myNum":[3,4],"someField":"a"}

POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT * FROM nested_obj n, n.message m"
}
```

Debugging for root cause.

PR #98, opened 2019-06-27
Fix Issue for SELECT * in Nested Field Query

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/95

*Description of changes:* The meaning of SELECT * in nested field query was not clear. To fix issue, we need to clarify its semantics.

For examples, in `SELECT * FROM A a, a.B b` and `SELECT * FROM A a, a.B b, a.C c` (suppose B and C are nested fields in A), regulars fields of A and nested fields in B should be returned for the first query. In addition, nested fields in C should be returned for the second query. **Basically, * just means regular fields. FROM clause decides if any nested fields should be flatten and present in result**. 

One exception is `SELECT * FROM A`. All fields are returned including nested fields in this case. Because we'd like to trade off consistency of SELECT * notation for more intuition from customer perspective.

Please find more details in my integration test.

A proposal is to switch current metadata query to mappings in `LocalClusterState`. Because mappings are being flatten and hierarchy is missing in the response of current query. Ideally, we should build domain object layer on top and avoid code like `fieldName.contains(".")`.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
@@ -68,6 +68,7 @@ public class Select extends Query {
 
 	public void addGroupBy(List<Field> fields) {
 		isAgg = true;
+		selectAll = false;
 		this.groupBys.add(fields);
 	}
 
@@ -107,8 +108,9 @@ public class Select extends Query {
 		if (field == null ) {
 			return;
 		}
-        if(field.getName().equals("*")){
+        if (field.getName().equals("*") && !isAgg) { // Ignore * for GROUP BY
             this.selectAll = true;
+            return;
         }
 
 		if(field instanceof  MethodField && aggsFunctions.contains(field.getName().toUpperCase())) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import static com.amazon.opendistroforelasticsearch.sql.domain.Field.STAR;
+
 /**
  * 将sql语句转换为select 对象
  * 
@@ -108,7 +110,7 @@ public class Select extends Query {
 		if (field == null ) {
 			return;
 		}
-        if (field.getName().equals("*") && !isAgg) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
+        if (field == STAR && !isAgg) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
             this.selectAll = true;
             return;
         }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
@@ -17,10 +17,11 @@ package com.amazon.opendistroforelasticsearch.sql.domain;
 
 import com.amazon.opendistroforelasticsearch.sql.domain.hints.Hint;
 import com.amazon.opendistroforelasticsearch.sql.parser.SubQueryExpression;
+import com.google.common.collect.ImmutableSet;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
 
 import static com.amazon.opendistroforelasticsearch.sql.domain.Field.STAR;
 
@@ -31,8 +32,14 @@ import static com.amazon.opendistroforelasticsearch.sql.domain.Field.STAR;
  */
 public class Select extends Query {
 
-    // Using this functions, will cause query to execute as aggregation.
-    private final List<String> aggsFunctions = Arrays.asList("SUM", "MAX", "MIN", "AVG", "TOPHITS", "COUNT", "STATS","EXTENDED_STATS","PERCENTILES","SCRIPTED_METRIC");
+    /** Using this functions will cause query to execute as aggregation. */
+    private final static Set<String> AGGREGATE_FUNCTIONS =
+        ImmutableSet.of(
+            "SUM", "MAX", "MIN", "AVG",
+            "TOPHITS", "COUNT", "STATS", "EXTENDED_STATS",
+            "PERCENTILES", "SCRIPTED_METRIC"
+        );
+
     private List<Hint> hints = new ArrayList<>();
     private List<Field> fields = new ArrayList<>();
     private List<List<Field>> groupBys = new ArrayList<>();
@@ -45,7 +52,7 @@ public class Select extends Query {
     public boolean isQuery = false;
     private boolean selectAll = false;
 
-    public boolean isAgg = false;
+    public boolean isAggregate = false;
 
     public Select() {
     }
@@ -69,7 +76,7 @@ public class Select extends Query {
     }
 
     public void addGroupBy(List<Field> fields) {
-        isAgg = true;
+        isAggregate = true;
         selectAll = false;
         this.groupBys.add(fields);
     }
@@ -107,16 +114,16 @@ public class Select extends Query {
 
 
     public void addField(Field field) {
-        if (field == null ) {
+        if (field == null) {
             return;
         }
-        if (field == STAR && !isAgg) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
+        if (field == STAR && !isAggregate) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
             this.selectAll = true;
             return;
         }
 
-        if(field instanceof  MethodField && aggsFunctions.contains(field.getName().toUpperCase())) {
-            isAgg = true;
+        if (field instanceof MethodField && AGGREGATE_FUNCTIONS.contains(field.getName().toUpperCase())) {
+            isAggregate = true;
         }
 
         fields.add(field);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
@@ -108,7 +108,7 @@ public class Select extends Query {
 		if (field == null ) {
 			return;
 		}
-        if (field.getName().equals("*") && !isAgg) { // Ignore * for GROUP BY
+        if (field.getName().equals("*") && !isAgg) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
             this.selectAll = true;
             return;
         }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
@@ -31,52 +31,52 @@ import static com.amazon.opendistroforelasticsearch.sql.domain.Field.STAR;
  */
 public class Select extends Query {
 
-	// Using this functions, will cause query to execute as aggregation.
-	private final List<String> aggsFunctions = Arrays.asList("SUM", "MAX", "MIN", "AVG", "TOPHITS", "COUNT", "STATS","EXTENDED_STATS","PERCENTILES","SCRIPTED_METRIC");
+    // Using this functions, will cause query to execute as aggregation.
+    private final List<String> aggsFunctions = Arrays.asList("SUM", "MAX", "MIN", "AVG", "TOPHITS", "COUNT", "STATS","EXTENDED_STATS","PERCENTILES","SCRIPTED_METRIC");
     private List<Hint> hints = new ArrayList<>();
-	private List<Field> fields = new ArrayList<>();
-	private List<List<Field>> groupBys = new ArrayList<>();
-	private Having having;
-	private List<Order> orderBys = new ArrayList<>();
-	private int offset;
-	private int rowCount = 200;
+    private List<Field> fields = new ArrayList<>();
+    private List<List<Field>> groupBys = new ArrayList<>();
+    private Having having;
+    private List<Order> orderBys = new ArrayList<>();
+    private int offset;
+    private int rowCount = 200;
     private boolean containsSubQueries;
     private List<SubQueryExpression> subQueries;
-	public boolean isQuery = false;
+    public boolean isQuery = false;
     private boolean selectAll = false;
 
-	public boolean isAgg = false;
+    public boolean isAgg = false;
 
-	public Select() {
-	}
+    public Select() {
+    }
 
-	public List<Field> getFields() {
-		return fields;
-	}
+    public List<Field> getFields() {
+        return fields;
+    }
 
-	public void setOffset(int offset) {
-		this.offset = offset;
-	}
+    public void setOffset(int offset) {
+        this.offset = offset;
+    }
 
-	public void setRowCount(int rowCount) {
-		this.rowCount = rowCount;
-	}
+    public void setRowCount(int rowCount) {
+        this.rowCount = rowCount;
+    }
 
-	public void addGroupBy(Field field) {
-		List<Field> wrapper = new ArrayList<>();
-		wrapper.add(field);
-		addGroupBy(wrapper);
-	}
+    public void addGroupBy(Field field) {
+        List<Field> wrapper = new ArrayList<>();
+        wrapper.add(field);
+        addGroupBy(wrapper);
+    }
 
-	public void addGroupBy(List<Field> fields) {
-		isAgg = true;
-		selectAll = false;
-		this.groupBys.add(fields);
-	}
+    public void addGroupBy(List<Field> fields) {
+        isAgg = true;
+        selectAll = false;
+        this.groupBys.add(fields);
+    }
 
-	public List<List<Field>> getGroupBys() {
-		return groupBys;
-	}
+    public List<List<Field>> getGroupBys() {
+        return groupBys;
+    }
 
     public Having getHaving() {
         return having;
@@ -86,41 +86,41 @@ public class Select extends Query {
         this.having = having;
     }
 
-	public List<Order> getOrderBys() {
-		return orderBys;
-	}
+    public List<Order> getOrderBys() {
+        return orderBys;
+    }
 
-	public int getOffset() {
-		return offset;
-	}
+    public int getOffset() {
+        return offset;
+    }
 
-	public int getRowCount() {
-		return rowCount;
-	}
+    public int getRowCount() {
+        return rowCount;
+    }
 
-	public void addOrderBy(String nestedPath, String name, String type, boolean isScriptField) {
-		if ("_score".equals(name)) {
-			isQuery = true;
-		}
-		this.orderBys.add(new Order(nestedPath, name, type, isScriptField));
-	}
+    public void addOrderBy(String nestedPath, String name, String type, boolean isScriptField) {
+        if ("_score".equals(name)) {
+            isQuery = true;
+        }
+        this.orderBys.add(new Order(nestedPath, name, type, isScriptField));
+    }
 
 
-	public void addField(Field field) {
-		if (field == null ) {
-			return;
-		}
+    public void addField(Field field) {
+        if (field == null ) {
+            return;
+        }
         if (field == STAR && !isAgg) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
             this.selectAll = true;
             return;
         }
 
-		if(field instanceof  MethodField && aggsFunctions.contains(field.getName().toUpperCase())) {
-			isAgg = true;
-		}
+        if(field instanceof  MethodField && aggsFunctions.contains(field.getName().toUpperCase())) {
+            isAgg = true;
+        }
 
-		fields.add(field);
-	}
+        fields.add(field);
+    }
 
     public List<Hint> getHints() {
         return hints;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -31,7 +31,7 @@ import static org.elasticsearch.action.admin.indices.mapping.get.GetFieldMapping
  */
 public class FieldMapping {
 
-    /** Field (name) to be parsed */
+    /** Name of the Field to be parsed */
     private final String fieldName;
 
     /** Native mapping information returned from ES */
@@ -118,7 +118,7 @@ public class FieldMapping {
          * When it is nested or contains "." in general (ex. fieldName.nestedName) the source is nestedName -> type
          */
         Map<String, Object> fieldMapping;
-        if (fieldPath.length < 2) {
+        if (fieldPath.length == 1) {
             fieldMapping = (Map<String, Object>) source.get(fieldName);
         } else {
             fieldMapping = (Map<String, Object>) source.get(fieldPath[1]);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -26,7 +26,7 @@ import static org.elasticsearch.action.admin.indices.mapping.get.GetFieldMapping
 /**
  * Field mapping that parse native ES mapping.
  *
- * NOTE that approaches in this class is NOT reliable because of the ES mapping query API used.
+ * NOTE that approaches in this class are NOT reliable because of the ES mapping query API used.
  * We should deprecate this in future and parse field mapping in more solid way.
  */
 public class FieldMapping {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -63,29 +63,20 @@ public class FieldMapping {
     }
 
     /**
+     * Verify if property field matches wildcard pattern specified in query
+     * @return true if matched
+     */
+    public boolean isWildcardSpecified() {
+        return specifiedFieldsByName.containsKey(path() + ".*");
+    }
+
+    /**
      * Is field a property field, which means either object field or nested field.
      * @return true for property field
      */
     public boolean isPropertyField() {
         int numOfDots = StringUtils.countMatches(fieldName, '.');
         return numOfDots > 1 || (numOfDots == 1 && !isMultiField());
-    }
-
-    /**
-     * Path of property field, for example "employee" in "employee.manager"
-     * @return path of property field
-     */
-    public String path() {
-        int lastDot = fieldName.lastIndexOf(".");
-        return fieldName.substring(0, lastDot);
-    }
-
-    /**
-     * Verify if property field matches wildcard pattern specified in query
-     * @return true if matched
-     */
-    public boolean isWildcardSpecified() {
-        return specifiedFieldsByName.containsKey(path() + ".*");
     }
 
     /**
@@ -102,6 +93,15 @@ public class FieldMapping {
      */
     public boolean isMetaField() {
         return fieldName.startsWith("_");
+    }
+
+    /**
+     * Path of property field, for example "employee" in "employee.manager"
+     * @return path of property field
+     */
+    public String path() {
+        int lastDot = fieldName.lastIndexOf(".");
+        return fieldName.substring(0, lastDot);
     }
 
     /**
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -117,13 +117,7 @@ public class FieldMapping {
          * When field is not nested the metaData source is fieldName -> type
          * When it is nested or contains "." in general (ex. fieldName.nestedName) the source is nestedName -> type
          */
-        Map<String, Object> fieldMapping;
-        if (fieldPath.length == 1) {
-            fieldMapping = (Map<String, Object>) source.get(fieldName);
-        } else {
-            fieldMapping = (Map<String, Object>) source.get(fieldPath[1]);
-        }
-
+        Map<String, Object> fieldMapping = (Map<String, Object>) source.get(fieldPath.length == 1 ? fieldName : fieldPath[1]);
         for (int i = 2; i < fieldPath.length; i++) {
             fieldMapping = (Map<String, Object>) fieldMapping.get(fieldPath[i]);
         }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -16,7 +16,7 @@
 package com.amazon.opendistroforelasticsearch.sql.esdomain.mapping;
 
 import com.amazon.opendistroforelasticsearch.sql.domain.Field;
-import com.google.common.base.CharMatcher;
+import com.amazon.opendistroforelasticsearch.sql.utils.StringUtils;
 
 import java.util.Map;
 
@@ -67,7 +67,7 @@ public class FieldMapping {
      * @return true for property field
      */
     public boolean isPropertyField() {
-        int numOfDots = CharMatcher.is('.').countIn(fieldName); // Move to our own StringUtils
+        int numOfDots = StringUtils.countMatches(fieldName, '.');
         return numOfDots > 1 || (numOfDots == 1 && !isMultiField());
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -40,7 +40,6 @@ public class FieldMapping {
     /** Maps a field name to Field object that specified in query explicitly */
     private final Map<String, Field> specifiedFieldsByName;
 
-
     public FieldMapping(String fieldName) {
         this(fieldName, emptyMap(), emptyMap());
     }
@@ -102,7 +101,7 @@ public class FieldMapping {
     public String path() {
         int lastDot = fieldName.lastIndexOf(".");
         if (lastDot == -1) {
-            throw new IllegalStateException("path() is being invoked on wrong field");
+            throw new IllegalStateException("path() is being invoked on the wrong field [" + fieldName + "]");
         }
         return fieldName.substring(0, lastDot);
     }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
@@ -101,6 +101,9 @@ public class FieldMapping {
      */
     public String path() {
         int lastDot = fieldName.lastIndexOf(".");
+        if (lastDot == -1) {
+            throw new IllegalStateException("path() is being invoked on wrong field");
+        }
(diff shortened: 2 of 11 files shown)


dai-chen:
Found another issue when I test this in DbVisualizer with JDBC client. It's fine to return whatever for column of `TEXT/KEYWORD` type. For example, client side will convert and visualize the array of JSON object for message in 2nd row. But it's unable to handle array value for myNum in 1st row because it's `LONG`.

```
{
  "schema": [
    {
      "name": "myNum",
      "type": "long"
    },
    {
      "name": "comment",
      "alias": "",
      "type": "text"
    },
    {
      "name": "message",
      "alias": "",
      "type": "text"
    }
  ],
  "total": 2,
  "datarows": [
    [
      [
        3,
        4
      ],
      {
        "data": [
          "aa",
          "bb"
        ],
        "likes": 10
      },
      [{
        "author": "zz",
        "dayOfWeek": 6,
        "info": "zz"
      }]
    ],
    [
      4,
      {
        "data": "ab",
        "likes": 1
      },
      [
        {
          "author": "h",
          "dayOfWeek": 4,
          "info": "c"
        },
        {
          "author": "i",
          "dayOfWeek": 5,
          "info": "a"
        }
      ]
    ]
  ]
}
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -366,9 +379,55 @@ private String getFieldName(Field field) {
...
+        return fieldMap.containsKey(fieldName);
+    }
+
+    /**
+     *  Verify if field is property field (object or nested ) and matched by wildcard pattern given in SELECT
+     *  A special case is text field with nested keyword. Ignore it in the case of SELECT * ?
+     */
+    private boolean isFieldPropertyAndMatchWildcard(Map<String, Field> fieldMap, String fieldName) {
+        int lastDot = fieldName.lastIndexOf(".");
+        if (lastDot > -1) {
```
arsen-es: np: I think the contract of `lastIndexOf` is it returns -1 if not found, and the index if found, so I would just check that it's not -1, i.e. `lastDot != -1`

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -366,9 +379,55 @@ private String getFieldName(Field field) {
...
+            String path = fieldName.substring(0, lastDot);
+            return !fieldName.endsWith(".keyword") && fieldMap.containsKey(path + ".*");
```
arsen-es: just curious, does es prohibit having a field with name **keyword** ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: > np: I think the contract of `lastIndexOf` is it returns -1 if not found, and the index if found, so I would just check that it's not -1, i.e. `lastDot != -1`

Yes, it is. I thought I was using `indexOf`. Will fix it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: > just curious, does es prohibit having a field with name **keyword** ?

I'm not sure about a regular field. But in this case, `endsWith(".keyword")` is not always the case. ES will do this automatically if you don't specify it beforehand. But customer can do this themselves with any other name. 

For example, "keyword" (not the one after "type") could be anything.

```
            "author": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above" : 256
                }
              }
            }
```

That is why the proposal in the Description of PR to build domain layer on top of another ES mapping query API.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
```diff
@@ -241,6 +241,12 @@ public static String getNestedTypeIndexMapping() {
                 "              },\n" +
                 "              \"author\": {\n" +
                 "                \"type\": \"keyword\",\n" +
+                "                \"fields\": {\n" +
```
galkk: Just observation: I think that we should move index definitions outside of codebase into file system, like it's done with test data (e.g. people2.json and people2.index.json), that will simplify working with those files and importing them into elasticsearch for offline testng

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PrettyFormatResponseIT.java
```diff
@@ -60,6 +64,8 @@
     private static final Set<String> messageFields = Sets.newHashSet(
             "message.dayOfWeek", "message.info", "message.author");
 
+    private static final Set<String> commentFields = Sets.newHashSet("comment.data", "comment.likes");
```
galkk: Set.of will create unmodificable set (same for rows above and below).

https://docs.oracle.com/en/java/javase/11/core/creating-immutable-lists-sets-and-maps.html

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -46,6 +46,7 @@ public static Field makeField(SQLExpr expr, String alias, String tableAlias) thr
             return makeField(makeBinaryMethodField((SQLBinaryOpExpr) expr, alias, true), alias, tableAlias);
 
         } else if (expr instanceof SQLAllColumnExpr) {
+            return new Field("*", "");
```
galkk: Constant for "*"? Maybe even special field type?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -366,9 +379,55 @@ private String getFieldName(Field field) {
...
+    }
+
+    /**
+     * Special case which trades off consistency of SELECT * meaning for more intuition from customer perspective.
+     * In other cases, * means all regular fields on the level.
+     * The only exception here is * picks all non-regular (nested) fields as JSON without flatten.
+     */
+    private void populateAllNestedFields(List<Schema.Column> columns, List<String> fields) {
+        Set<String> nestedFieldNames = fields.stream().
+                                       filter(f -> f.contains(".") && !f.endsWith("keyword")).
```
galkk: could you introduce named constants and add more comments? For example - the condition here is endsWith("keyword") and the condition in line 401 is endsWith(".keyword"). That is subtle difference that could be explained by the comment or constant name.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
dai-chen: > Just observation: I think that we should move index definitions outside of codebase into file system, like it's done with test data (e.g. people2.json and people2.index.json), that will simplify working with those files and importing them into elasticsearch for offline testng

Agree, especially there is no support for multiline string literal in Java at the moment (maybe available in JDK 13?). We can create another issue to improve this (probably as well as a script/gradle task to import our test mapping/data).

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: > could you introduce named constants and add more comments? For example - the condition here is endsWith("keyword") and the condition in line 401 is endsWith(".keyword"). That is subtle difference that could be explained by the comment or constant name.

Sorry, my mistake. It's suppose to be ".keyword" here too. Another example to support my proposal that build more solid domain layer rather than do this kind of tricky check. :) Will fix it.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PrettyFormatResponseIT.java
dai-chen: > Set.of will create unmodificable set (same for rows above and below).
> 
> https://docs.oracle.com/en/java/javase/11/core/creating-immutable-lists-sets-and-maps.html

Yeah, I'd love to use those new features too. Unfortunately I recall we use JDK 12 for ES build tools but still need to be compatible with JDK 8 in our code: https://github.com/opendistro-for-elasticsearch/sql/blob/master/build.gradle#L78

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
dai-chen: Sure, will add constant for now since I don't see different behavior worth creating a new domain class. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -92,7 +93,7 @@ private void loadFromEsState(Query query) {
         String typeName = fetchTypeName(query);
         String[] fieldNames = fetchFieldsAsArray(query);
 
-        if (fieldNames.length == 0 && !fieldsSelectedOnAnotherTable(query))
+        if (isSimpleQuerySelectAll(query) || isJoinQuerySelectAll(query, fieldNames))
```
arsen-es: please change to
```java
selectAll = isSimpleQuerySelectAll(query) || isJoinQuerySelectAll(query, fieldNames);
```

In general, let's keep consistent style and use curly braces even if the block will have 1 line.

dai-chen: > please change to
> 
> ```java
> selectAll = isSimpleQuerySelectAll(query) || isJoinQuerySelectAll(query, fieldNames);
> ```
> 
> In general, let's keep consistent style and use curly braces even if the block will have 1 line.

Sure, will fix.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
```diff
@@ -80,6 +82,17 @@ public static String toUpper(final String input) {
...
 
+    /**
+     * Count how many occurrences of character in this input {@code Sequence}.
+     *
+     * @param input  the input string
+     * @param match  char to be matched
+     * @return       number of occurrences
+     */
+    public static int countMatches(CharSequence input, char match) {
+        return CharMatcher.is(match).countIn(input);
```
arsen-es: To avoid using 3rd party libraries, with Java8 you can do:
```java
input.chars().filter(c -> c == match).count();
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
```diff
@@ -107,8 +110,9 @@ public void addField(Field field) {
 		if (field == null ) {
```
arsen-es: Can you please replace all the tab characters with 4 spaces each throughout this file before you push the change?

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+import com.amazon.opendistroforelasticsearch.sql.domain.Field;
+import com.amazon.opendistroforelasticsearch.sql.utils.StringUtils;
+
+import java.util.Map;
+
+import static java.util.Collections.emptyMap;
+import static org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsResponse.FieldMappingMetaData;
+
+/**
+ * Field mapping that parse native ES mapping.
```
arsen-es: parse`s`

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public class FieldMapping {
+
+    /** Field (name) to be parsed */
+    private final String fieldName;
+
+    /** Native mapping information returned from ES */
+    private final Map<String, FieldMappingMetaData> typeMappings;
+
+    /** Map of field name to Field object that specified in query explicitly */
```
arsen-es: np: `Maps a field name to the Field object explicitly specified in the query.`

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+    private final Map<String, Field> specifiedFieldByNames;
```
arsen-es: `specifiedFieldsByName`~s~

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+        this.fieldName = fieldName;
+        this.typeMappings = typeMappings;
+        this.specifiedFieldByNames = specifiedFieldByNames;
+    }
+
+    /**
+     * Is field specified explicitly in query
+     * @return true if specified
+     */
+    public boolean isSpecified() {
```
arsen-es: maybe too verbose, but I think you could just name this `isSpecifiedInQuery` and not add any javadoc at all.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+    public boolean isWildcardSpecified() {
+        return specifiedFieldByNames.containsKey(path() + ".*");
+    }
+
+    /**
+     * Is field a/in multi-field, for example, field "a.keyword" in field "a"
+     * @return true for multi field
+     */
+    public boolean isMultiField() {
+        return fieldName.endsWith(".keyword");
```
arsen-es: I am thinking if there is a better way to identify a multi-field. Because the name of the extra field can be anything else, not just `keyword`. So either let's rename this method if we only care about `.keyword` case, or find a better way.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+        FieldMappingMetaData metaData = typeMappings.get(fieldName);
+        Map<String, Object> source = metaData.sourceAsMap();
+        String[] fieldPath = fieldName.split("\\.");
+
+        /*
+         * When field is not nested the metaData source is fieldName -> type
+         * When it is nested or contains "." in general (ex. fieldName.nestedName) the source is nestedName -> type
+         */
+        Map<String, Object> fieldMapping;
+        if (fieldPath.length < 2) {
```
arsen-es: `== 1` ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+            fieldMapping = (Map<String, Object>) source.get(fieldName);
+        } else {
+            fieldMapping = (Map<String, Object>) source.get(fieldPath[1]);
```
arsen-es: `source.get(fieldPath[fieldPath.length - 1])` to make sure this works even when we support multiple levels of nesting.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        for (int i = 2; i < fieldPath.length; i++) {
```
arsen-es: I am a little confused, what is this code for? I thought we consider only paths with 2 terms, otherwise, as I guessed from your earlier comment (line 118), the fieldName to type mapping would use the most nested path term.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -305,10 +319,10 @@ private String getFieldName(Field field) {
          */
         Map<String, Field> fieldMap = fetchFieldMap(query);
         List<Schema.Column> columns = new ArrayList<>();
-        for (String field : fields) {
+        for (String fieldName : fields) {
```
arsen-es: np: to be consistent would be good to rename `fields` to `fieldNames`

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
dai-chen: > To avoid using 3rd party libraries, with Java8 you can do:
> 
> ```java
> input.chars().filter(c -> c == match).count();
> ```

Good point. Will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
dai-chen: Sure. We can also do this for all other files once so we don't need to add comment for this in each PR.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: yeah, it is clear but a little verbose because there is another much longer should be renamed too, `isWildcardSpecifiedInQuery`. Will rename both if you feel okay.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: Sure, will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: I recalled wrong the naming convention of Map. Will fix.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: > I am thinking if there is a better way to identify a multi-field. Because the name of the extra field can be anything else, not just `keyword`. So either let's rename this method if we only care about `.keyword` case, or find a better way.

I'd like to keep this as it is because this is the intention of the API (although our approach is bad). For now, it's hard to figure this out because the mapping returned from ES is "flat". I'm thinking of integrating this domain with the existing one `FieldMappings`(https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java#L392) which depends on different format of ES mapping.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: I just found there is an input argument taken the name. Will rename `fields` to fieldNameList.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
dai-chen: `count()` returns long, will truncate it to int since it's not useful and worth in our case.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: I'm not original author but I think that's because you can have multi-field in nested field, ex. "a.b.keyword".

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: I guess other code may break even if this works...

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
arsen-es: sounds good

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
arsen-es: sorry for nitpicking, but you could do
```
fieldMapping = (Map<String, Object>) source.get(fieldPath.length < 2 ? fieldName : fieldPath[1]);
```
to have no duplicated code.

dai-chen: > sorry for nitpicking, but you could do
> 
> ```
> fieldMapping = (Map<String, Object>) source.get(fieldPath.length < 2 ? fieldName : fieldPath[1]);
> ```
> 
> to have no duplicated code.

Sure, will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
```diff
@@ -22,58 +22,61 @@
...
  * 将sql语句转换为select 对象
  * 
  * @author ansj
  */
 public class Select extends Query {
 
-	// Using this functions, will cause query to execute as aggregation.
-	private final List<String> aggsFunctions = Arrays.asList("SUM", "MAX", "MIN", "AVG", "TOPHITS", "COUNT", "STATS","EXTENDED_STATS","PERCENTILES","SCRIPTED_METRIC");
+    // Using this functions, will cause query to execute as aggregation.
+    private final List<String> aggsFunctions = Arrays.asList("SUM", "MAX", "MIN", "AVG", "TOPHITS", "COUNT", "STATS","EXTENDED_STATS","PERCENTILES","SCRIPTED_METRIC");
```
galkk: formatting: missing spaces between parameters in second line.

consider renaming to aggregateFunctions

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
```diff
@@ -83,40 +86,41 @@ public void setHaving(Having having) {
...
+        }
+        if (field == STAR && !isAgg) { // Ignore GROUP BY since columns present in result are decided by column list in GROUP BY
             this.selectAll = true;
+            return;
         }
 
-		if(field instanceof  MethodField && aggsFunctions.contains(field.getName().toUpperCase())) {
-			isAgg = true;
-		}
+        if(field instanceof  MethodField && aggsFunctions.contains(field.getName().toUpperCase())) {
```
galkk: formatting: missing space between if and (

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
```diff
@@ -22,58 +22,61 @@
...
+    private int offset;
+    private int rowCount = 200;
     private boolean containsSubQueries;
     private List<SubQueryExpression> subQueries;
-	public boolean isQuery = false;
+    public boolean isQuery = false;
     private boolean selectAll = false;
 
-	public boolean isAgg = false;
+    public boolean isAgg = false;
```
galkk: consider renaming to isAggregate

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,128 @@
...
+    public boolean isMetaField() {
+        return fieldName.startsWith("_");
+    }
+
+    /**
+     * Path of property field, for example "employee" in "employee.manager"
+     * @return path of property field
+     */
+    public String path() {
+        int lastDot = fieldName.lastIndexOf(".");
```
galkk: will throw an exception when there's no dots in the fieldName.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,128 @@
...
+    public String type() {
+        FieldMappingMetaData metaData = typeMappings.get(fieldName);
+        Map<String, Object> source = metaData.sourceAsMap();
+        String[] fieldPath = fieldName.split("\\.");
+
+        /*
+         * When field is not nested the metaData source is fieldName -> type
+         * When it is nested or contains "." in general (ex. fieldName.nestedName) the source is nestedName -> type
+         */
+        Map<String, Object> fieldMapping = (Map<String, Object>) source.get(fieldPath.length == 1 ? fieldName : fieldPath[1]);
```
galkk: that's a bit hard to read. Consider moving that ternary operator result into local variable

src/test/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMappingTest.java
```diff
@@ -0,0 +1,101 @@
...
+                                           FieldMapping::isPropertyField);
+    }
+
+    private Map<String, Field> fieldsSpecifiedInQuery(String...fieldNames) {
+        return Arrays.stream(fieldNames).
+                      collect(Collectors.toMap(name -> name,
+                                               name -> new Field(name, "")));
+    }
+
+    private Map<String, FieldMappingMetaData> typeMappings() {
```
galkk: Please inline this method, it just makes test less understandable

src/test/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMappingTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+        assertThat(
+            new FieldMapping("employee.first.keyword"),
+            isPropertyField(true)
+        );
+    }
+
+    @Test
+    public void testFieldIsNotProperty() {
+        assertThat(
```
galkk: Please extract those sequential assertThat into different tests

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -334,41 +348,67 @@ private String getFieldName(Field field) {
...
+        return isSelectAll() && fetchFields(query).isEmpty();
+    }
+
+    /**
+     * Special case which trades off consistency of SELECT * meaning for more intuition from customer perspective.
+     * In other cases, * means all regular fields on the level.
+     * The only exception here is * picks all non-regular (nested) fields as JSON without flatten.
+     */
+    private void populateAllNestedFields(List<Schema.Column> columns, List<String> fields) {
+        Set<String> nestedFieldNames = fields.stream().
```
galkk: That collection to set makes me suspicious. We're losing information here and doing it silently.
Should we fail fast? Add aliases? Or this will be handled elsewhere?

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
dai-chen: I found this is supposed to be a constant of Set. Fixed it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
dai-chen: done!

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Select.java
dai-chen: done!

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: Any suggestion for the variable name?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMappingTest.java
dai-chen: Good point! Fixed it.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMappingTest.java
dai-chen: done! Was planning to use this for more test cases against `type()` and then gave up due to time constraint...

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: I did this on purpose based on the assumption that is field is property. As you're pointing out, I will add the verification. Also now I'm thinking since this method is part of the public API set, it is better to be more defensive.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: Will use `root`. Comment if any better idea.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
arsen-es: Could you clarify what information is being lost? I don't understand what the issue is.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,132 @@
...
+    /** Name of the Field to be parsed */
+    private final String fieldName;
+
+    /** Native mapping information returned from ES */
+    private final Map<String, FieldMappingMetaData> typeMappings;
+
+    /** Maps a field name to Field object that specified in query explicitly */
+    private final Map<String, Field> specifiedFieldsByName;
+
+
```
galkk: nitpick: unnecessary empty line

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Path of property field, for example "employee" in "employee.manager"
+     * @return path of property field
+     */
+    public String path() {
+        int lastDot = fieldName.lastIndexOf(".");
+        if (lastDot == -1) {
+            throw new IllegalStateException("path() is being invoked on wrong field");
```
galkk: add field name to the exception message

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -289,26 +303,26 @@ private String getFieldName(Field field) {
...
          * respective Field object to check for aliases.
          */
         Map<String, Field> fieldMap = fetchFieldMap(query);
         List<Schema.Column> columns = new ArrayList<>();
-        for (String field : fields) {
+        for (String fieldName : fieldNameList) {
             // _score is a special case since it is not included in typeMappings, so it is checked for here
-            if (field.equals("_score")) {
-                columns.add(new Schema.Column(field, fetchAlias(field, fieldMap), Schema.Type.FLOAT));
+            if (fieldName.equals("_score")) {
```
galkk: that _score string repeats twice, consider extracting to named constant

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
galkk: nested field names is set, and we're flattening field mappings here, so in theory if there will be 2 fields with the same name, we will keep only one. If there are no chances that there will be duplicates, I'd expect to see List instead of Set

galkk:
All my comments are minor and I don't see any significant issues with the PR. Good job

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: > nested field names is set, and we're flattening field mappings here, so in theory if there will be 2 fields with the same name, we will keep only one. If there are no chances that there will be duplicates, I'd expect to see List instead of Set

I guess that's what this method meant to do. Will debug to confirm.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: my bad, fixed.

src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMapping.java
dai-chen: I'm used to adding one more to separate member fields and constructor clearly :) removing.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: I was thinking cover it in new `FieldMapping` class. I feel it's not that common, so finally I only add a `isMetaField`. Will add constant for it.

dai-chen: I found that there are quite a few more in the whole codebase. For the time being, I'm just adding a constant for the 2 occurrences you mentioned in `SelectResultSet`. (`Field` is domain object for SQL domain area. I think we should move this constant to some new class in our `esdomain` package).

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
arsen-es: > 
> 
> nested field names is set, and we're flattening field mappings here, so in theory if there will be 2 fields with the same name, we will keep only one. If there are no chances that there will be duplicates, I'd expect to see List instead of Set

I see, thanks for explanation. Good point.

dai-chen: I just debugged to confirm. As I mentioned before, the culprit is still the ES API we're using. For SELECT *, It returns field list like "comment.likes", "comment.data", "message.author" etc. while we want to figure out all the nested field name like "comment", "message". That's why we're collecting their `path()` into Set. Does this make sense?

dai-chen:
Address all comments, merged latest change in master. All integration tests passed and did some sanity test manually in Kibana and DbVisualizer:

image
image
image
image

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
arsen-es: I think in that case the name of the variable should not be `nestedFieldNames`, but rather `nestedFieldPaths` or `complexFields` or something similar.

dai-chen: Sure, renaming to `nestedFieldPaths` for clarity.

Issue #103, opened 2019-06-28
Add ability to sort on date_format result


galkk:
The following query should sort accrodingly

```
POST _opendistro/_sql
{
  "query": "SELECT agent, ip, date_format(utc_time, 'dd-MM-YYYY') date FROM kibana_sample_data_logs ORDER BY date_format(utc_time, 'dd-MM-YYYY') desc, ip"
}
```

PR #105, opened 2019-06-28
Fix generated script for ORDER BY for ORDER BY date_format case

galkk:
*Issue #, if available: [103](https://github.com/opendistro-for-elasticsearch/sql/issues/103)

*Description of changes:Performing analysis of sorted field, and if it's "script" field then generate "script" order by

Testing: unit test, integration test, manual testing via Kibana on local elasticsearch instance

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -24,11 +24,17 @@ public class Order {
 	private String nestedPath;
 	private String name;
 	private String type;
+	private boolean isScriptField;
 
-	public Order(String nestedPath, String name, String type) {
+	public boolean isScriptField() {
+		return isScriptField;
+	}
+
+	public Order(String nestedPath, String name, String type, boolean isScriptField) {
         this.nestedPath = nestedPath;
 		this.name = name;
 		this.type = type;
+		this.isScriptField = isScriptField;
 	}
 
     public String getNestedPath() {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -26,15 +26,15 @@ public class Order {
 	private String type;
 	private boolean isScriptField;
 
-	public boolean isScriptField() {
-		return isScriptField;
-	}
+    public boolean isScriptField() {
+        return isScriptField;
+    }
 
 	public Order(String nestedPath, String name, String type, boolean isScriptField) {
         this.nestedPath = nestedPath;
 		this.name = name;
 		this.type = type;
-		this.isScriptField = isScriptField;
+        this.isScriptField = isScriptField;
 	}
 
     public String getNestedPath() {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -24,17 +24,17 @@ public class Order {
 	private String nestedPath;
 	private String name;
 	private String type;
-    private boolean isScriptField;
+    private boolean isScript;
 
-    public boolean isScriptField() {
-        return isScriptField;
+    public boolean isScript() {
+        return isScript;
     }
 
-	public Order(String nestedPath, String name, String type, boolean isScriptField) {
+	public Order(String nestedPath, String name, String type, boolean isScript) {
         this.nestedPath = nestedPath;
 		this.name = name;
 		this.type = type;
-        this.isScriptField = isScriptField;
+        this.isScript = isScript;
 	}
 
     public String getNestedPath() {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -24,7 +24,7 @@ public class Order {
 	private String nestedPath;
 	private String name;
 	private String type;
-	private boolean isScriptField;
+    private boolean isScriptField;
 
     public boolean isScriptField() {
         return isScriptField;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
@@ -215,22 +215,28 @@ public class SqlParser {
             String type = sqlSelectOrderByItem.getType().toString();
 
             SQLExpr expr = sqlSelectOrderByItem.getExpr();
-            Field f = FieldMaker.makeField(expr, null, null);
+            Field field = FieldMaker.makeField(expr, null, null);
 
             String orderByName;
-            if (f instanceof MethodField && f.getName().equals("script")) {
-                MethodField mf = (MethodField) f;
+            if (field instanceof MethodField && field.getName().equals("script")) {
+                // TODO The consumer of ordering doesn't know the type of the expression, and elasticsearch needs to
+                // specify it for the script field ordering. Explore opportinities to supply that information to
+                // client code. One of the options is to have metadata store about functions that will supply
+                // return type of the result function here.
 
-                // TODO(galk) Ensure that the index will be always 1
-                orderByName = mf.getParams().get(1).toString();
+                MethodField methodField = (MethodField) field;
+
+                // 0 - generated field name
+                final int SCRIPT_CONTENT_INDEX = 1;
+                orderByName = methodField.getParams().get(SCRIPT_CONTENT_INDEX).toString();
 
             } else {
-                orderByName = f.toString();
+                orderByName = field.toString();
             }
 
             orderByName = orderByName.replace("`", "");
             if (alias != null) orderByName = orderByName.replaceFirst(alias + "\\.", "");
-            select.addOrderBy(f.getNestedPath(), orderByName, type, f instanceof MethodField);
+            select.addOrderBy(field.getNestedPath(), orderByName, type, field instanceof MethodField);
 
         }
     }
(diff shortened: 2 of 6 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/DateFormatTest.java
```diff
@@ -61,6 +63,52 @@ public void equalCondition() {
...
+        assertThat(formula.getName(), containsString("DateTimeFormatter.ofPattern"));
+
+        Order ip = orderBys.get(1);
+
+        assertThat(ip.isScriptField(), is(false));
+        assertThat(ip.getName(), is("ip"));
+        assertThat(ip.getType(), is("ASC"));
+    }
+
+    public Select getSelect(String query) {
```
dai-chen: nitpick: `private`?

galkk: done

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
```diff
@@ -208,18 +209,34 @@ private void findOrderBy(MySqlSelectQueryBlock query, Select select) throws SqlP
...
+                orderByName = methodField.getParams().get(SCRIPT_CONTENT_INDEX).toString();
+
+            } else {
+                orderByName = field.toString();
+            }
+
             orderByName = orderByName.replace("`", "");
             if (alias != null) orderByName = orderByName.replaceFirst(alias + "\\.", "");
-            select.addOrderBy(f.getNestedPath(), orderByName, type);
+            select.addOrderBy(field.getNestedPath(), orderByName, type, field instanceof MethodField);
```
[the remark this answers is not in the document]
galkk: That will require another overload in addOrderBy, that will set default value to false in Order.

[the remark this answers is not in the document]
dai-chen: > That will require another overload in addOrderBy, that will set default value to false in Order.

Sorry, I didn't make this clear. I meant you pass true for all `MethodField` no matter if its name is "script" or not (different from your `if` check above). I think the variable naming `isScriptField` is inconsistent with what you pass in here.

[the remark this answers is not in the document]
galkk: Ok, I see. I'll rename

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
```diff
@@ -24,11 +24,17 @@
 	private String nestedPath;
 	private String name;
 	private String type;
+	private boolean isScriptField;
 
-	public Order(String nestedPath, String name, String type) {
+	public boolean isScriptField() {
+		return isScriptField;
```
arsen-es: please do not use tabs, only spaces. 4spaces for a single level indentation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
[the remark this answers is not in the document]
dai-chen: Thanks! I think you can either rename `isScriptField` or add check for `MethodField` name here. It depends on if your changes in `if (order.isScriptField())` in `setSorts` applies to all `MethodField` or only `MethodField` with name `script`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DateFormatIT.java
```diff
@@ -118,6 +119,22 @@ public void or() throws SqlParseException {
...
 
+
+    @Test
+    public void sortByDateFormat() throws IOException {
+        // Sort by expression in descending order, but sort inside in ascending order, so we increase our confidence
+        // that successful test isn't just random chance.
+
+        JSONArray hits =
+                getHits(executeQuery("SELECT all_client, insert_time " +
+                        "FROM " + TestsConstants.TEST_INDEX_ONLINE + "/online " +
```
arsen-es: I suggest we do not specify the type of the documents, since it's already deprecated, and we'll have to remove that usage soon in the future anyway. So do not add the `"/online"` part.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
```diff
@@ -208,9 +209,23 @@ private void setWhere(Where where) throws SqlParseException {
...
             } else {
-                request.addSort(order.getName(), SortOrder.valueOf(order.getType()));
+                if (order.isScriptField()) {
+                    // TODO: Investigate how to find the type of expression (string or number)
+                    // As of now this shouldn't be a problem, because the support is for date_format function
+
+                    ScriptSortBuilder.ScriptSortType string = ScriptSortBuilder.ScriptSortType.STRING;
+                    request.addSort(
+                            SortBuilders
+                                    .scriptSort(new Script(order.getName()), string)
```
arsen-es: Usage of `string` as an identifier is confusing, plus, I think it would be more readable to just use the enum value inline, instead of assigning it to a local var.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
```diff
@@ -24,11 +24,17 @@
...
+	}
+
+	public Order(String nestedPath, String name, String type, boolean isScriptField) {
```
arsen-es: Several observations here:
1. for formal part, `Order` is not a field, so this flag does not belong in this class, or it should be called something else.
1. you use this information only in `DefaultQueryAction` to actually add the sorters. But there are more than 2 types of sorters, e.g. see `GeoDistanceSortBuilder`, so if in the future we need to add more, following this approach we will need to add another flag to this class, so at least this flag should be replaced by some enum type.
1. considering the above two, it seems to me this extra information is just a property of the field, and since you need it anyway, I think it belongs more on the `Field` object, and then ideally `Order` has a reference to that `Field` object, not just the name of it. I think what I suggest is a bigger change, but to me it seems the better alternative in terms of maintainability.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
```diff
@@ -208,18 +209,34 @@ private void findOrderBy(MySqlSelectQueryBlock query, Select select) throws SqlP
...
             if (sqlSelectOrderByItem.getType() == null) {
                 sqlSelectOrderByItem.setType(SQLOrderingSpecification.ASC);
             }
             String type = sqlSelectOrderByItem.getType().toString();
 
+            SQLExpr expr = sqlSelectOrderByItem.getExpr();
+            Field field = FieldMaker.makeField(expr, null, null);
+
+            String orderByName;
+            if (field instanceof MethodField && field.getName().equals("script")) {
```
arsen-es: np: And to complement my suggestion with another point, this condition could be just replaced with `field.isScriptField()` .

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
dai-chen: > Several observations here:
> 
> 1. for formal part, `Order` is not a field, so this flag does not belong in this class, or it should be called something else.
> 2. you use this information only in `DefaultQueryAction` to actually add the sorters. But there are more than 2 types of sorters, e.g. see `GeoDistanceSortBuilder`, so if in the future we need to add more, following this approach we will need to add another flag to this class, so at least this flag should be replaced by some enum type.
> 3. considering the above two, it seems to me this extra information is just a property of the field, and since you need it anyway, I think it belongs more on the `Field` object, and then ideally `Order` has a reference to that `Field` object, not just the name of it. I think what I suggest is a bigger change, but to me it seems the better alternative in terms of maintainability.

For #2, sounds like we better replace this "tagged" class by subtyping.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DateFormatIT.java
galkk: Followed existing approach, didn't know about deprecated /online, thanks.
Removed

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
galkk: I moved that to separate method.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
galkk: While I agree with that in general, I feel that this entire part should be rearchitected, so as of now to invest here might be a wasted effort.

Example: the problem that currently marked as ignored on integration test. When such query is converted into elastic's DSL, the entire Order by is dropped, because it should have order by in aggregation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
```diff
@@ -24,11 +24,17 @@
 	private String nestedPath;
 	private String name;
 	private String type;
+    private boolean isScript;
```
arsen-es: I suggest `isForScriptField`

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
arsen-es: I also agree with you, but I think the more special cases we add now the harder it will be to refactor later. I quickly checked the code, the `Order` constructor and the `addOrderBy` methods are used only in one place, so you're already modifying all the code paths that will be affected if you move the `isScript` logic to the `Field` class. Anyway, since you're working on a single bug fix, this is just a suggestion, not a blocker.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
galkk: It's not technically correct.
The corresponding DSL for scripted ordering is script, and it's not related to field

```
      "_script": {
        "script": {
          "source": "def date_format_2043496359 = DateTimeFormatter.ofPattern('dd-MM-YYYY').withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(doc['insert_time'].value.getMillis()));return date_format_2043496359;",
          "lang": "painless"
        },
        "type": "string",
        "order": "desc"
      }
```

PR #106, opened 2019-07-03
Locale root

arsen-es:
*Issue #, if available:* N/A

*Description of changes:* Just creating utility methods not to make everyone use `Locale.ROOT` all the time. Plus fixed some warnings. Replaced all the usages of `String.format` in test classes, someone (maybe me) will need to do for non-test as well.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
@@ -1,0 +1,41 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.utils;
+
+import java.util.Locale;
+
+public class StringUtils {
+
+    public static String format(final String format, Object... args) {
+
+        return String.format(Locale.ROOT, format, args);
+    }
+
+    public static String toLower(final String input) {
+
+        return input.toLowerCase(Locale.ROOT);
+    }
+
+    public static String toUpper(final String input) {
+
+        return input.toUpperCase(Locale.ROOT);
+    }
+
+    private StringUtils() {
+
+        throw new AssertionError(getClass().getCanonicalName() + " is a utility class and must not be initialized");
+    }
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
@@ -17,25 +17,70 @@ package com.amazon.opendistroforelasticsearch.sql.utils;
 
 import java.util.Locale;
 
+/**
+ * Helper class for wrapping locale-dependent methods
+ * of {@link String} class in equivalent locale-independent
+ * methods which always use {@link Locale#ROOT|.
+ */
 public class StringUtils {
 
+    /**
+     * Returns a formatted string using the specified format string and
+     * arguments, as well as the {@link Locale#ROOT} locale.
+     *
+     * @param  format
+     *         format string
+     *
+     * @param  args
+     *         arguments referenced by the format specifiers in the format string
+     *
+     * @throws  java.util.IllegalFormatException
+     *          If a format string contains an illegal syntax, a format
+     *          specifier that is incompatible with the given arguments,
+     *          insufficient arguments given the format string, or other
+     *          illegal conditions.
+     *
+     * @return  A formatted string
+     *
+     * @see java.lang.String#format(Locale, String, Object...)
+     */
     public static String format(final String format, Object... args) {
-
         return String.format(Locale.ROOT, format, args);
     }
 
+    /**
+     * Converts all of the characters in this {@code String} to lower
+     * case using the rules of the {@link Locale#ROOT} locale. This is equivalent to calling
+     * {@link String#toLowerCase(Locale)} with {@link Locale#ROOT}.
+     *
+     * @param input
+     *        the input String
+     *
+     * @return  the {@code String}, converted to lowercase
+     *
+     * @see     java.lang.String#toLowerCase(Locale)
+     */
     public static String toLower(final String input) {
-
         return input.toLowerCase(Locale.ROOT);
     }
 
+    /**
+     * Converts all of the characters in this {@code String} to upper
+     * case using the rules of the {@link Locale#ROOT} locale. This is equivalent to calling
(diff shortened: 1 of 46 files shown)


arsen-es:
Sorry for the big change, I just had it done a while ago and now want to clean-up my local branches, so just decided to push.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
```diff
@@ -0,0 +1,41 @@
...
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.utils;
+
+import java.util.Locale;
+
+public class StringUtils {
```
galkk: Javadoc

galkk:
I think that for most of the cases (especially tests) we shouldn't use format at all.

For example:
```
        String result = explain(format("{\"query\":\"" +
                "SELECT * " +	                "SELECT * " +
                "FROM %s " +	                "FROM %s " +
                "WHERE firstname LIKE 'A%%' AND age > 20 " +	                "WHERE firstname LIKE 'A%%' AND age > 20 " +
                "GROUP BY gender " +	                "GROUP BY gender " +
                "ORDER BY _score\"}", TestsConstants.TEST_INDEX_ACCOUNT));	
```

We don't need format here at all, we can just add index name into the query.

Proposal: 

Split this PR into 2: StringUtils + the places where format is strictly necessary (non-test codebase, exception strings), and the rest (unit tests). The former PR will go easily, because it's needed, for the latter we can continue our conversation if format is actually needed there at all

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static String format(final String format, Object... args) {
+
```
galkk: Those empty lines on start of the method are catching the eye. If we will take same methid JDK as an example, they don't follow that apporach, as well as most of the codebase

https://hg.openjdk.java.net/jdk/jdk12/file/06222165c35f/src/java.base/share/classes/java/lang/String.java#l2987
```
    public static String format(String format, Object... args) {
        return new Formatter().format(format, args).toString();
    }

```

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
arsen-es: +1, updating

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
arsen-es: Will remove to comply with the style

arsen-es:
> 
> 
> I think that for most of the cases (especially tests) we shouldn't use format at all.
> 
> For example:
> 
> ```
>         String result = explain(format("{\"query\":\"" +
>                 "SELECT * " +	                "SELECT * " +
>                 "FROM %s " +	                "FROM %s " +
>                 "WHERE firstname LIKE 'A%%' AND age > 20 " +	                "WHERE firstname LIKE 'A%%' AND age > 20 " +
>                 "GROUP BY gender " +	                "GROUP BY gender " +
>                 "ORDER BY _score\"}", TestsConstants.TEST_INDEX_ACCOUNT));	
> ```
> 
> We don't need format here at all, we can just add index name into the query.
> 
> Proposal:
> 
> Split this PR into 2: StringUtils + the places where format is strictly necessary (non-test codebase, exception strings), and the rest (unit tests). The former PR will go easily, because it's needed, for the latter we can continue our conversation if format is actually needed there at all

1. I fixed the line length for everything going beyond 120 characters. I fixed some unused import warnings, and some others as well. So if I revert my change, I will lose all those fixes.
2. The issue you describe comes because most of our tests are a result of migration, and not written from scratch by us, and as long as they tested the functionality they were good enough. Bringing their code to a very good shape is indeed a useful thing, but I think in terms of priority that will be the last thing to address.
3. Removing the format calls is independent of this change, if we decide to do it, it has to be done anyway, so it's not that my work is wasted, at least not for the time until we remove it. Another point is when we add a new test we usually copy the skeleton code from an existing one, and having the `String.format` call in one place might propagate to the new tests.

So I still suggest I address all the other comments and push the full change. Whenever and whoever decides to address your concern about using `format` where it's not necessary or even discouraged, they can do that change. Addressing that change as 2nd part of this PR adds a lot of extra work, which I don't have time for right now, since there are higher priority items to work on.

@dai-chen, @abbashus , @penghuo - what do you think? If you are in agreement with Andy's proposal, I will just keep the StringUtils addition along with unit tests, and revert all the other changes.

dai-chen:
> > I think that for most of the cases (especially tests) we shouldn't use format at all.
> > For example:
> > ```
> >         String result = explain(format("{\"query\":\"" +
> >                 "SELECT * " +	                "SELECT * " +
> >                 "FROM %s " +	                "FROM %s " +
> >                 "WHERE firstname LIKE 'A%%' AND age > 20 " +	                "WHERE firstname LIKE 'A%%' AND age > 20 " +
> >                 "GROUP BY gender " +	                "GROUP BY gender " +
> >                 "ORDER BY _score\"}", TestsConstants.TEST_INDEX_ACCOUNT));	
> > ```
> > 
> > 
> > We don't need format here at all, we can just add index name into the query.
> > Proposal:
> > Split this PR into 2: StringUtils + the places where format is strictly necessary (non-test codebase, exception strings), and the rest (unit tests). The former PR will go easily, because it's needed, for the latter we can continue our conversation if format is actually needed there at all
> 
> 1. I fixed the line length for everything going beyond 120 characters. I fixed some unused import warnings, and some others as well. So if I revert my change, I will lose all those fixes.
> 2. The issue you describe comes because most of our tests are a result of migration, and not written from scratch by us, and as long as they tested the functionality they were good enough. Bringing their code to a very good shape is indeed a useful thing, but I think in terms of priority that will be the last thing to address.
> 3. Removing the format calls is independent of this change, if we decide to do it, it has to be done anyway, so it's not that my work is wasted, at least not for the time until we remove it. Another point is when we add a new test we usually copy the skeleton code from an existing one, and having the `String.format` call in one place might propagate to the new tests.
> 
> So I still suggest I address all the other comments and push the full change. Whenever and whoever decides to address your concern about using `format` where it's not necessary or even discouraged, they can do that change. Addressing that change as 2nd part of this PR adds a lot of extra work, which I don't have time for right now, since there are higher priority items to work on.
> 
> @dai-chen, @abbashus , @penghuo - what do you think? If you are in agreement with Andy's proposal, I will just keep the StringUtils addition along with unit tests, and revert all the other changes.

I'm fine with either way (check in what you have now or revert) because we haven't figured out how to do it much better. Concat many pieces of string or hardcode entire query, both look not perfect to me. I think we can continue the discussion on another Issue/PR.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
arsen-es: I think I removed all the ones I introduced, maybe even the ones that I didn't.

arsen-es:
Checked in the `StringUtils` class in a separate PR (#107), so this one now is pending the decision whether to update all usages of `String.format` with the new `format` method in the tests. Please see former conversation for more details.

arsen-es:
Closing this PR, since the essential part was already checked in as part of #107, and this code is outdated.

Issue #108, opened 2019-07-10
date_format doesn't work for JDBC format output


galkk:
Currently the query 
``` 
SELECT date_format(insert_time, 'dd-MM-YYYY') FROM elasticsearch-sql_test_index_online limit 1
```
fails with an error.

dai-chen:
Maybe related: https://github.com/opendistro-for-elasticsearch/sql-jdbc/issues/14.

galkk:
Fixed in PR

PR #109, opened 2019-07-10
Add support for function calls in select for JDBC calls.

galkk:
*Issue #*:[108](https://github.com/opendistro-for-elasticsearch/sql/issues/108)

*Description of changes: *

The script fields, that are used to implement date_format, have name "script", and the original function name is lost. Because of lost original function name, the result converter can't add type information. Also the SelectResultSet didn't copied any fields into output, so fixed that part as well.

This fix will fix the issue for date_format function, and will give a way for future built-in script functions. 

Testing: unit tests, integration tests, manual test via kibana.

Also now the prefixes for auto-generated fields will start with function name, instead of not-helping "field".

```
POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT date_format(utc_time, 'dd-MM-YYYY'), bytes/10 from kibana_sample_data_logs ORDER BY bytes/10 desc limit 2 "
}
```
returns
```json
{
  "schema": [
    {
      "name": "date_format_1541758184",
      "type": "text"
    },
    {
      "name": "divide_1874722970",
      "type": "double"
    }
  ],
  "total": 10000,
  "datarows": [
    [
      "18-09-2018",
      1998
    ],
    [
      "13-08-2018",
      1995
    ]
  ],
  "size": 2,
  "status": 200
}
```

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -27,7 +27,7 @@ public class Order {
     private Field sortField;
 
     public boolean isScript() {
-        return sortField != null && sortField instanceof ScriptMethodField;
+        return sortField != null && sortField.isScriptField();
     }
 
     public Order(String nestedPath, String name, String type, Field sortField) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -21,20 +21,20 @@ package com.amazon.opendistroforelasticsearch.sql.domain;
  *
  */
 public class Order {
-	private String nestedPath;
-	private String name;
-	private String type;
-    private boolean isScript;
+    private String nestedPath;
+    private String name;
+    private String type;
+    private Field sortField;
 
     public boolean isScript() {
-        return isScript;
+        return sortField != null && sortField instanceof ScriptMethodField;
     }
 
-	public Order(String nestedPath, String name, String type, boolean isScript) {
+    public Order(String nestedPath, String name, String type, Field sortField) {
         this.nestedPath = nestedPath;
-		this.name = name;
-		this.type = type;
-        this.isScript = isScript;
+        this.name = name;
+        this.type = type;
+		this.sortField = sortField;
 	}
 
     public String getNestedPath() {
@@ -46,19 +46,22 @@ public class Order {
     }
 
     public String getName() {
-		return name;
-	}
+        return name;
+    }
 
-	public void setName(String name) {
-		this.name = name;
-	}
+    public void setName(String name) {
+        this.name = name;
+    }
 
-	public String getType() {
-		return type;
-	}
+    public String getType() {
+        return type;
+    }
 
-	public void setType(String type) {
-		this.type = type;
-	}
+    public void setType(String type) {
+        this.type = type;
+    }
 
+    public Field getSortField() {
+        return sortField;
+    }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
@@ -34,8 +34,8 @@ public class Order {
         this.nestedPath = nestedPath;
         this.name = name;
         this.type = type;
-		this.sortField = sortField;
-	}
+        this.sortField = sortField;
+    }
 
     public String getNestedPath() {
         return nestedPath;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/ScriptMethodField.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/ScriptMethodField.java
@@ -29,4 +29,9 @@ public class ScriptMethodField extends MethodField {
     public String getFunctionName() {
         return functionName;
     }
+
+    @Override
+    public boolean isScriptField() {
+        return true;
+    }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/ScriptMethodField.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/ScriptMethodField.java
@@ -1,0 +1,32 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.domain;
+
+import java.util.List;
+
+/** Stores information about function name for script fields */
+public class ScriptMethodField extends MethodField {
+    private final String functionName;
+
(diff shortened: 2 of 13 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
```diff
@@ -103,6 +103,25 @@ protected Request getSqlRequest(String request, boolean explain) {
         return sqlRequest;
     }
 
+    protected String executeQuery(String query, String requestType) {
```
dai-chen: We could make this method more generic by accepting a URL param list rather than `requestType`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/ScriptMethodField.java
```diff
@@ -0,0 +1,32 @@
...
+/** Stores information about function name for script fields */
+public class ScriptMethodField extends MethodField {
+    private final String functionName;
+
+    public ScriptMethodField(String functionName, List<KVValue> params, String option, String alias) {
+        super("script", params, option, alias);
+        this.functionName = functionName;
+    }
+
+    public String getFunctionName() {
```
dai-chen: Is this method common to all `MethodField`? Should we pull this method up to parent class?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -273,6 +275,22 @@ private String getFieldName(Field field) {
...
             case "max":
                 return Schema.Type.DOUBLE;
+            case "script": {
+                ScriptMethodField smf = (ScriptMethodField) field;
+                // TODO: this information is disconnected from the function definitions in SQLFunctions.
+                // Refactor SQLFunctions to have functions self-explanatory (types, scripts) and pluggable
+                // (similar to Strategy pattern).
+
+                switch (smf.getFunctionName()) {
+                    case "date_format": {
```
dai-chen: I remember another immediate requirement is to support arithmetic expression, ex. "SELECT age / 2 FROM A".

For "date_format", you've aware of it, why still return TEXT? I thought we either maintain a mapping from known function to its return value type (like what you do here), or we return TEXT for everything for now.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -312,6 +312,12 @@ public static MethodField makeMethodField(String name, List<SQLExpr> arguments,
             paramers.addAll(tempParamers);
         }
 
-        return new MethodField(finalMethodName, paramers, option == null ? null : option.name(), alias);
+        if (!builtInScriptFunction) {
```
dai-chen: A little more readable by checking `builtInScriptFunction` instead of `!builtInScriptFunction` and switch logic in `if` and `else`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
```diff
@@ -103,6 +103,25 @@ protected Request getSqlRequest(String request, boolean explain) {
...
+            String endpoint = "/_opendistro/_sql?format=" + requestType;
+            String requestBody = makeRequest(query);
+
+            Request sqlRequest = new Request("POST", endpoint);
+            sqlRequest.setJsonEntity(requestBody);
+
+            RestClient restClient = ESIntegTestCase.getRestClient();
+            Response response = restClient.performRequest(sqlRequest);
+            Assert.assertEquals(200, response.getStatusLine().getStatusCode());
+            String responseString = TestUtils.getResponseBody(response, true);
```
dai-chen: Is `true` required? If not, all these 4 lines seem same as `executeRequest(request)`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
galkk: Changed in next version, will publish after all tests will pass

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/ScriptMethodField.java
galkk: No, there are method fields that are not script fields

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
galkk: agree, fixed

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
galkk: It's easier to debug it this way (you don't need to do additional formatting) :)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
```diff
@@ -218,7 +219,7 @@ private void addOrderByToSelect(Select select, List<SQLSelectOrderByItem> items,
             Field field = FieldMaker.makeField(expr, null, null);
 
             String orderByName;
-            if (isScriptField(field)) {
+            if (field instanceof ScriptMethodField) {
```
dai-chen: Since we've had a class/abstraction for script method field, I think we should move this to a new method in Field class hierachy like `getOrderbyName` to take advantage of polymorphism and avoid `instanceof` as much as possible (as you said, we've had too many).

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -312,6 +312,11 @@ public static MethodField makeMethodField(String name, List<SQLExpr> arguments,
             paramers.addAll(tempParamers);
         }
 
-        return new MethodField(finalMethodName, paramers, option == null ? null : option.name(), alias);
+        if (builtInScriptFunction) {
+            return new ScriptMethodField(name, paramers, option == null ? null : option.name(), alias);
+        } else {
+          return new MethodField(
```
dai-chen: nit: need a little reformat here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Order.java
```diff
@@ -21,20 +21,20 @@
...
 
-	public Order(String nestedPath, String name, String type, boolean isScript) {
+    public Order(String nestedPath, String name, String type, Field sortField) {
         this.nestedPath = nestedPath;
-		this.name = name;
-		this.type = type;
-        this.isScript = isScript;
+        this.name = name;
+        this.type = type;
+		this.sortField = sortField;
```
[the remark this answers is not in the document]
galkk: Slipped through, sorry

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
galkk: I feel that this is an overkill as of now. When time will come, it can be refactored.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
galkk: done

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SqlParser.java
galkk: Done

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
dai-chen: It's okay. Thanks!

Issue #40, opened 2019-04-25
CSV export of JSON values


kakulukia:
GET _opendistro/_sql?format=csv

is not exporting quoted values and thus having JSON with additional commas in one of the exported fields will kill the following fields order. 
Can quoted output be forced somehow?

dai-chen:
> GET _opendistro/_sql?format=csv
> 
> is not exporting quoted values and thus having JSON with additional commas in one of the exported fields will kill the following fields order.
> Can quoted output be forced somehow?

Could you please provide a sample of your data and actual output (and maybe expected output as well)? Thanks!

kakulukia:
I cant share our projects data, but let me give one small example:

    {
      "one", "one",
      "two": [1, 2],
      "three": {"another": true, "dict": false}
    }

This will result in (spreading the data in 5 rows rather than the intended 3 - and also breaking it):

    one,two,three
    one,[1, 2], {another=true, dict=false}

But it should result in at least:

    one,two,three
    "one","[1, 2]","{another=true, dict=false}"

There might even be a way of keeping the JSON intact, but i dont mind that much.

arsen-es:
The quotation part is required by [CSV spec](https://tools.ietf.org/html/rfc4180) (see 2.6).

As for nested objects, I did not find a standard for conversion, but I checked some implementations and all are flattening the JSON by putting the fully qualified path to the field in the header, e.g. in the provided example the headers will be the following (the delimiter between the path terms is also a subject of preference, some use `/` instead of `_`):
`one,two,three_another,three_dict`.

Another option is flattening the arrays as well:
`one,two_0,two_1,three_another,three_dict`.

kakulukia:
Please dont alter the data, since whats exported now can still be parsed in a nice way. Please, just add the quotation, so that the exported csv is actually valid.

arsen-es:
Resolved by #42

kakulukia:
Sadly i just tested a newer version of Kibana 6.7.1 only to see that the CSV export got worse than before. Its still not featuring quoted values, but lists that are separated by space now. Having spaces in the values, this makes it very hard to fix the existing export data. :(

Actual JSON values are untouched as far as i can tell.
Here is an example of an exported list: "[[value-1 value-2]]"

How can i see which version of SQL is included in my Kibana release?

kakulukia:
Ignore the previous post, our Kibana release still uses the old SQL version and the changed list format results from a now broken data import by our reports team.
[...]

PR #112, opened 2019-07-12
Fixing the escaping issue in CSV formatted output

arsen-es:
*Issue #, if available:* #40 

*Description of changes:* This complements the change in PR #42, which did not fix the escaping of quotes in values. Now all double quotes are replaced with 2 double quotes.

*How tested:* 
```
PUT nested_tst
{
  "settings": {
    "number_of_shards": 1
  },
  "mappings": {
      "properties": {
        "message": {
          "type": "nested",
          "properties": {
            "info": {
              "type": "keyword",
              "index": "true"
            },
            "author": {
              "type": "keyword",
              "index": "true"
            },
            "dayOfWeek": {
              "type": "long"
            }
          }
        },
        "comment": {
          "type": "nested",
          "properties": {
            "data": {
              "type": "keyword",
              "index": "true"
            },
            "likes": {
              "type": "long"
            }
          }
        },
        "myNum": {
          "type": "long"
        },
        "someField": {
          "type": "keyword",
          "index": "true"
        }
      }
  }
}

POST /nested_tst/_doc
{
    "message" : [{"info":"zz","author":"z\"z","dayOfWeek":6}],
    "comment": {
      "data":["aa","bb"],
      "likes":10
    },
    "myNum":[3,4],
    "someField":"a"
}

POST /nested_tst/_doc
{
    "message" : [{"info":"rr","author":"this \"value\" contains quotes","dayOfWeek":3}],
    "comment": {
      "data":["asdf","sdfg"],
      "likes":56
    },
    "myNum":[1,2,4],
    "someField":"ert"
}

POST /_opendistro/_sql?format=csv
{
  "query": "SELECT * FROM nested_tst";
}
```

the SQL query result:
```
"[3, 4]","{data=[aa, bb], likes=10}",a,"[{dayOfWeek=6, author=z""z, info=zz}]"
"[1, 2, 4]","{data=[asdf, sdfg], likes=56}",ert,"[{dayOfWeek=3, author=this ""value"" contains quotes, info=rr}]"
```

Dropped the output into a `.csv` file and opened with Excel, everything works as expected.


src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/csv/CSVResultsExtractor.java
```diff
@@ -336,8 +336,16 @@ private String findFieldValue(String header, Map<String, Object> doc, boolean fl
 
     private String quoteValueIfRequired(final String input, final String separator) {
```
galkk: please remove empty lines at start of methods (here and everywhere in PR)

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/csv/CSVResultsExtractor.java
```diff
@@ -336,8 +336,16 @@ private String findFieldValue(String header, Map<String, Object> doc, boolean fl
...
 
-        final String quote = input.contains(separator) ? "\"" : "";
-        return quote + input + quote + separator;
+        final String quote = "\"";
+        final String result;
```
galkk: consider removing this variable and just returning values from functions in if, following the advice from

https://blog.codinghorror.com/flattening-arrow-code/
#4: Always opportunistically return as soon as possible from the function.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -136,6 +139,23 @@ public void arraysAreQuotedInFlatMode() throws IOException {
...
+    @Test
+    public void doubleQuotesAreEscapedWithDoubleQuotes() throws IOException {
+
+        final String query = "SELECT * FROM " + TEST_INDEX_NESTED_WITH_QUOTES;
+
+        final CSVResult csvResult = executeCsvRequest(query, false);
+        final List<String> rows = csvResult.getLines();
+        Assert.assertThat(rows.size(), equalTo(2));
+
+        final String expectedValue1 = "\"[{dayOfWeek=6, author=z\"\"z, info=zz}]\"";
```
galkk: consider inlining variables

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
```diff
@@ -136,6 +139,23 @@ public void arraysAreQuotedInFlatMode() throws IOException {
...
+        final String expectedValue2 = "\"[{dayOfWeek=3, author=this \"\"value\"\" contains quotes, info=rr}]\"";
+
+        for (String row : rows) {
```
galkk: instead of loop use
assertThat(rows, containsInAnyOrder(expectedValue1, expectedValue2));

(consider inlining)

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/csv/CSVResultsExtractor.java
arsen-es: ok

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/csv/CSVResultsExtractor.java
arsen-es: I like the other advice which says minimize number of return statements in the function. It becomes really helpful when debugging the code, in other words putting 1 breakpoint instead of N. But this case is pretty simple, doesn't matter for me, will update.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: that will always fail in this case, because I am doing substring match, not the whole string.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CsvFormatResponseIT.java
arsen-es: Again, maybe it's more lines than necessary, but I like explicitly showing what's expected in the tests, and this looks cleaner to me. Not updating, we can discuss more if you disagree.

PR #113, opened 2019-07-12
Set response of "_explain" request as json content type

galkk:
*Description of changes:*

As of now this is quality of life improvement, that will enable client tools that can render json, render response of "_explain" nicely.

Example

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
@@ -104,7 +104,7 @@ public class RestSqlAction extends BaseRestHandler {
 
             if (request.path().endsWith("/_explain")) {
                 final String jsonExplanation = queryAction.explain().explain();
-                return sendResponse(jsonExplanation, OK);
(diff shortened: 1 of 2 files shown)


abbashus:
Though, this must have been manually tested and verified. Can we add a integration test here?

galkk:
> Though, this must have been manually tested and verified. Can we add a integration test here?

While I feel it's an overkill, it's easy to implement. Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -104,7 +104,7 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
 
             if (request.path().endsWith("/_explain")) {
                 final String jsonExplanation = queryAction.explain().explain();
-                return sendResponse(jsonExplanation, OK);
+                return channel -> channel.sendResponse(new BytesRestResponse(OK, "application/json; charset=UTF-8", jsonExplanation));
```
arsen-es: Probably we need to add some tests to catch this kind of a change if it didn't break any of the existing ones.

Issue #111, opened 2019-07-12
JDBC format of aggregation query with date_format adds unnecessary column


galkk:
the query 
```
POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT date_format(utc_time, 'dd-MM-YYYY'), count(*)  from kibana_sample_data_logs group by date_format(utc_time, 'dd-MM-YYYY')"
}
```

Works but returns one more null column in result schema.

```
  "schema": [
    {
      "name": "date_format_1410573132",
      "type": "text"
    },
    {
      "name": "date_format_1025417136",
      "type": "text"
    },
    {
      "name": "COUNT(*)",
      "type": "long"
    }
  ],
  "total": 10,
  "datarows": [
    [
      "10-08-2018",
      null,
      244
    ],
    [
```

The JDBC result formatting is in the SQL plugin, so creating issue here

PR #116, opened 2019-07-16
Remove of null column in JDBC output when script function is present

galkk:
*Issue #, if available:* [111](https://github.com/opendistro-for-elasticsearch/sql/issues/111)

*Description of changes:*

Fix of JDBC output when there's script function in group by. It is done by additional analysis in schema preparation step to differentiate script and method fields.

After this fix the output
```
POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT date_format(utc_time, 'YYYY-MM'),  count(*) from kibana_sample_data_logs group by date_format(utc_time, 'YYYY-MM')"
}
```

returns, as expected
```json
{
  "schema": [
    {
      "name": "date_format_1721549462",
      "type": "text"
    },
    {
      "name": "COUNT(*)",
      "type": "long"
    }
  ],
  "total": 3,
  "datarows": [
    [
      "2018-08",
      7132
    ],
    [
      "2018-09",
      4552
    ],
    [
      "2018-07",
      2321
    ]
  ],
  "size": 3,
  "status": 200
}
```

Testing:
integration test, manual via Kibana UI

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistroforelasticsearch.sql.domain;
 
+import com.alibaba.druid.sql.ast.SQLExpr;
 import com.amazon.opendistroforelasticsearch.sql.parser.ChildrenType;
 import com.amazon.opendistroforelasticsearch.sql.parser.NestedType;
 
@@ -35,8 +36,9 @@ public class Field implements Cloneable{
 	private String alias;
     private NestedType nested;
     private ChildrenType children;
+    private SQLExpr expression;
 
-	public Field(String name, String alias) {
+    public Field(String name, String alias) {
 		this.name = name;
 		this.alias = alias;
         this.nested = null;
@@ -126,5 +128,13 @@ public class Field implements Cloneable{
     /** Returns true if the field is script field. */
     public boolean isScriptField() {
         return false;
+    }
+
+    public void setExpression(SQLExpr expression) {
+        this.expression = expression;
+    }
+
+    public SQLExpr getExpression() {
+        return expression;
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
```diff
@@ -35,8 +36,9 @@
 	private String alias;
     private NestedType nested;
     private ChildrenType children;
+    private SQLExpr expression;
```
dai-chen: I don't see the use of this new field member. Did I miss it anywhere?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -226,23 +226,23 @@ private boolean containsWildcard(Query query) {
...
+
         if (queryResult instanceof Aggregations) {
-            fields = select.getGroupBys().isEmpty() ? new ArrayList<>() : select.getGroupBys().get(0);
-            for (Field field : select.getFields()) {
-                if (field instanceof MethodField) {
-                    fields.add(field);
+            List<Field> groupByFields = select.getGroupBys().isEmpty() ? new ArrayList<>() : select.getGroupBys().get(0);
+
+            for (Field selectField : select.getFields()) {
+                if (selectField instanceof MethodField && !selectField.isScriptField()) {
```
dai-chen: I assume what we want exactly here is de-duplicate. The `!selectField.isScriptField` check works on the assumption that there is no aggregation function implemented by painless script right? If so, is this true?

galkk: It's even worse (because as per Elastic's DSL script aggregations can be for both Bucket and Metric aggregations), but it will require considerable redesign, because the process of transforming of parsed SQL into current domain model loses a lot of information. For example - for query like

select date_format(format), count(*)
from index
group by date_format(format)
order by date_format(format)

there's no easy way to understand that the date_format is actually the same, because at this stage we're operating only with transformed expressions.

Adding of SQLExpr into field information in this PR is related to future attempt to de-duplication that is necessary for order by.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
galkk: see answer below. It's part of upcoming PR that I've slipped here. I can move it into different PR or leave it as is, because the change will be part of next PR anyway

Issue #69, opened 2019-06-04
Issue with select


lolobosse:
Hey there,

I'm kind of new to ElasticSearch but I'm not new to SQL and I cannot figure out why the following queries are not working (I'm using AWS ES (with the SQL release of 28.05)).

My data:

```
POST _opendistro/_sql
{
  "query": "select * from jobs"
}
```

```
{
  "took" : 1,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 2,
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "jobs",
        "_type" : "_doc",
        "_id" : "hAAwI2sBRGj0mG3xcZdf",
        "_score" : 1.0,
        "_source" : {
          "category" : 18,
          "city" : "Sankt Peter-Ording",
          "clickout" : null,
          "company" : "Kobs Köster GmbH",
          "country" : "de",
          "description" : "blabla",
          "email" : "blabla",
          "import_id" : "829ef41c0ca9f3a8cfd9fa4b6735a532",
          "location" : "54.30000,8.63330",
          "origin" : "Kobs",
          "title" : "Tischler (m/w/d) in Festanstellung",
          "created_at" : "2019-06-04T17:51:45.759721",
          "finished_at" : null,
          "impressions" : 0,
          "updated_at" : "2019-06-04T17:51:45.759733"
        }
      },
      {
        "_index" : "jobs",
        "_type" : "_doc",
        "_id" : "gwAuI2sBRGj0mG3xkZfH",
        "_score" : 1.0,
        "_source" : {
          "category" : 18,
          "city" : "Itzehoe",
          "clickout" : null,
          "company" : "Kobs Köster GmbH",
          "country" : "de",
          "description" : "blabla",
          "email" : "blabla",
          "import_id" : "d656e2a73c954703ef983f8d8fd9549e",
          "location" : "53.92324,9.51766",
          "origin" : "Kobs",
          "title" : "Tischler (m/w/d) in Festanstellung",
          "created_at" : "2019-06-04T17:49:42.984842",
          "finished_at" : null,
          "impressions" : 0,
          "updated_at" : "2019-06-04T17:49:42.984851"
        }
      }
    ]
  }
}
```

But when I do that:

```
POST _opendistro/_sql
{
[...]

PR #66, opened 2019-05-31
Migrate JSONRequestIT, MathFunctionsIT, MethodQueryIT to esintgtest package

penghuo:
*Issue #, if available:*

*Description of changes:*
Migrate JSONRequestIT, MathFunctionsIT, MethodQueryIT to esintgtest package

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #126, opened 2019-07-20
Support IN predicate subquery

penghuo:
*Issue #, if available:*

*Description of changes:*
There are two issues with existing subquey solution.
1. *inefficiency execution logic*. In current solution, the subquery execution is divided into two stage, (1) inner query execution. (2) outer query execution with the result of (1). This solution has the following performance issues, (1) memory issue, if the subquery running on the large dataset, all the result need to be buffered in the memory before running the outer query which could lead to OOM issue. (2) it also need to transfer out the whole dataset of the subquery when running outer query, it will has network impact.
2. *running block operation in transport thread*. The SQL plugin explicit split the SQL execution into two stages, (1) explain stage which running in the transport thread, (2) execution stage which running in the worker thread. Because the subquery need to be executed in the explain stage, which means the subquery need to be executed in the transport thread which will impact the system throughput. when enabling launch elasticsearch with -ea, it will throw 
`java.lang.AssertionError: Expected current thread to not be a transport thread`

*Solution*
The proposed solution is rewrite Subquery as JOIN query, for example
`SELECT * FROM A WHERE a IN (SELECT b FROM B)`
will be rewritten to
`SELECT A.* FROM A JOIN B ON A.a = B.b`

*Scope*
This CR is the first step to support predicate subquery. Currently, only IN predicate subquery is supported.

**update**
revert the change in #83. after this pr, the subquery parser doesn't have blocking operation and could run in the transport thread. This is no need to run subquery parser in the seperate worker thread any more.

**update**
add the optimizer to rewrite subquery as join.

**update**
change the optimizer to RewriteRuleExecutor.


src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * IN Subquery Rewriter.
+ * For example,
+ *  SELECT * FROM A WHERE a IN (SELECT b FROM B) should be rewritten to
+ *  SELECT A.* FROM A JOIN B ON A.a = B.b.
+ */
+public class InSubqueryRewriter implements SubqueryRewriter {
+
+    private final MySqlSelectQueryBlock rootQuery;
+
```
galkk: consider removing spaces between fields of the same type, it makes code hard to read and most of the code is written in different style (e.g. SubQueryExpression.java, CaseWhenParser.java) etc

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+        rootQuery.accept0(new SubqueryAliasRewriter());
+        this.rootQuery = rootQuery;
+        this.inSubQueryExpr = inSubQueryExpr;
+        this.inSubQuery = (MySqlSelectQueryBlock) inSubQueryExpr.getSubQuery().getQuery();
+
+        joinLeftExpr = inSubQueryExpr.getExpr();
+        joinRightExpr = rightExpr(inSubQuery);
+    }
+
+    /**
```
galkk: personal preference, feel free to ignore, but 
/** {@inheritDoc} */

Will do the job in more compact fastion, and, in general, I feel that this is useless style of doc and can be safely removed

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryType.java
```diff
@@ -0,0 +1,34 @@
...
+package com.amazon.opendistroforelasticsearch.sql.parser.subquery;
+
+public enum SubqueryType {
+    /**
+     * SELECT * FROM A WHERE a IN (SELECT * FROM B)
+     */
+    IN(0),
+    UNSUPPORTED(32);
+
+    private int v;
```
galkk: id?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @param sqlExpr sqlExpr will be rewritten.
+     */
+    public static void rewrite(SQLQueryExpr sqlExpr) {
+        MySqlSelectQueryBlock rootQuery = (MySqlSelectQueryBlock) sqlExpr.getSubQuery().getQuery();
+        final FindSubqueryInWhere subqueryInWhere = new FindSubqueryInWhere();
+        sqlExpr.accept(subqueryInWhere);
+
+        if (subqueryInWhere.subqueryType() == SubqueryType.IN) {
+            final InSubqueryRewriter rewriter = new InSubqueryRewriter(rootQuery,
```
galkk: Feels like this is the place where additional comment would help a reader

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryType.java
```diff
@@ -0,0 +1,34 @@
...
+
+    SubqueryType(int v) {
+        this.v = v;
+    }
+
+    public static boolean isSupported(SubqueryType subqueryType) {
+        return subqueryType.v < SubqueryType.UNSUPPORTED.v;
```
galkk: I like this 👍

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    private void rewrite(Table table, SQLIdentifierExpr expr) {
+        final String tableAlias = table.getAlias();
+        final String tableName = table.getName();
+
+        final String exprName = expr.getName();
+        if (exprName.startsWith(tableName + DOT) || exprName.startsWith(tableAlias + DOT)) {
+            expr.setName(exprName.replace(tableName + DOT, tableAlias+DOT));
```
galkk: inconsistent spacing around +

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
```diff
@@ -0,0 +1,74 @@
...
+        return sqlExistsExprs;
+    }
+
+
+    /**
+     * @return {@link SubqueryType}
+     */
+    public SubqueryType subqueryType() {
+        if (getSqlInSubQueryExprs().size() == 1 &&
+                getSqlExistsExprs().size() == 0 &&
```
galkk: .size() == 0 -> isEmpty()

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    protected void init() throws Exception {
+        loadIndex(Index.ACCOUNT);
+        loadIndex(Index.DOGSSUBQUERY);
+    }
+
+    @Test
+    public void test() throws IOException {
+        String query = String.format("SELECT * FROM %s/dog A JOIN %s/account B ON A.v = B.v",
```
galkk: Those format strings are unnecessary.

1) you need to use Locale there
2) I recommend to inline all index names, and make this query single string. It will help later in understanding what's going on (you can just copy query to kibana) and will give better overview of what's happening

(here and everywhere).

I'm open to conversation about it, but I find the following style of tests

```
        JSONObject aggregations = executeQuery(
                "SELECT date_format(insert_time, 'dd-MM-YYYY') " +
                        "FROM elasticsearch-sql_test_index_online " +
                        "GROUP BY date_format(insert_time, 'dd-MM-YYYY') " +
                        "ORDER BY date_format(insert_time, 'dd-MM-YYYY') DESC")
                .getJSONObject("aggregations");

```

much more readable.

Here and everywhere

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/visitor/FindSubqueryInWhereTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+        assertEquals(SubqueryType.UNSUPPORTED, subqueryInWhere.subqueryType());
+        assertEquals(1, subqueryInWhere.getSqlExistsExprs().size());
+        assertThat(subqueryInWhere.getSqlExistsExprs().get(0), is(instanceOf(SQLExistsExpr.class)));
+        assertFalse(subqueryInWhere.getSqlExistsExprs().get(0).isNot());
+        assertEquals(0, subqueryInWhere.getSqlInSubQueryExprs().size());
+    }
+
+    @Test
+    public void findNOTEXIST() {
```
galkk: The naming of tests is inconsistent with next one. I recommend to rename all of them to Java's camelcase (findExist, findNotIn, findInAtSelect)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+            throw new IllegalStateException("Unsupported subquery with multiple select " + rightQuery.getSelectList());
+        }
+        return rightQuery.getSelectList().get(0).getExpr();
+    }
+
+    private SQLBinaryOpExpr createSubqueryReplacementCondition() {
+        /**
+         * IN subquery is rewritten to JOIN, the null value from the right table should be filtered out.
+         */
+        SQLBinaryOpExpr replacementCond = new SQLBinaryOpExpr(joinRightExpr,
```
galkk: the use of final is inconsistent. Given that replacementCond is no overwritten anywhere in the method, it also could be final. Actually I recommend to get rid of most of those finals, because they clutter the code

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+    private final SQLInSubQueryExpr inSubQueryExpr;
+
+    private final MySqlSelectQueryBlock inSubQuery;
+
+    private SQLExpr joinLeftExpr;
+
+    private SQLExpr joinRightExpr;
+
+    public InSubqueryRewriter(MySqlSelectQueryBlock rootQuery, SQLInSubQueryExpr inSubQueryExpr) {
+        rootQuery.accept0(new SubqueryAliasRewriter());
```
galkk: can we move that rewrite operation into rewriteSubquery? It doesn't seems like there is anything in other parameters that depends on the rootQuery?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+
+        return sqlJoin;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public SQLExpr getWhere() {
+        return rewriteSubquery(rootQuery.getWhere());
```
galkk: What will (or will at all) something happen if I will call getWhere several times?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+        joinLeftExpr = inSubQueryExpr.getExpr();
+        joinRightExpr = rightExpr(inSubQuery);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public SQLJoinTableSource getFrom() {
+        final SQLJoinTableSource sqlJoin = new SQLJoinTableSource();
```
galkk: is there an order of operation that should be maintained? (e.g. call getFrom then getWhere, vice versa, doesn't matter)? Can there be an error?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,148 @@
...
+    private Deque<Table> tableScope =  new ArrayDeque<>();
+
+    private int aliasSuffix = 0;
+
+    private final static String DOT = ".";
+
+    @Override
+    public boolean visit(MySqlSelectQueryBlock query) {
+        final SQLExprTableSource from = (SQLExprTableSource) query.getFrom();
+        String tableName = from.getExpr().toString().replaceAll(" ", "");
```
galkk: same comment about finals. 
tableName isn't being rewritten inside of the method (if I'm not hallucinating), so it also could be final. Does it needs to? I recommend to not overuse finals.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -71,21 +74,22 @@ public static QueryAction create(Client client, String sql) throws SqlParseExcep
                 if(isMulti(sqlExpr)){
                     sqlExpr.accept(new TermFieldRewriter(client, TermRewriterFilter.MULTI_QUERY));
                     MultiQuerySelect multiSelect = new SqlParser().parseMultiSelect((SQLUnionQuery) sqlExpr.getSubQuery().getQuery());
-                    handleSubQueries(client,multiSelect.getFirstSelect());
-                    handleSubQueries(client,multiSelect.getSecondSelect());
                     return new MultiQueryAction(client, multiSelect);
                 }
+                else if(isSubquery(sqlExpr)){
+                    SubqueryRewriterHelper.rewrite(sqlExpr);
```
galkk: This is minor, but why we do not define new rewriter and do not write the statement as a 

    sqlExpr.accept(new SubqueryRewriter())? (for example it's done on the top for NesttedFieldRewriter)?

for more consistent codebase?

this is minor, I'm just curious

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/InSubqueryRewriterTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class InSubqueryRewriterTest {
+
+    @Rule
+    public ExpectedException exceptionRule = ExpectedException.none();
+
+    @Test
+    public void testFrom() {
```
galkk: btw, if you start methods with test, you don't need @Test annotation on them

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/SubqueryRewriterHelperTest.java
```diff
@@ -0,0 +1,49 @@
...
+    public void isSubquery(){
+        assertTrue(SubqueryRewriterHelper.isSubquery(parse("SELECT * FROM A WHERE a IN (SELECT b FROM B)")));
+    }
+
+    @Test
+    public void notSupportedSubquery(){
+        assertFalse(SubqueryRewriterHelper.isSubquery(parse("SELECT * FROM A WHERE a NOT IN (SELECT * FROM B)")));
+    }
+
+    private SQLQueryExpr parse(String sql) {
```
galkk: that method seem to be repeated several times in codebase (QueryPlannerTest, NestedFieldProjectionTest, NetstedFieldRewriterTest, HavingTest) - please make existing implementation public static and reuse it instead of copying it here as well

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/SubqueryRewriterHelperTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class SubqueryRewriterHelperTest {
+
+    @Test
+    public void isSubquery(){
+        assertTrue(SubqueryRewriterHelper.isSubquery(parse("SELECT * FROM A WHERE a IN (SELECT b FROM B)")));
+    }
+
+    @Test
+    public void notSupportedSubquery(){
+        assertFalse(SubqueryRewriterHelper.isSubquery(parse("SELECT * FROM A WHERE a NOT IN (SELECT * FROM B)")));
```
galkk: is it valid query at all, in general (select * from B for IN )? Shouldn't we throw an SqlFeatureNotSupported exception or something along this line in this case?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/rewriter/SubqueryAliasRewriterTest.java
```diff
@@ -0,0 +1,135 @@
...
+                        rewrite(parse(
+                                "SELECT * FROM TbA as A WHERE TbA.a IN (SELECT b FROM TbB ) and TbA.c > 10"))));
+    }
+
+
+    private String sqlString(SQLExpr expr) {
+        return SQLUtils.toMySqlString(expr);
+    }
+
+    private SQLQueryExpr parse(String sql) {
```
galkk: same comment - please reuse some existing method for parsing string to sql instead of repeating it in every test

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Remove the space. Make the style consistent with other class.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,148 @@
...
+        if (from.getAlias() != null) {
+            tableScope.push(new Table(tableName, from.getAlias()));
+        } else {
+            from.setAlias(createAlias(tableName));
+            tableScope.push(new Table(tableName, from.getAlias()));
+        }
+
+        for (SQLSelectItem item : query.getSelectList()) {
+            final SQLExpr itemExpr = item.getExpr();
+            if (itemExpr instanceof SQLIdentifierExpr) {
```
dai-chen: I think it would be cleaner to implement `visit(SQLIdentifierExpr)` because I notice you're rewriting it in multiple places in this class. Does the following sample code make sense?

```
    @Override
    public boolean visit(SQLIdentifierExpr expr) {
        if (isInSelect(expr) || isInWhere(expr) || isInSubquery(expr) {
            rewrite(tableScope.peek(), expr);
        }
        return true;
    }
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,148 @@
...
+            rewrite(tableScope.peek(), (SQLIdentifierExpr)expr.getExpr());
+        }
+    }
+
+    private void rewrite(Table table, SQLIdentifierExpr expr) {
+        final String tableAlias = table.getAlias();
+        final String tableName = table.getName();
+
+        final String exprName = expr.getName();
+        if (exprName.startsWith(tableName + DOT) || exprName.startsWith(tableAlias + DOT)) {
```
dai-chen: Need a little clarification. So even if it starts with "`tableAlias.`", we replace "`tableName.`" with "`tableAlias.`"?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+    private SQLExpr rightExpr(MySqlSelectQueryBlock rightQuery) {
+        if (rightQuery.getSelectList().size() > 1) {
+            throw new IllegalStateException("Unsupported subquery with multiple select " + rightQuery.getSelectList());
+        }
+        return rightQuery.getSelectList().get(0).getExpr();
+    }
+
+    private SQLBinaryOpExpr createSubqueryReplacementCondition() {
+        /**
+         * IN subquery is rewritten to JOIN, the null value from the right table should be filtered out.
```
dai-chen: I wonder why filter out NULL from right table is required? Because you're transforming subquery to inner join. Is this logic reserved for left join transform in future?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Agree. It is a generic opeartion for all the subquery clause. Move it to SubqueryRewriterHelper.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: It is NOT idempotent API. Refactor the code.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Unexpected result.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Enable the unnecessary final checking in Idea and remove unnecessary final for local var and para.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryType.java
penghuo: Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
```diff
@@ -0,0 +1,138 @@
...
+                TEST_INDEX_DOGSUBQUERY, TEST_INDEX_ACCOUNT);
+        System.out.println(explainQuery(query));
```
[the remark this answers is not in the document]
penghuo: remove it.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
penghuo: Done. reformat the query.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/InSubqueryRewriterTest.java
penghuo: Keep @Test for consistency

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/SubqueryRewriterHelperTest.java
penghuo: Extract to common test utils class. SqlParserUtils

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
penghuo: Good question. This is actually my first choice. the major concern is this solution change the AST before the parser know whether it can support it or not. 
There are two reasons why I come with the proposed solution
- The soluiton define two steps. (1) Is it supported subquery (2) if yes, rewrite the subquery. which clearly define the query is supported or not and this structure make it easy to add more subquery use case in future
- In my understanding, we should not change the original AST, but create and optimize the logical plan for different query type. for example, SELECT * FROM A WHERE a IN (SELECT b FROM B), The AST should represent the subquery, but the logical plan should be JOIN. This soluiton make it on the right way.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public class InSubqueryRewriter implements SubqueryRewriter {
+
+    private final MySqlSelectQueryBlock rootQuery;
+
+    private final SQLInSubQueryExpr inSubQueryExpr;
+
+    private final MySqlSelectQueryBlock inSubQuery;
+
+    private SQLExpr joinLeftExpr;
```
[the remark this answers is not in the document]
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Yes. Two purpose, (1) change to LEFT JOIN in future. (2) simply the logic merge with existing where clause.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryType.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
penghuo: Good suggestion! Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/SubqueryRewriterHelperTest.java
penghuo: Change to this SELECT * FROM A WHERE a NOT IN (SELECT b FROM B)

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/subquery/rewriter/SubqueryAliasRewriterTest.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
penghuo: Yes.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    /**
+     * Does the query include the supported subquery in where.
+     * Currently, ONLY IN is supported.
+     *
+     * @return true if it is yes, otherwise false.
+     */
+    public static boolean isSubquery(SQLQueryExpr sqlExpr) {
```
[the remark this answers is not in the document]
penghuo: Done. Throw SQLFeatureNotSupportedException when the subquery is not supported yet.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
```diff
@@ -0,0 +1,60 @@
...
+public class FindSubqueryInWhere extends MySqlASTVisitorAdapter {
+    private final List<SQLInSubQueryExpr> sqlInSubQueryExprs = new ArrayList<>();
+    private final List<SQLExistsExpr> sqlExistsExprs = new ArrayList<>();
+
+    /**
+     * Return Subquery
+     * @return {@link Subquery}
+     */
+    public Subquery subquery() {
+        if (sqlExistsExprs.isEmpty() && sqlInSubQueryExprs.isEmpty()) return new Subquery(SubqueryType.NOT_SUBQUERY);
```
arsen-es: To me this is a code smell: it is illogical to create a subquery which is not a subquery. I didn't go through the whole change yet, will probably come back to this, but that's my initial reaction.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Good point. In case this is no state transition, I add the tree transition. Add the state transition for

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriter.java
```diff
@@ -0,0 +1,30 @@
...
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.parser.subquery;
+
+import com.alibaba.druid.sql.dialect.mysql.ast.statement.MySqlSelectQueryBlock;
+
+/**
+ * Rewrite the Predicate Subquery to JOIN.
```
arsen-es: I think the documentation should say "rewrites the query to replace the subquery with a join". It does not re-write just the subquery part.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriter.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public interface SubqueryRewriter {
+
+    /**
+     * Rewrite the subquery with correspond JOIN query.
+     * It is NOT idempotent operation.
+     */
+    void rewrite(MySqlSelectQueryBlock query);
```
arsen-es: Why not keep the input parameter intact and return the re-written one as a result?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,72 @@
...
+import com.alibaba.druid.sql.ast.expr.SQLQueryExpr;
+import com.alibaba.druid.sql.dialect.mysql.ast.statement.MySqlSelectQueryBlock;
+import com.amazon.opendistroforelasticsearch.sql.parser.subquery.model.SubqueryType;
+import com.amazon.opendistroforelasticsearch.sql.parser.subquery.rewriter.SubqueryAliasRewriter;
+import com.amazon.opendistroforelasticsearch.sql.parser.subquery.visitor.FindSubqueryInWhere;
+
+import java.sql.SQLFeatureNotSupportedException;
+
+/**
+ * {@link SubqueryRewriter} Helper Function.
```
arsen-es: "Provides helper functions for {@link SubqueryRewriter}."

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriter.java
penghuo: I think it is not necessary. In this case, we need copy the MySqlSelectQueryBlock.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriter.java
penghuo: Agree. Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Rewrite the {@link SQLQueryExpr}'s from and where clause.
+     *
+     * @param sqlExpr sqlExpr will be rewritten.
+     */
+    public static void rewrite(SQLQueryExpr sqlExpr) {
+        MySqlSelectQueryBlock rootQuery = (MySqlSelectQueryBlock) sqlExpr.getSubQuery().getQuery();
+        final FindSubqueryInWhere subqueryInWhere = new FindSubqueryInWhere();
+        sqlExpr.accept(subqueryInWhere);
```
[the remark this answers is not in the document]
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+ *  SELECT * FROM A WHERE a IN (SELECT b FROM B) and c > 10 should be rewritten to
+ *  SELECT A.* FROM A JOIN B ON A.a = B.b WHERE c > 10 and B.b IS NOT NULL.
+ */
+public class InSubqueryRewriter implements SubqueryRewriter {
+    private final SQLInSubQueryExpr inSubQueryExpr;
+    private final MySqlSelectQueryBlock inSubQuery;
+    private final SQLExpr joinLeftExpr;
+    private final SQLExpr joinRightExpr;
+
+    public InSubqueryRewriter(Subquery subquery) {
```
arsen-es: I think this method accepts the query, not the subquery.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+     *                       SELECT FORM
+     *                         |     |
+     *                         b     B
+     *
+     * The FROM logic should be
+     *     JOIN
+     *     / | \
+     *    A  B A.a == B.b
+     */
+    private SQLJoinTableSource buildFrom(MySqlSelectQueryBlock rootQuery) {
```
arsen-es: np: I would call this `buildFromClause` and the other one `buildWhereClause` to avoid potential confusion with this method being a factory method for the class, which builds an instance from the argument.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+    private SQLExpr buildWhere(MySqlSelectQueryBlock rootQuery) {
+        return rewriteSubquery(rootQuery.getWhere());
+    }
+
+    private SQLExpr rewriteSubquery(SQLExpr expr) {
+        if (expr instanceof SQLInSubQueryExpr) {
+            return createSubqueryReplacementCondition();
+        }
+        else if (expr instanceof SQLBinaryOpExpr) {
+            ((SQLBinaryOpExpr) expr).setLeft(rewriteSubquery(((SQLBinaryOpExpr) expr).getLeft()));
```
arsen-es: store `(SQLBinaryOpExpr) expr` in a variable?

arsen-es: I don't like the fact that we keep modifying methods' arguments instead of returning the result. It might not be obvious or might not even cause issues now, but it will be difficult later to track down where things have changed. Is the copy operation of the SQLBinaryOpExpr expensive?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+            ((SQLBinaryOpExpr) expr).setRight(rewriteSubquery(((SQLBinaryOpExpr) expr).getRight()));
+            return expr;
```
arsen-es: Can you `return expr;` out of if-else blocks and only once?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+    private SQLExpr rightExpr(MySqlSelectQueryBlock rightQuery) {
+        if (rightQuery.getSelectList().size() > 1) {
+            throw new IllegalStateException("Unsupported subquery with multiple select " + rightQuery.getSelectList());
+        }
+        return rightQuery.getSelectList().get(0).getExpr();
+    }
+
+    private SQLBinaryOpExpr createSubqueryReplacementCondition() {
+        // IN subquery is rewritten to JOIN, the null value from the right table should be filtered out.
+        SQLBinaryOpExpr replacementCond = new SQLBinaryOpExpr(joinRightExpr,
```
arsen-es: np: `replacementCond` -> `condition`

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+     * @param query Subquery will be written.
+     */
+    @Override
+    public void rewrite(MySqlSelectQueryBlock query) {
+        query.setFrom(buildFrom(query));
+        query.setWhere(buildWhere(query));
+    }
+
+    /**
+     * Build From clause from input query.
```
galkk: I like amount of the documentation that you've added!

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
galkk: This is private method, so I don't think that there might be much of a confusion

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
galkk: I have similar feelings to Arsen, and also would like to add that it feels like additional comment is needed here.

I'm looking into usage

```
            if (SubqueryType.IN == subquery.getSubqueryType()) {
                InSubqueryRewriter rewriter = new InSubqueryRewriter(subquery);
                rewriter.rewrite(rootQuery);
            }
```

And it feels like it would be benefecial for a reader to add comment that actually will say:
this is root query: SELECT * FROM A WHERE a IN (SELECT b FROM B) 
and this is subquery (or whatever the subquery is): a IN (SELECT b FROM B)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
galkk: I support existing code as it is. 
Only one exit makes code harder to read
https://blog.codinghorror.com/flattening-arrow-code/

> Always opportunistically return as soon as possible from the function. Once your work is done, get the heck out of there! This isn't always possible – you might have resources you need to clean up. But whatever you do, you have to abandon the ill-conceived idea that there should only be one exit point at the bottom of the function.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class SubqueryRewriterHelper {
+
+    /**
+     * Rewrite the {@link SQLQueryExpr}'s from and where clause.
+     *
+     * @param sqlExpr sqlExpr will be rewritten.
+     */
+    public static void rewrite(SQLQueryExpr sqlExpr, Subquery subquery) throws SQLFeatureNotSupportedException {
+        if (isSubquery(subquery)) {
```
galkk: minor issue, but I recomment to decrease indentation by

if (!isSubquery(subquery)) {
   return;
}

...

And the naming of parameter and type here is confusing: We already passed Subquery as a type and subquery as parameter name, so the condition 

isSubquery(subquery : Subquery) 

feels very unnatural. Maybe isSupportedSubquery is better name?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * @return true if it is yes, otherwise false.
+     */
+    public static boolean isSubquery(Subquery subquery) throws SQLFeatureNotSupportedException {
+        SubqueryType subqueryType = subquery.getSubqueryType();
+        if (subqueryType.isNotSubquery()) return false;
+        else if (subqueryType.isSupported()) {
+            return true;
+        } else {
+            throw new SQLFeatureNotSupportedException("Unsupported subquery");
```
galkk: Add more information to the exception message: 
("Unsupported subquery: only IN is supported, %s", subqueryExpression)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Does the query include the supported subquery in where.
+     * Currently, ONLY IN is supported.
+     *
+     * @return true if it is yes, otherwise false.
+     */
+    public static boolean isSubquery(Subquery subquery) throws SQLFeatureNotSupportedException {
+        SubqueryType subqueryType = subquery.getSubqueryType();
+        if (subqueryType.isNotSubquery()) return false;
```
galkk: the style is inconsistent.
Consider changing to 

if (subqueryType.isNotSubquery()) {
    return false;
} else if ...

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/model/SubqueryType.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.amazon.opendistroforelasticsearch.sql.parser.subquery.model;
+
+public enum SubqueryType {
+    /**
+     * SELECT * FROM A WHERE a IN (SELECT * FROM B)
+     */
+    IN(0),
+    UNSUPPORTED(32),
+    NOT_SUBQUERY(33);
```
galkk: I second comment that Arsen has made earlier. That 
NOT_SUBQUERY feels very unnatural. I think that it shouldn't be a part of this enum. The check if query is not supported should be made earlier (e.g. SubqueryRewriterHelper) / elsewhere (e.g. extract that logic to separate class), and reuse it.

In this case everwhere code will get Subquery, it will know that this is a subquery, without additional checks.

In it's current form it reminds me a meme

enum Boolean {
    TRUE,
    FALSE,
    FILE_NOT_FOUND
}

:)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,171 @@
...
+    private boolean inSelect(SQLAllColumnExpr expr) {
+        return expr.getParent() instanceof SQLSelectItem;
+    }
+
+    private boolean inWhere(SQLIdentifierExpr expr) {
+        return expr.getParent() instanceof SQLBinaryOpExpr && !isESTable((SQLBinaryOpExpr) expr.getParent());
+    }
+
+    /**
+     * The table name in elasticsaerch could be "index/type". Which represent as SQLBinaryOpExpr in AST.
```
galkk: typo: saerch

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private Integer next() {
+        return aliasSuffix++;
+    }
+
+    /**
+     * Table Bean.
+     */
+    private static class Table {
```
galkk: minor, personal preference, feel free to ignore (although I recommend to remove empty line before first member):

The formatting of this class feels a bit unnatural (getters in the beginning, field after, javadocs over private fields). Consider.

```
   /**
     * Table Bean.
     */
    private static class Table {
        private String name;
        private String alias;

         public Table(String name, String alias) {
            this.name = name;
            this.alias = alias;
        }

         /** @return name of the table */
         public String getName() {
            return name;
        }

         /** @return alias of the table */
         public String getAlias() {
            return alias;
        }
    }
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
galkk: +1. That should be just a method:

isSubquery(sqlExpr)

and the rest of the code should not be even entered

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
```diff
@@ -0,0 +1,60 @@
...
+        if (sqlInSubQueryExprs.size() == 1 &&
+                sqlExistsExprs.isEmpty() &&
+                !sqlInSubQueryExprs.get(0).isNot()) {
+            return new Subquery(SubqueryType.IN, sqlInSubQueryExprs.get(0));
+        }
+        return new Subquery(SubqueryType.UNSUPPORTED);
```
galkk: Can we just throw exception right here, without passing that UNSUPPORTED to the rest of the code (where we actually do that lateR)

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -107,26 +103,8 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
-
-                    final ThreadPool threadPool = client.threadPool();
-                    threadPool.schedule(threadPool.preserveContext(requestExecution), new TimeValue(0L),
-                            AsyncRestExecutor.SQL_WORKER_THREAD_POOL_NAME);
-                };
-            } else {
-                final QueryAction queryAction = explainRequest(client, sqlRequest);
-                return channel -> executeSqlRequest(request, queryAction, client, channel);
-            }
+            final QueryAction queryAction = explainRequest(client, sqlRequest);
```
galkk: That change doesn't seem to relate to the rest of the PR, so: why did all the code on the left go?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
penghuo: Done. Remove the UNSUPPORTED.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Move the subquery param from the constructor to function interface.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/rewriter/SubqueryAliasRewriter.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/visitor/FindSubqueryInWhere.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/model/SubqueryType.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
arsen-es: It's subjective, but to me naming and access are orthogonal concerns.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+     * With the input query.
+     *       Query
+     *     /   |   \
+     * SELECT FROM     WHERE
+     *    |    |     /   |  \
+     *    *    A  c>10 AND INSubquery
+     *                      /    \
+     *                     a    Query
+     *                          /    \
+     *                       SELECT FORM
```
abbashus: typo: FROM

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Understood. In case there are several vistor follow the same pattern. I think it should be fine to keep this.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
arsen-es: @galkk  I wouldn't take someone's recommendation as a hard rule to follow in every case. I prefer avoiding duplicated code, if you need to return a value, but conditionally do something before that, why return that value twice, just do whatever you want to do conditionally and return that value. The only thing that makes me agree to current code is that there is another condition above (the `if` block) that I somehow missed.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/SubqueryRewriterHelper.java
penghuo: Done. But ignore the subqueryExpression. It is a little bit redundant.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/subquery/InSubqueryRewriter.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+            ((SQLBinaryOpExpr) expr).setRight(rewriteSubquery(((SQLBinaryOpExpr) expr).getRight()));
+            return expr;
+        } else {
+            return expr;
+        }
+    }
+
+    private SQLExpr rightExpr(MySqlSelectQueryBlock rightQuery) {
+        if (rightQuery.getSelectList().size() > 1) {
+            throw new IllegalStateException("Unsupported subquery with multiple select " + rightQuery.getSelectList());
```
arsen-es: Do you consider this case being a bug in our code? If it's possible to get to this from invalid user input, then you should consider changing the exception type to one of the exceptions that are considered client errors (see in RestSqlAction.java).

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/InSubqueryRewriter.java
```diff
@@ -0,0 +1,146 @@
...
+     *     WHERE
+     *     /  |   \
+     *   c>10 AND B.b is NOT NULL
+     */
+    private SQLExpr buildWhere(MySqlSelectQueryBlock rootQuery) {
+        return rewriteSubquery(rootQuery.getWhere());
+    }
+
+    private SQLExpr rewriteSubquery(SQLExpr expr) {
+        if (expr instanceof SQLInSubQueryExpr) {
```
arsen-es: np: I still think this is cleaner:

```java
        if (expr instanceof SQLInSubQueryExpr) {
            return createSubqueryReplacementCondition();
        } else {
            if (expr instanceof SQLBinaryOpExpr) {
                SQLBinaryOpExpre binaryExpr = (SQLBinaryOpExpr) expr;
                binaryExpr.setLeft(rewriteSubquery((binaryExpr).getLeft()));
                binaryExpr.setRight(rewriteSubquery((binaryExpr.getRight()));
            }
            return expr;
        }
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/Optimizer.java
```diff
@@ -0,0 +1,66 @@
...
+        public BuilderOptimizer<T> withRule(OptimizeRule<T> rule) {
+            if (optimizerRules == null) {
+                optimizerRules = new ArrayList<>();
+            }
+            optimizerRules.add(rule);
+            return this;
+        }
+
+        public Optimizer<T> build() {
+            return new Optimizer<>(optimizerRules);
```
arsen-es: np: should this clear the `optimizerRules` before returning the optimizer?

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/Optimizer.java
```diff
@@ -0,0 +1,66 @@
...
+     * Build {@link Optimizer}
+     */
+    public static <T> BuilderOptimizer<T> builder() {
+        return new BuilderOptimizer<T>();
+    }
+
+    /**
+     * Builder of {@link Optimizer}
+     */
+    public static class BuilderOptimizer<T> {
```
arsen-es: just `Builder` since it is already a nested class. Also, should it be parameterized again? can't you do something like `Optimizer<MyExprType>.Builder`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/OptimizeRule.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.optimizer;
+
+import java.sql.SQLFeatureNotSupportedException;
+
+/**
+ * Query Optimize Rule
+ */
+public interface OptimizeRule<T> {
```
arsen-es: class and interface names should be objects/nouns, not verbs (which would denote a method). Should this be `OptimizationRule` or `OptimizerRule` or something else?

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/OptimizeRule.java
```diff
@@ -0,0 +1,39 @@
...
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.optimizer;
+
+import java.sql.SQLFeatureNotSupportedException;
+
+/**
+ * Query Optimize Rule
```
arsen-es: This documentation is not very useful, either give more info what it does, or delete completely if the interface is pretty straightforward.

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/SubqueryOptimizeRule.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public boolean match(SQLQueryExpr expr) throws SQLFeatureNotSupportedException {
+        subquery = subquery(expr);
+        if (subquery.isPresent()) {
+            SubqueryType subqueryType = subquery.get().getSubqueryType();
+            if (subqueryType.isSupported()) {
+                return true;
+            } else {
+                throw new SQLFeatureNotSupportedException("Unsupported subquery. only IN is supported.");
```
arsen-es: np: "only IN is supported" requires maintenance, ideally this message should be built using all the values except the unsupported.

arsen-es: If the method returns a boolean result, to me the contract requires this case to just return false. I understand the intention, but I think the rule should not handle that error, some other higher level class should take care of that.

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/SubqueryOptimizeRule.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    public void optimize(SQLQueryExpr expr) {
+        if (subquery == null) subquery = subquery(expr);
```
arsen-es: Please use curly braces even for a one-liner.

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/OptimizeRule.java
galkk: 1. I think that this is not Optimize, but RewriteRule
2. does it needs to be generic? Can we put more restrictions to T (e.g. SQLExpr, at least)?

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/model/SubqueryType.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.optimizer.subquery.model;
+
+public enum SubqueryType {
```
arsen-es: I am thinking if it is better to put this in the `Subquery` class as a static nested type: `Subquery.Type`

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/model/SubqueryType.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * SELECT * FROM A WHERE a IN (SELECT * FROM B)
+     */
+    IN(0),
+    UNSUPPORTED(32);
```
arsen-es: I am ok with leaving this for now, but logically unsupported does not represent a type, we could have types A, B, C, D, E, where B and E are unsupported.

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/model/SubqueryType.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private int id;
+
+    SubqueryType(int id) {
+        this.id = id;
+    }
+
+    public boolean isSupported() {
+        return id < SubqueryType.UNSUPPORTED.id;
```
arsen-es: isn't it safe to just do `return this != UNSUPPORTED;` ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,171 @@
...
+    @Override
+    public boolean visit(MySqlSelectQueryBlock query) {
+        SQLExprTableSource from = (SQLExprTableSource) query.getFrom();
+        String tableName = from.getExpr().toString().replaceAll(" ", "");
+
+        if (from.getAlias() != null) {
+            tableScope.push(new Table(tableName, from.getAlias()));
+        } else {
+            from.setAlias(createAlias(tableName));
+            tableScope.push(new Table(tableName, from.getAlias()));
```
arsen-es: duplicated line on #66 and #69, so it can be just
```java
if (from.getAlias() == null) {
    from.setAlias(createAlias(tableName));
}
tableScope.push(new Table(tableName, from.getAlias()));
return true;
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,171 @@
...
+    public boolean visit(SQLIdentifierExpr expr) {
+        if (inSelect(expr) || inWhere(expr) || inSubquery(expr)) {
+            rewrite(tableScope.peek(), expr);
+        }
+        return true;
+    }
+
+    @Override
+    public boolean visit(SQLAllColumnExpr expr) {
+        if (inSelect(expr) ) {
```
arsen-es: extra space

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/rewriter/SubqueryAliasRewriter.java
```diff
@@ -0,0 +1,171 @@
...
+    public void endVisit(MySqlSelectQueryBlock query) {
+        tableScope.pop();
+    }
+
+    private void rewrite(Table table, SQLIdentifierExpr expr) {
+        String tableAlias = table.getAlias();
+        String tableName = table.getName();
+
+        String exprName = expr.getName();
+        if (exprName.startsWith(tableName + DOT) || exprName.startsWith(tableAlias + DOT)) {
```
arsen-es: why do we need the second condition? can you add a comment for that

src/main/java/com/amazon/opendistroforelasticsearch/sql/optimizer/subquery/InSubqueryRewriter.java
galkk: Looking into the logic, we can simplify even more

````
if (expr instanceof SQLInSubQueryExpr) {
            return createSubqueryReplacementCondition();
        } 

if (expr instanceof SQLBinaryOpExpr) {
    SQLBinaryOpExpre binaryExpr = (SQLBinaryOpExpr) expr;
    binaryExpr.setLeft(rewriteSubquery((binaryExpr).getLeft()));
    binaryExpr.setRight(rewriteSubquery((binaryExpr.getRight()));
}
return expr;
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/RewriteRuleExecutor.java
```diff
@@ -0,0 +1,68 @@
...
+    public RewriteRuleExecutor(List<RewriteRule<T>> rewriteRules) {
+        this.rewriteRules = rewriteRules;
+    }
+
+    /**
+     * Execute the registered {@link RewriteRule} in order on the Query.
+     */
+    public void executeOn(T expr) throws SQLFeatureNotSupportedException {
+        for (RewriteRule<T> rule : rewriteRules) {
+            if (rule.match(expr)) rule.rewrite(expr);
```
arsen-es: np: add curly braces

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/RewriteRuleExecutor.java
```diff
@@ -0,0 +1,68 @@
...
+     * Build {@link RewriteRuleExecutor}
+     */
+    public static <T extends SQLQueryExpr> BuilderOptimizer<T> builder() {
+        return new BuilderOptimizer<T>();
+    }
+
+    /**
+     * Builder of {@link RewriteRuleExecutor}
+     */
+    public static class BuilderOptimizer<T extends SQLQueryExpr> {
```
arsen-es: just call this class `Builder`

arsen-es: I think you don't need to re-specify the parameter type, you can access T inside the inner class.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/RewriteRuleExecutor.java
```diff
@@ -0,0 +1,68 @@
...
+import com.alibaba.druid.sql.ast.expr.SQLQueryExpr;
+
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Query RewriteRuleExecutor which will execute the {@link RewriteRule} with registered order.
+ */
+public class RewriteRuleExecutor<T extends SQLQueryExpr> {
```
arsen-es: I think this class should not be generic at all, only the internal collection should be generic, because that way you can have rules for different types of expressions instead of one fixed type.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/subquery/SubqueryRewriteRule.java
```diff
@@ -0,0 +1,75 @@
...
+                throw new SQLFeatureNotSupportedException("Unsupported subquery. only IN is supported.");
+            }
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    public void rewrite(SQLQueryExpr expr) {
+        if (subquery == null) subquery = subquery(expr);
```
arsen-es: curly braces

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/subquery/visitor/FindSubqueryInWhere.java
```diff
@@ -0,0 +1,61 @@
...
+public class FindSubqueryInWhere extends MySqlASTVisitorAdapter {
+    private final List<SQLInSubQueryExpr> sqlInSubQueryExprs = new ArrayList<>();
+    private final List<SQLExistsExpr> sqlExistsExprs = new ArrayList<>();
+
+    /**
+     * Return Subquery
+     * @return {@link Subquery}
+     */
+    public Optional<Subquery> subquery() {
+        if (sqlExistsExprs.isEmpty() && sqlInSubQueryExprs.isEmpty()) return Optional.empty();
```
arsen-es: curly braces

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/subquery/visitor/FindSubqueryInWhere.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class FindSubqueryInWhere extends MySqlASTVisitorAdapter {
+    private final List<SQLInSubQueryExpr> sqlInSubQueryExprs = new ArrayList<>();
+    private final List<SQLExistsExpr> sqlExistsExprs = new ArrayList<>();
+
+    /**
+     * Return Subquery
+     * @return {@link Subquery}
+     */
+    public Optional<Subquery> subquery() {
```
arsen-es: `find()` ?

Issue #125, opened 2019-07-20
Elasticsearch 7.2 compatibility


dai-chen:
Make the plugin compatible with Elasticsearch 7.2

PR #91, opened 2019-06-21
Upgrading ES version to 7.1.1

arsen-es:
*Issue #, if available:* #90 

*Description of changes:* Upgrading es version to 7.1.1. Nothing breaks, so no additional changes.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #129, opened 2019-07-23
Es 7.2 release

galkk:
*Issue #, if available:* #125 

*Description of changes:*
Plugin release for ElasticSearch 7.2.0

Testing:
unit/integration tests, manual spot checks using elasticsearch-oss-7.2.0-linux-x86_64.tar.gz


src/test/java/com/amazon/opendistroforelasticsearch/sql/util/HasFieldWithValue.java
```diff
@@ -43,7 +42,6 @@ private HasFieldWithValue(String name, String desc, Matcher<? super U> matcher)
         this.fieldName = name;
     }
 
-    @Factory
```
abbashus: Sorry, not very familiar with this annotation. What is the reason we are removing this?

opendistro-elasticsearch-sql.release-notes
```diff
@@ -1,4 +1,17 @@
-## 2019-06-21, Version 1.1.0 (Current)
+## 2019-07-23, Version 1.2.0
```
abbashus: Should we add `(Current)` here as well ?

gradlew
```diff
@@ -28,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
abbashus: Are theses JVM options auto-generated by Gradle, or we added it explicitly?

galkk: Autogenerated by Gradle

opendistro-elasticsearch-sql.release-notes
galkk: Good catch, done

src/test/java/com/amazon/opendistroforelasticsearch/sql/util/HasFieldWithValue.java
galkk: Elastic has migrated to new version of harmcrest and this annotation was deprecated.
Seems like this is some old annotation that was used in previous versions of harmcrest. 

As per harmcrest doc:

https://github.com/hamcrest/JavaHamcrest/blob/master/CHANGES.md
org.hamcrest.Factory has been removed. This was only used in old implementations of the hamcrest build toolchain, and can safely be deleted from client code without any effect.

abbashus:
Thanks. Looks good to me!

build.gradle
```diff
@@ -46,7 +46,7 @@ repositories {
 }
 
 ext {
-    opendistroVersion = '1.1.0'
+    opendistroVersion = '1.2.0'
```
dai-chen: The same for the version in `gradle.properties`? https://github.com/opendistro-for-elasticsearch/sql/pull/91/files#diff-503f218d646c10f484fdc9d6315bf2e3

PR #135, opened 2019-07-31
Fix single condition results for text+keyword field for nested query

abbashus:
**Description of changes:**
Simply changing the identifier generated incorrect nested path leading to error.
This change fixes that by deriving the correct nested path and preserving in nested function before re-written by the TermFieldRewriter.

**Testing:**
Gradle build passing.
Existing unit and integration test passing.
Modified test in `NestedFieldRewriterTest` to reflect new changes
Added new integration tests. 


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
```diff
@@ -275,6 +278,65 @@ public void testJoinAliasMissing() throws IOException {
...
+        String result = explainQuery(String.format(Locale.ROOT,
+            "SELECT e.id, p.name FROM %s e, e.projects p GROUP BY  p.name ",
+            TEST_INDEX_EMPLOYEE_NESTED));
+        assertThat(result, containsString("\"terms\":{\"field\":\"projects.name.keyword\""));
+        assertThat(result, containsString("\"nested\":{\"path\":\"projects\""));
+    }
+
+    @Test
+    public void testNestedSingleOrderBy() throws IOException {
+        String result = explainQuery(String.format(Locale.ROOT,
```
galkk: consider inlining index name and getting rid of string format entirely

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
```diff
@@ -91,4 +101,10 @@ private SQLMethodInvokeExpr wrapNestedFunction(SQLExpr expr) {
         return nestedFunc;
     }
 
+    String pathFromIdentifier(SQLExpr identifier) {
+        String field = Util.extendedToString(identifier);
+        int lastDot = field.lastIndexOf(".");
```
galkk: those indexOfs have bitten us earlier. Are we 100% sure that field will always have a . in it?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
abbashus: I am all for it, just wanted to follow the style of other tests in this testsuite .

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: Though I am 99.9% sure we will atleast `.`, but let me rethink any scenario where this would not be the case.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
```diff
@@ -47,8 +49,12 @@
      */
     abstract void rewrite(Scope scope);
 
-    /** Replace expr by nested(expr) and set pointer in parent properly */
     SQLMethodInvokeExpr replaceByNestedFunction(SQLExpr expr) {
+        return replaceByNestedFunction(expr, null);
```
dai-chen: Because `nestedFunc` is returned, could we do this in the opposite way to avoid passing `null` explicitly?

```
    SQLMethodInvokeExpr replaceByNestedFunction(SQLExpr expr, String path) {
        SQLMethodInvokeExpr nestedFunc = replaceByNestedFunction(expr);
        nestedFunc.getParameters().add(1, new SQLCharExpr(path));
        return nestedFunc;
    }

    SQLMethodInvokeExpr replaceByNestedFunction(SQLExpr expr) {
       // same as original...
    }

File: src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java [hunk 4]
```diff
@@ -214,14 +215,23 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
-            );
+                validIdentifierHelper(expr)
+                || (
+                    expr.getParent() instanceof SQLMethodInvokeExpr
+                    && validIdentifierHelper((SQLMethodInvokeExpr) expr.getParent())
+                    )
+               );
+    }
+
+    private boolean validIdentifierHelper(SQLExpr expr) {
```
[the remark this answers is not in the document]
penghuo: +1. we could also have another method for SQLMethodInvokeExpr.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
```diff
@@ -81,6 +87,10 @@ else if (parent instanceof MySqlSelectQueryBlock) {
         else {
             throw new IllegalStateException("Unsupported place to use nested field under parent: " + parent);
         }
+
+        if (path != null) {
+            nestedFunc.getParameters().add(1, new SQLCharExpr(path));
```
penghuo: why the SQLCharExpr is added to the specified position in the list? is there any test cases to cover this?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/NestedFieldRewriterTest.java
```diff
@@ -58,47 +58,47 @@ public void selectWithoutFrom() {
     public void selectAll() {
         same(
             query("SELECT * FROM team t, t.employees"),
-            query("SELECT *, nested(employees.*) FROM team")
+            query("SELECT *, nested(employees.*, 'employees') FROM team")
```
penghuo: do we still support "SELECT *, nested(employees.*) FROM team"? If yes, do we need to have test cases for it?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: Yes, this seems much better.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/NestedFieldRewriterTest.java
abbashus: We don't explicitly support this syntax but the SQL++ one, but it still works. These tests checks whether the new AST generated from NestedRewriter matches the AST generated by explicitly passing nested function.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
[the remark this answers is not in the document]
abbashus: @dai-chen  `isValidIdentifier` seems more ambiguous. Too me current name looks appropriate to intended function. Can you suggest a better one, I am happy to change?

@penghuo Apologies. Did not understand why we need another method for `SQLMethodInvokeExpr`.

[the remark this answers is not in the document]
penghuo: I mean another function to encapsulation
```
expr.getParent() instanceof SQLMethodInvokeExpr  && validIdentifierHelper((SQLMethodInvokeExpr) expr.getParent())
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -214,14 +215,23 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+        return
```
arsen-es: Why not include `!expr.getName().startsWith("_")` in this method? Seems that's also a requirement for validity.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
arsen-es: also, please do not just put `1` constant, assign it to a local final variable with descriptive name, something like `final int pathParameterIndex = 1;`

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
arsen-es: seems Chen's PR earlier had some special case with multi-field columns/fields.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
```diff
@@ -275,6 +278,65 @@ public void testJoinAliasMissing() throws IOException {
 
     }
 
+    @Test
+    public void testNestedSingleConditionAllFields() throws IOException {
+        String result = explainQuery(String.format(Locale.ROOT,
```
arsen-es: we have `StringUtils.format` to avoid specifying the locale, but I think Andy wants us to avoid using format for straightforward queries. Up to you which one you pick, but if you keep format, use ours.

abbashus: Will be inlining index name and getting rid of string format entirely a suggested by @galkk

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
abbashus: 1. This logic is already being checked in `isValidIdentifierForTerm`.
2. It applies only to `SQLIdentifierExpr`.

The whole intention of this function is to de-duplicate code.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
abbashus: Done, made them all inline.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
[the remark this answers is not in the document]
abbashus: @penghuo Extracted into another method `checkIfNestedIdentifier()`

@dai-chen Going with the suggested method name `isValidIdentifier() ` , its less verbose.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: @arsen-es Done.

@penghuo The unittests [here](https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/NestedFieldRewriterTest.java) goes through AST match check.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -213,15 +214,25 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
+    private boolean checkIfNestedIdentifier(SQLIdentifierExpr expr) {
+        return
+            expr.getParent() instanceof SQLMethodInvokeExpr
+            && ((SQLMethodInvokeExpr) expr.getParent()).getMethodName().equals("nested")
+            && isValidIdentifier((SQLMethodInvokeExpr) expr.getParent());
+    }
+
+    private boolean isValidIdentifier(SQLExpr expr) {
+        return
+            ( expr.getParent() instanceof SQLBinaryOpExpr
```
arsen-es: np: seems like assigning `expr.getParent()` to variable would make this method body much simpler.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
```diff
@@ -47,6 +50,12 @@
      */
     abstract void rewrite(Scope scope);
 
+    SQLMethodInvokeExpr replaceByNestedFunction(SQLExpr expr, String nestedPath) {
```
arsen-es: np: to me "replace with" sounds more a bit more explicit ("replace by" nested function might be confusing in a sense that the replacement is done by the function).

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
```diff
@@ -36,6 +38,7 @@
 abstract class SQLClause<T> {
 
     protected final T expr;
+    final int nestedPathIndex = 1;
```
arsen-es: I was saying put this in the method as final var. It's too specific to be on the class level, however it's still ok, but then you must make it static.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
arsen-es: np: I am a little skeptical about this one: it's fine not to use format, because the index names do not have any special characters which depend on the locale, but I still like keeping them in variables. In a case if we need to change the name, we would change in one place, instead of doing find and replace across the file. With static vars, just looking at the test, you can quickly identify which index(es) the test operates on because of the all caps naming, whereas with index names you need to scan through the query. Finally, since we have to have the index names in variables anyway (for initialization code), it naturally makes sense to re-use those variables.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: Since I am overloading an existing method of this name, I am keeping it as it for now.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TermQueryExplainIT.java
abbashus: Keeping it inline for now.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: Keeping the logic same for know to match [existing behavior](https://github.com/opendistro-for-elasticsearch/sql/pull/98/files#diff-128e1606ce985235d9324d79ae94cddcR101) , this might break with inner fields in object type in nested docs. Will verify that special case and handle in a separate PR.
Created issue #142

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/SQLClause.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
galkk: +1

abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -213,15 +214,25 @@ public boolean isValidIdentifierForTerm(SQLIdentifierExpr expr) {
...
-                || expr.getParent() instanceof MySqlSelectGroupByExpr
-            );
+            && (isValidIdentifier(expr) || checkIfNestedIdentifier(expr));
+    }
+
+    private boolean checkIfNestedIdentifier(SQLIdentifierExpr expr) {
+        return
+            expr.getParent() instanceof SQLMethodInvokeExpr
+            && ((SQLMethodInvokeExpr) expr.getParent()).getMethodName().equals("nested")
+            && isValidIdentifier((SQLMethodInvokeExpr) expr.getParent());
```
galkk: I assume that SQLMethodInvokeExpr is child of SQLExpr. Do we need the cast here?

abbashus: The last revision has cast removed, don't know why it is showing stale code.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
arsen-es: I think now you changed the semantics of the method, at least how I understand it it checks that the sub/child expression is a valid identifier, because when you passed in X, you checked for X.parent in the method, but now you pass in X, and check for X. I would prefer to keep the older code, with the extra variable assignment, or now you need to change the method name to properly reflect the semantics.

abbashus: Reverted with extra variable assignment to keep semantics intact.

PR #141, opened 2019-08-06
Issue 101

davidcui-amzn:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
@@ -86,6 +86,7 @@ public class DefaultQueryAction extends QueryAction {
         this.request = new SearchRequestBuilder(client, SearchAction.INSTANCE);
         setIndicesAndTypes();
 
+        // Potential bug source
         setFields(select.getFields());
 
         setWhere(select.getWhere());
@@ -134,7 +135,7 @@ public class DefaultQueryAction extends QueryAction {
      */
     public void setFields(List<Field> fields) throws SqlParseException {
 
-        if (!select.getFields().isEmpty()) {
+        if (!select.getFields().isEmpty() && !select.isSelectAll()) {
             ArrayList<String> includeFields = new ArrayList<>();
             ArrayList<String> excludeFields = new ArrayList<>();
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
@@ -86,7 +86,6 @@ public class DefaultQueryAction extends QueryAction {
         this.request = new SearchRequestBuilder(client, SearchAction.INSTANCE);
         setIndicesAndTypes();
 
-        // Potential bug source
         setFields(select.getFields());
 
         setWhere(select.getWhere());


abbashus:
Hi @davidcui-amzn , can you edit the description to reflect what is this change about and also add the corresponding issue (is any) and testing details.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
```diff
@@ -134,7 +134,7 @@ private void setIndicesAndTypes() {
      */
     public void setFields(List<Field> fields) throws SqlParseException {
 
-        if (!select.getFields().isEmpty()) {
+        if (!select.getFields().isEmpty() && !select.isSelectAll()) {
```
arsen-es: Please add tests to cover this case.

arsen-es: Based on Andy's recommendation, also add integ tests that have GROUP BY and ORDER BY to reduce the chance of breaking something else.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,49 @@ public void multipleFromTest() throws IOException {
         Assert.assertEquals(14, getTotalHits(response));
     }
 
+    @Test
+    public void selectAllAndSpecificFieldTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
+        Set<String> expectedSource = new HashSet<>(Arrays.asList(arr));
+        JSONObject response = executeQuery(String.format(Locale.ROOT,
+                "SELECT *, age FROM %s/account", TestsConstants.TEST_INDEX_BANK));
```
arsen-es: np: maybe add LIMIT with a small number like 5, not to iterate through all the documents for this test.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,49 @@ public void multipleFromTest() throws IOException {
...
+        Assert.assertTrue(response.has("hits"));
+        JSONArray hits = getHits(response);
+        for (int i = 0; i < hits.length(); ++i) {
+            JSONObject hit = hits.getJSONObject(i);
+            Assert.assertEquals(expectedSource, getSource(hit).keySet());
```
arsen-es: Do you want to also verify the absence of a duplicate field `age` in the result set? Since you're using sets, you only make sure there's at least one.

davidcui-amzn: The test case accounts for that- if there is a duplicate field, the `assertEquals()` statement will fail against `arr `

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
davidcui-amzn: Good idea- will add now

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,49 @@ public void multipleFromTest() throws IOException {
...
+
+    @Test
+    public void selectAllAndSpecificFieldwithGroupByTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
+        Set<String> expectedSource = new HashSet<>(Arrays.asList(arr));
+        JSONObject response = executeQuery(String.format(Locale.ROOT,
+                "SELECT *, age FROM %s/account GROUP BY age", TestsConstants.TEST_INDEX_BANK));
+        Assert.assertTrue(response.has("hits"));
+        JSONArray hits = getHits(response);
+        for (int i = 0; i < hits.length(); ++i) {
```
arsen-es: This loop does not execute since your hits is empty in case of aggregation query.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,49 @@ public void multipleFromTest() throws IOException {
...
+    }
+
+    @Test
+    public void selectAllAndSpecificFieldwithOrderByTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
+        Set<String> expectedSource = new HashSet<>(Arrays.asList(arr));
+        JSONObject response = executeQuery(String.format(Locale.ROOT,
+                "SELECT *, age FROM %s/account ORDER BY age", TestsConstants.TEST_INDEX_BANK));
+        Assert.assertTrue(response.has("hits"));
+        JSONArray hits = getHits(response);
```
arsen-es: verify that hits is not empty, or has a certain size, before looping.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
arsen-es: The code compares two sets, which never have duplicates.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
dai-chen: If possible, please check JDBC format too.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,49 @@ public void multipleFromTest() throws IOException {
         Assert.assertEquals(14, getTotalHits(response));
     }
 
+    @Test
+    public void selectAllAndSpecificFieldTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
```
dai-chen: Another option is Guava's `Sets.newHashSet("account_number", ...)`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
dai-chen: I think we should give alias to column `age` in the SELECT.

dai-chen:
Btw, please put issue link for the "Issue #, if available:" so your PR would be referenced by the issue automatically.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
galkk: +1.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,49 @@ public void multipleFromTest() throws IOException {
...
+    }
+
+    @Test
+    public void selectAllAndSpecificFieldwithGroupByTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
+        Set<String> expectedSource = new HashSet<>(Arrays.asList(arr));
+        JSONObject response = executeQuery(String.format(Locale.ROOT,
+                "SELECT *, age FROM %s/account GROUP BY age LIMIT 5", TestsConstants.TEST_INDEX_BANK));
+        Assert.assertTrue(response.has("hits"));
+        JSONArray hits = getHits(response);
```
galkk: remove copypaste from tests (that hits check seem to be repeated between all tests introduced).

Please look into SubqueryIT.java for better approach of validation of the query results.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,59 @@ public void multipleFromTest() throws IOException {
         Assert.assertEquals(14, getTotalHits(response));
     }
 
+    @Test
+    public void selectAllAndSpecificFieldTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
+        String fieldName = "age";
+        Set<String> expectedSource = new HashSet<>(Arrays.asList(arr));
+        JSONObject response = executeQuery(String.format(Locale.ROOT,
+                "SELECT * as a, %s as t " +
```
arsen-es: What is `* as a` supposed to do? I am unfamiliar with this syntax.

davidcui-amzn: To give an alias to all the field names involved, trying to replicate the same conditions as the original issue

davidcui-amzn: Update- you're right, the syntax is invalid as I tested against the incomplete _opendistro/_sql plugin. Will make change for next commit

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +99,59 @@ public void multipleFromTest() throws IOException {
         Assert.assertEquals(14, getTotalHits(response));
     }
 
+    @Test
+    public void selectAllAndSpecificFieldTest() throws IOException {
+        String[] arr = new String[] {"account_number", "firstname", "address", "birthdate", "gender", "city", "lastname", "balance", "employer", "state", "age", "email", "male"};
+        String fieldName = "age";
+        Set<String> expectedSource = new HashSet<>(Arrays.asList(arr));
+        JSONObject response = executeQuery(String.format(Locale.ROOT,
+                "SELECT *, %s as t " +
```
arsen-es: So what is the expected behavior : should return all the fields and "t" or just all the fields?

davidcui-amzn: Should return all the fields which naturally includes t

arsen-es: I think this should return all the fields with original names, plus the extra "age" field but named "t". Please confirm with a Sql or MySql implementation. We still have many issues in our plugin, so please do not completely rely on it when determining the correct behavior.

arsen-es: Alright, so I tested with different online editors (claiming to be either Sql or MySql compliant), and in all cases there is a column with the alias. However, in some cases the original column was there, in some cases not, e.g.

```sql
create table Test(id integer, title varchar(100));
insert into Test(id, title) values(1, "Hello");
insert into Test(id, title) values(2, "World");
insert into Test(id, title) values(3, "blah");

select *, title as word from Test;
```

In some cases this results in:
```
id  |  title |  word
...     ...     ...
```

and in others
```
id  |  word
...    ...
```

I am more inclined to support the first option -> all fields + alias, to be consistent with queries like `SELECT someField, someField AS otherField FROM tableX` which always returns 2 columns.

davidcui-amzn: Verified and you are right- will be more thorough testing in the future. Currently opendistro does not rename the column correctly, as it still shows up as `address`. Should I open up another issue for the column alias?

davidcui-amzn: Had lots of trouble fixing column alias bug- Chen observed that aliases are supported with JDBC format, including the original issue. Left in original change that will return all fields in query with * and selected field

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +105,20 @@ public void multipleFromTest() throws IOException {
...
 
+    @Test
+    public void selectAllWithFieldThrowsException() throws IOException {
+        String fieldName = "age";
+        exception.expect(ResponseException.class);
+        exception.expectMessage("\"error\": {\n" +
+                "    \"reason\": \"Invalid SQL query\",\n" +
+                "    \"details\": \"Other expressions may not be present in the select list when '*' is used\",\n" +
+                "    \"type\": \"SqlParseException\"");
+        executeQuery(String.format(Locale.ROOT,
```
dai-chen: Please use our own `StringUtils` instead.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -91,6 +91,13 @@ else if(isJoin(sqlExpr,sql)){
                 else {
                     sqlExpr.accept(new TermFieldRewriter(client));
                     Select select = new SqlParser().parseSelect(sqlExpr);
+                    if (select.isSelectAll() && !select.getFields().isEmpty()) {
```
dai-chen: How about we move this new added logic to `SqlParser.findSelect()` to fail fast?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -91,6 +91,13 @@ else if(isJoin(sqlExpr,sql)){
...
+                        for (int i = 0; i < select.getFields().size(); ++i) {
+                            if (select.getFields().get(i) != null && !select.getFields().get(i).isNested()) {
+                                throw new SqlParseException("Other expressions may not be present in the select list when '*' is used");
```
galkk: **Nested** expressions may not be present, and you could add the expression itself (field

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -99,6 +105,20 @@ public void multipleFromTest() throws IOException {
...
+                "SELECT *, %s as t " + "" +
+                        "FROM %s/account ORDER BY %s LIMIT 5",
```
galkk: Do not use String.format for queries, inline everything. It will improve query readability and allow users to copy-paste it to kibana

PR #146, opened 2019-08-09
Added support for PERCENTILES in JDBC driver

### LINKED ISSUE #26: PERCENTILES aggregation does not work with JDBC format.
PERCENTILES aggregation works with ES DSL output but it does not work with jdbc output format.

Please add support for PERCENTILES aggregation. 

Example query 

```
curl "localhost:9200/_opendistro/_sql?format=jdbc" -H 'Content-Type: application/json' -d '{
  "query": "SELECT PERCENTILES(column_one,95.0) FROM my_index"
}'
```

galkk:
Adds PERCENTILE output to JDBC 

```json
GET /_opendistro/_sql?format=jdbc
{
  "query": "SELECT percentiles(age, 25.0, 50.0, 75.0, 99.9) age_percentiles FROM peoples"
}

{
  "schema": [{
    "name": "age_percentiles",
    "type": "double"
  }],
  "total": 1,
  "datarows": [[{
    "75.0": 36.5,
    "99.9": 39,
    "25.0": 31.5,
    "50.0": 33.5
  }]],
  "size": 1,
  "status": 200
}
```

Added new unchecked exception that enables us to say that we clearly know that feature is not implemented (differentiating it from not supported)

Fixes #26

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
@@ -1,0 +1,19 @@
+package com.amazon.opendistroforelasticsearch.sql.exception;
+
+/**
+ * Intended for cases when we knowingly omitted some case, letting users know that we didn't implemented feature, but
+ * it may be implemented in future.
+ */
+public class SqlFeatureNotImplementedException extends RuntimeException {
+    static final long serialVersionUID = 1;
+
+    public SqlFeatureNotImplementedException() { }
+
+    public SqlFeatureNotImplementedException(String message) {
+        super(message);
+    }
+
+    public SqlFeatureNotImplementedException(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
@@ -7,8 +7,6 @@ package com.amazon.opendistroforelasticsearch.sql.exception;
 public class SqlFeatureNotImplementedException extends RuntimeException {
     private static final long serialVersionUID = 1;
 
-    public SqlFeatureNotImplementedException() { super();}
-
     public SqlFeatureNotImplementedException(String message) {
         super(message);
     }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
@@ -1,3 +1,17 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
 package com.amazon.opendistroforelasticsearch.sql.exception;
 
 /**
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
@@ -5,9 +5,9 @@ package com.amazon.opendistroforelasticsearch.sql.exception;
  * it may be implemented in future.
  */
 public class SqlFeatureNotImplementedException extends RuntimeException {
-    static final long serialVersionUID = 1;
+    private static final long serialVersionUID = 1;
 
-    public SqlFeatureNotImplementedException() { }
+    public SqlFeatureNotImplementedException() { super();}
 
     public SqlFeatureNotImplementedException(String message) {
         super(message);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
@@ -23,6 +23,7 @@ import com.amazon.opendistroforelasticsearch.sql.domain.ScriptMethodField;
 import com.amazon.opendistroforelasticsearch.sql.domain.Select;
 import com.amazon.opendistroforelasticsearch.sql.domain.TableOnJoinSelect;
 import com.amazon.opendistroforelasticsearch.sql.esdomain.mapping.FieldMapping;
+import com.amazon.opendistroforelasticsearch.sql.exception.SqlFeatureNotImplementedException;
 import com.amazon.opendistroforelasticsearch.sql.utils.SQLFunctions;
 import org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsRequest;
 import org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsResponse;
@@ -35,6 +36,8 @@ import org.elasticsearch.search.aggregations.Aggregation;
 import org.elasticsearch.search.aggregations.Aggregations;
 import org.elasticsearch.search.aggregations.bucket.terms.Terms;
 import org.elasticsearch.search.aggregations.metrics.NumericMetricsAggregation;
+import org.elasticsearch.search.aggregations.metrics.Percentile;
+import org.elasticsearch.search.aggregations.metrics.Percentiles;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -44,6 +47,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import static java.util.stream.Collectors.toSet;
 import static org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsResponse.FieldMappingMetaData;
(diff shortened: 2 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
```diff
@@ -0,0 +1,19 @@
+package com.amazon.opendistroforelasticsearch.sql.exception;
+
+/**
+ * Intended for cases when we knowingly omitted some case, letting users know that we didn't implemented feature, but
+ * it may be implemented in future.
+ */
+public class SqlFeatureNotImplementedException extends RuntimeException {
+    static final long serialVersionUID = 1;
+
+    public SqlFeatureNotImplementedException() { }
```
arsen-es: I think this should call the base constructor with some default message.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -555,8 +561,21 @@ private boolean hasTermAggregations(List<Aggregation> aggs) {
...
-            data.put(numericAggs.getName(), numericAggs.value());
+            if (aggregation instanceof NumericMetricsAggregation.SingleValue){
+                NumericMetricsAggregation.SingleValue singleValueAggregation = (NumericMetricsAggregation.SingleValue) aggregation;
+                data.put(singleValueAggregation.getName(), !Double.isInfinite(singleValueAggregation.value()) ? singleValueAggregation.getValueAsString() : "null");
+            }
+            else if (aggregation instanceof Percentiles) {
+                Percentiles percentiles = (Percentiles) aggregation;
+
+                data.put(percentiles.getName(), StreamSupport
+                        .stream(percentiles.spliterator(), false)
```
dai-chen: I'm not familiar with `Spliterator`. I'm curious what's the benefit we get from it? Anything special about `Percentile`?

And I notice `Percentile` doesn't override it and here is what I found from the Javadoc in `Iterable` interface. Any concern?

```
     * The default implementation should usually be overridden.  The
     * spliterator returned by the default implementation has poor splitting
     * capabilities, is unsized, and does not report any spliterator
     * characteristics. Implementing classes can nearly always provide a
     * better implementation.
     ...
    default Spliterator<T> spliterator() {
        return Spliterators.spliteratorUnknownSize(iterator(), 0);
    }
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
```diff
@@ -24,6 +24,22 @@
     @Override
     protected void init() throws Exception {
         loadIndex(Index.ONLINE);
+        loadIndex(Index.PEOPLE);
+    }
+
+    public void testPercentilesQuery() {
+        JSONObject response = executeJdbcRequest(
+                "SELECT percentiles(age, 25.0, 50.0, 75.0, 99.9) age_percentiles" +
```
arsen-es: IIRC, the non-sql percentiles aggregation query honors the order specified in the query (or it always sorts in ascending). The example response you put in the PR description does none of those. I know this is a useful fix even without that, but ideally the sql plugin should be consistent with DSL if possible.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
```diff
@@ -24,6 +24,22 @@
...
+                        " FROM elasticsearch-sql_test_index_people");
```
abbashus: np: indentation w.r.t SELECT. Similarly for `testGroupByInQuery`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
galkk: You can't stream over ordinary Iterable, only foreach. The streamsupport enables you to do that

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -555,8 +561,21 @@ private boolean hasTermAggregations(List<Aggregation> aggs) {
...
+                        .collect(Collectors.toMap(Percentile::getPercent, Percentile::getValue)));
+            }
+            else {
+                throw new SqlFeatureNotImplementedException("Aggregation type " + aggregation.getType()
```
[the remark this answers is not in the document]
galkk: Done

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
galkk: Unfortunately our serialization library (org.json) doesn't work well in that matter, and doesn't respect the order at all. Even if you will put percentiles into TreeMap (I've tried that in the beginning), it will still serialize it in random order, because underlying collection is hashmap. We need to move to either Jackson or GSON for that to be fixed

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
galkk: Done

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
galkk: It seems like typical pattern is just to call super() there, fixed

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
```diff
@@ -0,0 +1,19 @@
+package com.amazon.opendistroforelasticsearch.sql.exception;
+
+/**
+ * Intended for cases when we knowingly omitted some case, letting users know that we didn't implemented feature, but
+ * it may be implemented in future.
+ */
+public class SqlFeatureNotImplementedException extends RuntimeException {
+    static final long serialVersionUID = 1;
```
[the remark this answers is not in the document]
galkk: No, fixed

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
dai-chen: > You can't stream over ordinary Iterable, only foreach. The streamsupport enables you to do that

I see. I was confused by the name `spliterator` and thought there are some parallelism benefits. Guava seems have  `Streams.stream(Iterable)` which looks much clearer to me. Unfortunately our version is too low.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
```diff
@@ -0,0 +1,19 @@
+package com.amazon.opendistroforelasticsearch.sql.exception;
+
+/**
+ * Intended for cases when we knowingly omitted some case, letting users know that we didn't implemented feature, but
+ * it may be implemented in future.
+ */
+public class SqlFeatureNotImplementedException extends RuntimeException {
+    private static final long serialVersionUID = 1;
+
+    public SqlFeatureNotImplementedException() { super();}
```
arsen-es: I still think we need a default message giving some context about what went wrong, otherwise we will have to enforce always catching this and converting to a more user-friendly response/exception. I would even prefer not to provide the default constructor if you feel there's not a single good message.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
arsen-es: Could you please double check if it's the same for plain json response (without 'format' parameter in the request url)? I suspect this was working as expected for non-jdbc queries. I just wonder if we use two separate serialization libraries for JDBC vs non-JDBC cases.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
galkk: 2nd parameter (false) disables parallelism :)

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
galkk: It's just typical pattern, look at the implementation of 

```
 */
public
class IllegalArgumentException extends RuntimeException {
    /**
     * Constructs an <code>IllegalArgumentException</code> with no
     * detail message.
     */
    public IllegalArgumentException() {
        super();
    }
```
for example

arsen-es: `IllegalArgumentException` is such a common type of exception, and since it's used in different layers, so the type without any extra info is enough to do things like

```
... } catch (IllegalArgumentException e) {
...
} catch (OtherException e) {
...
}
```
What I am saying, at least from the description of the class, your exception is user-facing, so it doesn't make much sense having it without any messaging unless someone will always do
```
catch (SqlFeatureNotImplementedException e) {
   return buildResponse(e); // made up function for illustration
}
```

Finally, maybe it's better [not to add it if we're not using it](https://www.martinfowler.com/bliki/Yagni.html).

Up to you if you want to keep it as is, just seems a bit off to me.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
galkk: Default works because we don't do any post-processing or deserialization there - it's just elastic output, that makes things right.

arsen-es: I see, thanks for the clarification.

src/main/java/com/amazon/opendistroforelasticsearch/sql/exception/SqlFeatureNotImplementedException.java
galkk: Fixed

Issue #152, opened 2019-08-14
Lack of support for nested documents with more than one level of nesting


zhongnansu:
For both the one level nested and two level nested query, the `_explain` API  returns the same DSL.

- Mapping
```
{
  "employees_nested_2" : {
    "mappings" : {
      "properties" : {
        "comments" : {
          "type" : "nested",
          "properties" : {
            "date" : {
              "type" : "date"
            },
            "likes" : {
              "type" : "long"
            },
            "message" : {
              "type" : "text",
              "fields" : {
                "keyword" : {
                  "type" : "keyword",
                  "ignore_above" : 256
                }
              }
            }
          }
        },
        "id" : {
          "type" : "long"
        },
        "name" : {
          "type" : "text",
          "fields" : {
            "keyword" : {
              "type" : "keyword",
              "ignore_above" : 256
            }
          }
        },
        "projects" : {
          "type" : "nested",
          "properties" : {
            "address" : {
              "type" : "nested",
              "properties" : {
                "city" : {
                  "type" : "text",
                  "fields" : {
                    "keyword" : {
                      "type" : "keyword",
                      "ignore_above" : 256
                    }
                  }
                },
                "state" : {
                  "type" : "text",
                  "fields" : {
                    "keyword" : {
                      "type" : "keyword",
                      "ignore_above" : 256
                    }
                  }
                }
              }
            },
            "name" : {
              "type" : "text",
              "fields" : {
                "keyword" : {
                  "type" : "keyword",
                  "ignore_above" : 256
                }
              }
            },
            "started_year" : {
              "type" : "long"
            }
          }
        },
        "title" : {
          "type" : "text",
          "fields" : {
            "keyword" : {
              "type" : "keyword",
              "ignore_above" : 256
            }
          }
        }
      }
    }
  }
}
```

- SQL query
```

POST _opendistro/_sql/?
{
"query" : "SELECT * FROM employees_nested_2 e, e.projects p"
}


POST _opendistro/_sql/?
{
"query" : "SELECT * FROM employees_nested_2 e, e.projects p, p.address as a"
}
```

- DSL
```
{
  "from": 0,
  "size": 200,
  "query": {
    "bool": {
      "filter": [
        {
          "bool": {
            "must": [
              {
                "nested": {
                  "query": {
                    "match_all": {
                      "boost": 1
                    }
                  },
                  "path": "projects",
                  "ignore_unmapped": false,
                  "score_mode": "none",
[...]

PR #153, opened 2019-08-15
Support certain types of query with deeply nested field

zhongnansu:
*Issue #, if available:*
#152
Only supporting plain Elasticsearch output for now.

*Description of changes:*
- Fixed `aliasFullPath` mapping in `Scope` to store more than one level of nesting path and its alias.
- Fixed the logic of creating nested query in DSL with deeply nested fields. The goal is, for nested field, its corresponding DSL also need to be nested.

*Support query* (See tests for more details)
- Select All: 
    - `SELECT * FROM employees e, e.projects p, p.address a`
- Deeper select all:
    - `SELECT * FROM employees e, e.projects p, p.address a, a.country c`
- Deep select all from different nesting levels:
    - `SELECT * FROM employees e, e.projects p, p.address a, e.comments c, c.message m`
- Select multiple nested fields from different nesting level (regular, 1st, 2nd):
    - `SELECT title, p.started_year, a.city, FROM employees e, e.projects p, p.address a`
- Deep select nested fields: 
    - `SELECT p.name, a.city, p.year, a.state FROM employees e, e.projects p, p.address a`

*Testing*
- Passed gradle build tests
- Added unit tests and integration tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
@@ -108,6 +108,12 @@ class From extends SQLClause<SQLTableSource> {
         Identifier table = new Identifier((SQLIdentifierExpr) ((SQLExprTableSource) expr).getExpr());
         if (table.path().equals(scope.getParentAlias())) {
             scope.addAliasFullPath(emptyIfNull(expr.getAlias()), table.name());
+        } else {
+            String fullPath = scope.getFullPath(table.path());
+
+            if (!fullPath.isEmpty()) {
+                scope.addAliasFullPath(emptyIfNull(expr.getAlias()),fullPath + "." + table.name());
+            }
         }
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
@@ -106,14 +106,12 @@ class From extends SQLClause<SQLTableSource> {
         }
 
         Identifier table = new Identifier((SQLIdentifierExpr) ((SQLExprTableSource) expr).getExpr());
-        if (table.path().equals(scope.getParentAlias())) {
-            scope.addAliasFullPath(emptyIfNull(expr.getAlias()), table.name());
-        } else {
-            String fullPath = scope.getFullPath(table.path());
 
-            if (!fullPath.isEmpty()) {
-                scope.addAliasFullPath(emptyIfNull(expr.getAlias()),fullPath + "." + table.name());
-            }
+        // the top level parent table has an empty path. Don't add it to {alias -> path} mapping
+        if (!table.path().isEmpty()) {
+            String fullPath = scope.getFullPath(table.path());
+            String prefix = fullPath.isEmpty() ? "" : fullPath + ".";
+            scope.addAliasFullPath(emptyIfNull(expr.getAlias()), prefix + table.name());
         }
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
@@ -60,7 +60,7 @@ class Identifier extends SQLClause<SQLIdentifierExpr> {
     }
 
     private int separatorIndex() {
-        return expr.getName().indexOf(SEPARATOR);
+        return expr.getName().lastIndexOf(SEPARATOR);
     }
 
     /**
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
@@ -52,14 +52,14 @@ class Identifier extends SQLClause<SQLIdentifierExpr> {
     }
 
     String path() {
-        return separatorIndex() == -1 ? "" : expr.getName().substring(0, separatorIndex());
+        return lastSeparatorIndex() == -1 ? "" : expr.getName().substring(0, lastSeparatorIndex());
     }
 
     String name() {
-        return expr.getName().substring(separatorIndex() + 1);
+        return expr.getName().substring(lastSeparatorIndex() + 1);
     }
 
-    private int separatorIndex() {
+    private int lastSeparatorIndex() {
         return expr.getName().lastIndexOf(SEPARATOR);
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
@@ -43,13 +43,93 @@ import static org.elasticsearch.index.query.QueryBuilders.nestedQuery;
 
 /**
  * Nested field projection class to make ES return matched rows in nested field.
+ *
+ * Example:
+ * SQL: SELECT * FROM employees e, e.projects p, p.address a, e.comments c, c.message m
+ * Rewrite: SELECT *, nested(projects.*), nested(projects.address.*), nested(comments.*),
+ *          nested(comments.message.*) FROM employee
+ *
+ *
+ * State transition table (state here means lastPath + DSL)
+ *  ______________________________________________________________________________________
+ *  |    lastPath       |               DSL                                              |
+ *  |------------------------------------------------------------------------------------|
+ *  |       ""          |       "must": [                                                |
+ *  |                   |         "nested": {                                            |
+ *  |                   |            "query":{},                                         |
+ *  |                   |            "path": "projects.address"                          |
+ *  |                   |          }                                                     |
+ *  |                   |       ]                                                        |
+ *  |------------------------------------------------------------------------------------|
+ *  | projects.address  |      "must": [                                                 |
+ *  |                   |        {                                                       |
+ *  |                   |          "nested": {                                           |
+ *  |                   |            "query": {                                          |
+ *  |                   |              "nested": {                                       |
+ *  |                   |                "query":{},                                     |
+ *  |                   |                "path": "projects.address"                      |
+ *  |                   |               }                                                |
+ *  |                   |             },                                                 |
+ *  |                   |          "path": "projects"                                    |
+ *  |                   |          }                                                     |
+ *  |                   |        }                                                       |
+ *  |                   |      ]                                                         |
+ *  |------------------------------------------------------------------------------------|
+ *  | projects          |      "must": [                                                 |
+ *  |                   |        {                                                       |
+ *  |                   |          "nested": {                                           |
+ *  |                   |            "query": {                                          |
+ *  |                   |              "nested": {                                       |
+ *  |                   |                "query":{},                                     |
+ *  |                   |                "path": "projects.address"                      |
+ *  |                   |               }                                                |
+ *  |                   |             },                                                 |
+ *  |                   |          "path": "projects"                                    |
+ *  |                   |          }                                                     |
+ *  |                   |        },                                                      |
+ *  |                   |        {                                                       |
+ *  |                   |          "nested": {                                           |
+ *  |                   |            "query":{},                                         |
+ *  |                   |            "path": "comments.message"                          |
+ *  |                   |          }                                                     |
+ *  |                   |        }                                                       |
+ *  |                   |      ]                                                         |
+ *  |------------------------------------------------------------------------------------|
+ *  | comments.message  |      "must": [                                                 |
+ *  |                   |        {                                                       |
+ *  |                   |          "nested": {                                           |
+ *  |                   |            "query": {                                          |
+ *  |                   |              "nested": {                                       |
+ *  |                   |                "query":{},                                     |
+ *  |                   |                "path": "projects.address"                      |
+ *  |                   |               }                                                |
+ *  |                   |             },                                                 |
+ *  |                   |          "path": "projects"                                    |
+ *  |                   |          }                                                     |
+ *  |                   |        },                                                      |
+ *  |                   |        {                                                       |
+ *  |                   |          "nested": {                                           |
+ *  |                   |            "query": {                                          |
+ *  |                   |              "nested": {                                       |
+ *  |                   |                "query":{},                                     |
+ *  |                   |                "path": "comments.message"                      |
+ *  |                   |               }                                                |
+ *  |                   |             },                                                 |
+ *  |                   |          "path": "comments"                                    |
+ *  |                   |          }                                                     |
+ *  |                   |       }                                                        |
+ *  |                   |     ]                                                          |
+ *  |------------------------------------------------------------------------------------|
+ *
+ *  More Details:
+ *    1) Keep adding nesting layers if the path are nested.
+ *    2) if the path are not nested, Stop adding nesting layers. Instead create new nested domain
  */
 public class NestedFieldProjection {
 
     private final SearchRequestBuilder request;
     private static final String SEPARATOR = ".";
-    private String lastFieldInPath = "";
-    private List<String> lastFieldNames = new ArrayList<>();
+    private String lastPath = "";
 
     public NestedFieldProjection(SearchRequestBuilder request) {
         this.request = request;
@@ -65,16 +145,14 @@ public class NestedFieldProjection {
             // Extract nested query from WHERE clause
             List<NestedQueryBuilder> nestedQueries = extractNestedQueries(query());
             groupFieldNamesByPath(fields).forEach(
-                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path, fieldNames))
+                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path))
             );
         }
     }
 
-    /** Reset values of lastFieldInPath and lastFieldNames to prepare for next check */
-    private void setLastValues(String path, List<String> fieldNames) {
-        int pos = path.lastIndexOf(SEPARATOR);
-        lastFieldInPath = path.substring(pos + 1);
-        lastFieldNames = fieldNames;
+    /** Reset values of lastPath to prepare for next check */
+    private void setLastPath(String path) {
+        lastPath = path;
     }
 
     /** Check via traditional for loop first to avoid lambda performance impact on all queries even though those without nested field */
@@ -96,6 +174,10 @@ public class NestedFieldProjection {
         }
     }
 
+    /** Use TreeMap to generate mapping {path -> list of field domain object} in reverse alphabetical order.
+     *  This order is used to generate nested query from the deepest level to the top level.
+     *  e.g. path 'project.address.city' should be nested in 'project.address', should be nested in 'project'
+     */
     private Map<String, List<String>> groupFieldNamesByPath(List<Field> fields) {
         return fields.stream().
                       filter(Field::isNested).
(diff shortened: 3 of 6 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
```diff
@@ -108,6 +108,12 @@ private void addIfNestedField(Scope scope) {
         Identifier table = new Identifier((SQLIdentifierExpr) ((SQLExprTableSource) expr).getExpr());
         if (table.path().equals(scope.getParentAlias())) {
             scope.addAliasFullPath(emptyIfNull(expr.getAlias()), table.name());
+        } else {
+            String fullPath = scope.getFullPath(table.path());
+
+            if (!fullPath.isEmpty()) {
+                scope.addAliasFullPath(emptyIfNull(expr.getAlias()),fullPath + "." + table.name());
```
arsen-es: missing space after comma

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
```diff
@@ -60,7 +60,7 @@ String name() {
     }
 
     private int separatorIndex() {
```
arsen-es: np: maybe rename to `lastSeparatorIndex` ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -57,14 +62,21 @@ public NestedFieldProjection(SearchRequestBuilder request) {
...
-                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path))
+                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path, fieldNames))
             );
         }
     }
 
+    /** Reset values of lastFieldInPath and lastFieldNames to prepare for next check */
+    private void setLastValues(String path, List<String> fieldNames) {
+        int pos = path.lastIndexOf(SEPARATOR);
+        lastFieldInPath = path.substring(pos + 1);
```
arsen-es: Any chance the path doesn't include a separator?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +135,64 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
     }
 
     /** Create a nested query with match all filter to place inner hits */
-    private Supplier<NestedQueryBuilder> createEmptyNestedQuery(String path) {
+    private Supplier<NestedQueryBuilder> createNestedQuery(String path, List<String> fieldNames) {
         return () -> {
-            NestedQueryBuilder nestedQuery = nestedQuery(path, matchAllQuery(), ScoreMode.None);
-            ((BoolQueryBuilder) query().filter().get(0)).must(nestedQuery);
-            return nestedQuery;
+            NestedQueryBuilder newNestedQuery;
```
arsen-es: `final`

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +135,64 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
+
+            if (mustList.isEmpty() || !hasSameParent(path, mustList.get(lastIndex))) {
+                // create empty nested query
+                newNestedQuery = nestedQuery(path, matchAllQuery(), ScoreMode.None);
+            } else {
+                // create a nested query containing another nested query
+                NestedQueryBuilder oldNestedQuery = (NestedQueryBuilder) mustList.get(lastIndex);
+                newNestedQuery = nestedQuery(path, oldNestedQuery, ScoreMode.None);
+
+                // remove outdated nested query
```
arsen-es: np: is it outdated or already processed?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/NestedFieldQueryIT.java
```diff
@@ -170,6 +172,107 @@ public void singleCondition() throws IOException {
...
+                    )
+                )
+            )
+        );
+    }
+
+    @Test
+    public void deepSelectAll() throws IOException {
+        assertThat(
+            execute("SELECT * FROM elasticsearch-sql_test_index_employee_nested e, e.projects p, p.address a"),
```
arsen-es: np: I still think using the static constants for index names is better.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/NestedFieldProjectionTest.java
```diff
@@ -239,10 +240,290 @@ public void multipleDifferentNestedFields() {
...
+                                )
+                        )
+                    )
+                )
+            )
+        );
+    }
+
+    @Test
+    @Ignore("Deep query with WHERE condition is not fully supported")
```
arsen-es: Thanks for adding the tests for the unsupported cases.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -57,14 +62,21 @@ public NestedFieldProjection(SearchRequestBuilder request) {
     public void project(List<Field> fields) {
         if (isAnyNestedField(fields)) {
             initBoolQueryFilterIfNull();
-
+            // Extract nested query from WHERE clause
             List<NestedQueryBuilder> nestedQueries = extractNestedQueries(query());
             groupFieldNamesByPath(fields).forEach(
-                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path))
+                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path, fieldNames))
```
galkk: Could you add comment, explaining what's going on here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -88,7 +100,7 @@ private void initBoolQueryFilterIfNull() {
         return fields.stream().
                       filter(Field::isNested).
                       filter(not(Field::isReverseNested)).
-                      collect(groupingBy(Field::getNestedPath, mapping(Field::getName, toList())));
+                      collect(groupingBy(Field::getNestedPath, () -> new TreeMap<>(Collections.reverseOrder()), mapping(Field::getName, toList())));
```
galkk: Why do we need that treemap (and an order, that it implies)?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +135,64 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
         return nestedQueries.stream().
-                             filter(query -> isSamePath(path, query)).
-                             findAny().
-                             orElseGet(createEmptyNestedQuery(path));
+                filter(query -> isSamePath(path, query, fieldNames)).
+                findAny().
+                orElseGet(createNestedQuery(path, fieldNames));
+    }
+
+    /** Check against NestedQuery of WHERE condition */
```
galkk: Please add better explanation, what exactly is checked

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +135,64 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
+    private boolean isSamePath(String path, NestedQueryBuilder query, List<String> fieldNames) {
+        boolean isSamePath = nestedQuery(path, query.query(), query.scoreMode()).equals(query);
+        if (isSamePath) {
```
galkk: This is misleading - you function mutates the state, that is not clear for consumer.
Instead of this, remove condition here and rewrite function above as

```
 return nestedQueries.stream()
                             .filter(query -> isSamePath(path, query, fieldNames)).
                             .findAny()
                             .map(x->setLastValues(path,fieldNames))
                             orElseGet(createEmptyNestedQuery(path));	                orElseGet(createNestedQuery(path, fieldNames));
```

This findAny is also quite suspicious

galkk: or use peek() if you want to modify the collection

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +135,64 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
 
-    private boolean isSamePath(String path, NestedQueryBuilder query) {
-        return nestedQuery(path, query.query(), query.scoreMode()).equals(query);
+    /**
+     * Since groupFieldNamesByPath() uses TreeMap to maintain reverse alphabetical order. Here we maintain previous
+     * path and fieldName value to compare with current nested query.
+     * */
+    private boolean hasSameParent(String path, QueryBuilder query) {
+        // query could also be termQueryBuilder
+        if (!lastFieldInPath.isEmpty() && (query instanceof NestedQueryBuilder)) {
```
galkk: I highly recommend to internalize the https://blog.codinghorror.com/flattening-arrow-code/

and embrace ealier exits

if (lastFieldInPath.isEmpty() || ! (query instanceOf NestedQueryBuilder)) {
    return false;
}

main logic

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/NestedFieldQueryIT.java
galkk: I think that the most of the problem is that our generated index names are too long.

static constants - why just for index names, why not for fields? statements? Current way is cleanest possible, and doesn't hide anything from reader

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/NestedFieldQueryIT.java
```diff
@@ -170,6 +172,107 @@ public void singleCondition() throws IOException {
         );
     }
 
+    @Test
+    public void regularAndDeepNestedFieldInSelect() throws IOException {
+        assertThat(
+            execute("select title, a.city " +
+                    "from elasticsearch-sql_test_index_employee_nested e, e.projects p, p.address a"),
+            hits(
```
galkk: I like it so much, this form should be mandatory for integration tests

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/NestedFieldQueryIT.java
```diff
@@ -170,6 +172,107 @@ public void singleCondition() throws IOException {
...
+                            )
+                        )
+                    )
+                )
+            )
+        );
+    }
+
+    @Test
+    @Ignore("Deep query with WHERE condition is not fully supported")
```
galkk: Please also add a test (could also be ignored) with group by

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
zhongnansu: you got me. will fix it. I am thinking why don't we use any auto-formatting tool. Because when I do gradle build,  It gives me so many warnings, and even a report on style checking. Not only does it contain classes I changed, but also many other classes. That's where I found we are not using any formatting tool.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: will fix that

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
zhongnansu: good point

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: It could. But it doesn't matter, because it doesn't break the code and returns that exact path back to us.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -57,14 +62,21 @@ public NestedFieldProjection(SearchRequestBuilder request) {
...
             List<NestedQueryBuilder> nestedQueries = extractNestedQueries(query());
             groupFieldNamesByPath(fields).forEach(
-                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path))
+                (path, fieldNames) -> buildInnerHit(fieldNames, findNestedQueryWithSamePath(nestedQueries, path, fieldNames))
             );
         }
     }
 
+    /** Reset values of lastFieldInPath and lastFieldNames to prepare for next check */
+    private void setLastValues(String path, List<String> fieldNames) {
```
dai-chen: rename to avoid unclear `values`? maybe `setLastField...`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -45,6 +47,9 @@
 public class NestedFieldProjection {
 
     private final SearchRequestBuilder request;
+    private static final String SEPARATOR = ".";
+    private String lastFieldInPath = "";
```
dai-chen: I think we need to add more detailed comments when introducing state to a class. Please refer to https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldRewriter.java#L32.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
penghuo: Is it possible to simply this If-else logic with something below
```
            String fullPath = scope.getFullPath(table.path());
            String prefix = fullPath.isEmpty() ? "" : fullPath+".";
            scope.addAliasFullPath(emptyIfNull(expr.getAlias()), prefix+table.name());
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: The logic here is created a new nested query containing this existing nested query. Then we remove that existing nested query from `request`. Finally we add a new nested query block to the must domain ` ((BoolQueryBuilder) query().filter().get(0)).must(newNestedQuery);` So It is outdated because we have generated a new one to replace his position.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: - Use TreeMap to generate mapping {path -> list of field domain object} in reverse alphabetical order. When we use `foreach` to traverse this map, then this order is used to generate nested query from the deepest level to the top level.
- e.g. path `project.address.city` should be nested in `project.address`, should be nested in `project`

I will add documentation for the method.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: Agreed! will apply to the logic.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: will use peek here. Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: After removing the inner `if` condition. I'll also remove the documentation as well.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: will change to `setLastField`

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: will add documentation to class level as dai-chen pointed out. https://github.com/opendistro-for-elasticsearch/sql/pull/153#discussion_r314913904

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/From.java
zhongnansu: Will simplify if-else logic, thx!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/NestedFieldQueryIT.java
zhongnansu: Haven't figured out the right DSL for `GROUP BY`, but I will look into that.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: will remove the fieldNames, will clean the inner logic and then no need to pass this variable.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: Added documentation with similar state transition table.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/NestedFieldProjectionTest.java
zhongnansu: No problem

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/Identifier.java
```diff
@@ -52,15 +52,15 @@ void rewrite(Scope scope) {
     }
 
     String path() {
-        return separatorIndex() == -1 ? "" : expr.getName().substring(0, separatorIndex());
+        return lastSeparatorIndex() == -1 ? "" : expr.getName().substring(0, lastSeparatorIndex());
     }
 
     String name() {
-        return expr.getName().substring(separatorIndex() + 1);
+        return expr.getName().substring(lastSeparatorIndex() + 1);
```
arsen-es: Is it possible to have multi-field in a deeply nested object, e.g. have something like `table1.subobject1.field1.keyword`, where the field name is actually `field1.keyword`, not just `keyword`? I am not saying you should support it with this change, but if yes, please add a TODO comment, or better create an issue of type `Enhancement`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
arsen-es: I am still not sure why we need alphabetical order. Ideally, we should output the resulting fields in the same order as specified in the query.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +217,59 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
-                             filter(query -> isSamePath(path, query)).
-                             findAny().
-                             orElseGet(createEmptyNestedQuery(path));
+                filter(query -> isSamePath(path, query)).
+                peek(x -> setLastPath(path)).
+                findAny().
+                orElseGet(createNestedQuery(path));
     }
 
     private boolean isSamePath(String path, NestedQueryBuilder query) {
```
arsen-es: could you add a comment on this method explaining why paths are equal only if queries are equal?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +217,59 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
+     * groupFieldNamesByPath() uses TreeMap to maintain reverse alphabetical order. Therefore we use previous
+     * path to compare with current path, to help decide if we are done with keeping wrapping nested query
+     * to current nested query.
+     * */
+    private boolean isParentPath(String path, QueryBuilder query) {
+        // query could also be termQueryBuilder, where condition will generate termQueryBuilder
+        if (lastPath.isEmpty() || !(query instanceof NestedQueryBuilder)) {
+            return false;
+        } else {
+            int lastSeparatorIndex = lastPath.lastIndexOf(SEPARATOR);
```
arsen-es: np: seems like these two lines are duplicate from the code in Identifier.java. Maybe move these utility method to a separate class and use it from both places?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +217,59 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
+            final NestedQueryBuilder newNestedQuery;
+            List<QueryBuilder> mustList = ((BoolQueryBuilder) query().filter().get(0)).must();
+            int lastIndex = mustList.size() - 1;
+
+            if (mustList.isEmpty() || !isParentPath(path, mustList.get(lastIndex))) {
+                // create empty nested query
+                newNestedQuery = nestedQuery(path, matchAllQuery(), ScoreMode.None);
+            } else {
+                // create a nested query containing another nested query
+                NestedQueryBuilder oldNestedQuery = (NestedQueryBuilder) mustList.get(lastIndex);
```
arsen-es: I would love to have more documentation on this method especially when you introduce the `oldNestedQuery` along with `newNestedQuery`. Since we are not all very familiar with all the codepaths, it will be very useful if someone spending time to understand one path will share his/her knowledge in comments.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
zhongnansu: Sure I will, but this method is not my change. Some of my understanding:  there is no way to get the path from `query`, which is an instance of `NestedQueryBuider`. It just doesn't have a `getter` method for path, and the `path` variable can not be accessed. But any NestedQueryBuilder is initialized by a path, so here it compares the path by first creating a tmp query and then compare with the existing query.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -123,23 +217,59 @@ private void buildInnerHit(List<String> fieldNames, NestedQueryBuilder query) {
...
-            NestedQueryBuilder nestedQuery = nestedQuery(path, matchAllQuery(), ScoreMode.None);
-            ((BoolQueryBuilder) query().filter().get(0)).must(nestedQuery);
-            return nestedQuery;
+            final NestedQueryBuilder newNestedQuery;
+            List<QueryBuilder> mustList = ((BoolQueryBuilder) query().filter().get(0)).must();
+            int lastIndex = mustList.size() - 1;
+
+            if (mustList.isEmpty() || !isParentPath(path, mustList.get(lastIndex))) {
+                // create empty nested query
+                newNestedQuery = nestedQuery(path, matchAllQuery(), ScoreMode.None);
```
arsen-es: is there a way to not specify the `ScoreMode`?

zhongnansu: All constructors of `NestedQueryBuilder` require `ScoreMode` as an argument

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/nestedfield/NestedFieldProjection.java
```diff
@@ -41,10 +43,93 @@
...
+ *  |                   |        },                                                      |
+ *  |                   |        {                                                       |
+ *  |                   |          "nested": {                                           |
+ *  |                   |            "query": {                                          |
+ *  |                   |              "nested": {                                       |
+ *  |                   |                "query":{},                                     |
+ *  |                   |                "path": "comments.message"                      |
+ *  |                   |               }                                                |
+ *  |                   |             },                                                 |
+ *  |                   |          "path": "comments"                                    |
```
penghuo: does the comments be added only after processing comments?

zhongnansu: will add another column for `cur path`

anirudha:
deprecated feature

Issue #122, opened 2019-07-18
query with custom-function doesn't respect LIMIT


galkk:
The following request
```
GET _opendistro/_sql 
{
  "query" : "select date_format(utc_time, 'YYYY-MM-dd'), count(*) from kibana_sample_data_logs group by date_format(utc_time, 'YYYY-MM-dd') order by date_format(utc_time, 'YYYY-MM-dd') desc limit 1"
}
```

is expected to return 1 row, but returns all rows (limit is ignored)

galkk:
Related to #75

galkk:
Referenced PR sets requested limit properly, so it fixed this issue as well

Issue #156, opened 2019-08-20
Same query produces different results for JDBC format


galkk:
The result of first query has only 10 first rows, but has correct order and the key/value.

Expected behavior
1. Increase size for first query to at least 200
2. Second query should throw an error (disallow alias in GROUP BY)

#147 
opendistro-for-elasticsearch/sql-jdbc#23

```
GET /_opendistro/_sql?format=jdbc
{
  "query": """
  SELECT date_format(utc_time, 'Y-MM-dd') x, count(*) 
  FROM kibana_sample_data_logs
  WHERE host='www.elastic.co'
  GROUP BY date_format(utc_time, 'Y-MM-dd')
  """
}
```

The result of this query has supposedly all the rows, but has missing field in the output
```
GET /_opendistro/_sql?format=jdbc
{
  "query": """
  SELECT date_format(utc_time, 'Y-MM-dd') x, count(*) 
  FROM kibana_sample_data_logs
  WHERE host='www.elastic.co'
  GROUP BY x
  """
}
```

dai-chen:
I tested both queries and issue seems gone. See the response below.

```
{
  "schema": [
    {
      "name": "x",
      "alias": "x",
      "type": "text"
    },
    {
      "name": "COUNT(*)",
      "type": "integer"
    }
  ],
  "total": 62,
  "datarows": [
    [
      "2020-05-15",
      93
    ],
    [
      "2020-04-08",
      92
    ],
    [
      "2020-04-16",
      91
    ],
    [
      "2020-04-06",
      90
    ],
    [
      "2020-04-28",
      89
    ],
    [
      "2020-04-13",
      88
    ],
    [
      "2020-04-22",
      88
    ],
    [
      "2020-04-24",
      88
    ],
    [
      "2020-06-01",
      87
    ],
    [
      "2020-04-17",
      86
    ],
    [
      "2020-05-20",
      86
    ],
    [
      "2020-05-02",
      85
    ],
    [
      "2020-05-06",
      85
    ],
    [
      "2020-05-11",
      85
    ],
    [
      "2020-05-12",
      84
    ],
    [
      "2020-04-25",
      83
    ],
    [
      "2020-05-16",
      83
    ],
    [
      "2020-04-12",
      82
    ],
    [
      "2020-04-30",
      82
    ],
    [
      "2020-04-20",
      81
    ],
    [
      "2020-05-17",
      81
    ],
    [
      "2020-04-14",
      80
    ],
    [
[...]

PR #160, opened 2019-08-21
More records in the aggregation query output for script functions, is…

galkk:
Issue #156 
Issue #opendistro-for-elasticsearch/sql-jdbc#23

*Description of changes:*

Changes DSL, generated for aggregation query to have more records returned by default.
Testing: 

unit tests,

    ./gradlew build,

and manual testing via kibana ui

```
GET /_opendistro/_sql
{
  "query": """
  SELECT date_format(utc_time, 'Y-MM-dd') x, count(*) 
  FROM kibana_sample_data_logs
  WHERE host='www.elastic.co'
  GROUP BY date_format(utc_time, 'Y-MM-dd')
  """
}
```

to 
```
GET /kibana_sample_data_logs/_search
{
  "from": 0,
  "size": 0,

...

  "_source": {
    "includes": [
      "script",
      "COUNT"
    ],
    "excludes": []
  },
  "stored_fields": "x",
  "script_fields": {
    "x": {
      "script": {
        "source": "def date_format_1282644910 = DateTimeFormatter.ofPattern('Y-MM-dd').withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(doc['utc_time'].value.getMillis()));return date_format_1282644910;",
        "lang": "painless"
      },
      "ignore_failure": false
    }
  },
  "aggregations": {
    "x": {
      "terms": {
        "script": {
          "source": "def date_format_1282644910 = DateTimeFormatter.ofPattern('Y-MM-dd').withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(doc['utc_time'].value.getMillis()));return date_format_1282644910;",
          "lang": "painless"
        },
        "size": 200, <----------------------------------------------------------------------------------------------------------
        "min_doc_count": 1,
        "shard_min_doc_count": 0,
        "order": [
          {
            "_count": "desc"
          },
          {
            "_key": "asc"
          }
        ]
      },
      "aggregations": {
        "COUNT(*)": {
          "value_count": {
            "field": "_index"
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
@@ -76,11 +76,10 @@ public class AggregationQueryAction extends QueryAction {
             if (!groupBy.isEmpty()) {
                 Field field = groupBy.get(0);
 
-
                 //make groupby can reference to field alias
                 lastAgg = getGroupAgg(field, select);
 
-                if (lastAgg != null && lastAgg instanceof TermsAggregationBuilder && !(field instanceof MethodField)) {
+                if (lastAgg instanceof TermsAggregationBuilder && !(field instanceof MethodField)) {
                     //if limit size is too small, increasing shard  size is required
                     if (select.getRowCount() < 200) {
                         ((TermsAggregationBuilder) lastAgg).shardSize(2000);
@@ -92,9 +91,13 @@ public class AggregationQueryAction extends QueryAction {
                             }
                         }
                     }
-                    if(select.getRowCount()>0) {
+                    if (select.getRowCount() > 0) {
                         ((TermsAggregationBuilder) lastAgg).size(select.getRowCount());
                     }
+                }
+
+                if (lastAgg instanceof TermsAggregationBuilder && field instanceof ScriptMethodField && select.getRowCount() > 0) {
+                    ((TermsAggregationBuilder) lastAgg).size(select.getRowCount());
                 }
 
                 if (field.isNested()) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
@@ -79,25 +79,31 @@ public class AggregationQueryAction extends QueryAction {
                 //make groupby can reference to field alias
                 lastAgg = getGroupAgg(field, select);
 
-                if (lastAgg instanceof TermsAggregationBuilder && !(field instanceof MethodField)) {
-                    //if limit size is too small, increasing shard  size is required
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
```diff
@@ -92,11 +91,15 @@ public SqlElasticSearchRequestBuilder explain() throws SqlParseException {
                             }
                         }
                     }
-                    if(select.getRowCount()>0) {
+                    if (select.getRowCount() > 0) {
                         ((TermsAggregationBuilder) lastAgg).size(select.getRowCount());
                     }
                 }
 
+                if (lastAgg instanceof TermsAggregationBuilder && field instanceof ScriptMethodField && select.getRowCount() > 0) {
```
dai-chen: Any chance to add this to the `if` block above by refactoring it? 

And with your changes, for both regular and script method field, we set `size`. What about method field?

galkk: I was thinking about it, but proper fix quickly becomes much larger starts breaking ton of seemingly unrelated integration tests with different output (e.g. JSON response contains old size), so I wanted to make targeted fix for particular customer issue than full-blown refactoring.

When/if we will decide to fix ignored test, then it's going to be implemented

arsen-es: @galkk  I think even without changing the logic, keeping the outer `if` for only checking the object type will look cleaner:

```java
if (lastAgg instanceof TermsAggregationBuilder) {
  if (!(field instanceof MethodField) {
    ...
  } else if (field instanceof ScriptMethodField && select.getRowCount() > 0) {
    ...
  }
}
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/DateFormatTest.java
```diff
@@ -34,6 +34,7 @@
 import org.elasticsearch.index.query.RangeQueryBuilder;
 import org.hamcrest.*;
```
arsen-es: could you please get rid of * import?

galkk: The reges import .*\.\* shows 30 matches in 17 files, I think it's better to do that cleanup in separate PR

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
galkk: Thanks, I've got what you've meant.

I think that it could be even simplier (with ||), added the change

galkk:
Fixed #122 as well

PR #163, opened 2019-08-23
Enabled checkstyle and fixed the issues for the code to build

arsen-es:
*Issue #, if available:* N/A

*Description of changes:* To avoid making comments on everyone's PRs about style and formatting, I am enabling checkstyle. Fixed all the issues to make the build succeed, plus fixed some compiler warnings along the way.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+        "https://checkstyle.org/dtds/configuration_1_3.dtd">
+
+<module name="Checker">
+  <property name="charset" value="UTF-8" />
+
+  <module name="SuppressionFilter">
+    <property name="file" value="${suppressions}" />
+  </module>
+
+  <!-- Checks Java files and forbids empty Javadoc comments -->
+  <module name="RegexpMultiline">
+    <property name="id" value="EmptyJavadoc"/>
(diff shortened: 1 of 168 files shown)


arsen-es:
For now, I fixed the issues in src folder only, someone will need to take care about test later, it would be a bit too much for single PR.

galkk:
I'm fast tracking this change, good job

config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,120 @@
...
+
+  <!-- Checks that no tab characters are used in source files -->
+  <module name="FileTabCharacter">
+    <property name="eachLine" value="true"/>
+  </module>
+
+  <module name="TreeWalker">
+    <module name="LineLength">
+      <property name="max" value="120"/>
+      <property name="ignorePattern" value="^ *\* *https?://[^ ]+$|^ ?\* +\+\-+|^ ?\* +\||^ ?\* ?\-+"/>
```
dai-chen: Does this pattern already ignore `import` or not? Should we ignore it too because some package like our own `opendistroforelasticsearch` is quite long? https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml#L51

arsen-es: From official checkstyle documentation (https://checkstyle.sourceforge.io/config_sizes.html#LineLength):

>Package and import statements (lines matching pattern ^(package|import) .*) are not verified by this check.

arsen-es: This pattern is for javadoc comments, mostly the long ones we have for tables (state transitions, options, etc.) in comments, which don't fit in 120 char length.

Issue #158, opened 2019-08-20
Aliases aren't working for ORDER BY and GROUP BY


galkk:
```
POST /_opendistro/_sql
{
    "query": """
      SELECT OriginCountry country 
      FROM kibana_sample_data_flights 
      ORDER BY country DESC
       """
}
```

```
POST /_opendistro/_sql
{
    "query": """
      SELECT OriginCountry country 
      FROM kibana_sample_data_flights 
      GROUP BY country DESC
       """
}
```

```
GET /_opendistro/_sql
{
  "query": """
  SELECT date_format(utc_time, 'Y-MM-dd') date
  FROM kibana_sample_data_logs
  WHERE host='www.elastic.co'
  ORDER BY date DESC
  """
}
```

Returns an error

```
{
  "error": {
    "root_cause": [
      {
        "type": "query_shard_exception",
        "reason": "No mapping found for [date] in order to sort on",
        "index_uuid": "epE-j4sESu-65Qakx_rDgw",
        "index": "kibana_sample_data_logs"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "kibana_sample_data_logs",
        "node": "smjPOjlXQ6enw_cfj9Lhug",
        "reason": {
          "type": "query_shard_exception",
          "reason": "No mapping found for [date] in order to sort on",
          "index_uuid": "epE-j4sESu-65Qakx_rDgw",
          "index": "kibana_sample_data_logs"
        }
      }
    ]
  },
  "status": 400
}
```

While same query without script function works without any issues

```
GET /_opendistro/_sql
{
  "query": """
  SELECT clientip ip
  FROM kibana_sample_data_logs
  WHERE host='www.elastic.co'
  ORDER BY ip DESC
  """
}
```

galkk:
The problems from the issue description were fixed by PR 171

Issue #149, opened 2019-08-12
Fix script functions: remove randomization


galkk:
* Remove any randomization from generated scripts in SQLFunctions. All parameters must be passed as params.

Currently, the scripts for SQLFUnctions randomize names, for example the query

```
GET /_opendistro/_sql/_explain
{
  "query": """
    select date_format(utc_time, 'dd-MM-yyyy') as t FROM kibana_sample_data_logs WHERE date_format(utc_time, 'dd-MM-yyyy') 
  """
}
```

produces the following script, with randomized names

```json
{"from":0,"size":200,
"_source":{"includes":[],"excludes":[]},
"script_fields":{"t":{"script":{"source":"def date_format_1400932066 = DateTimeFormatter.ofPattern('dd-MM-yyyy').withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(doc['utc_time'].value.getMillis()));return date_format_1400932066;","lang":"painless"},"ignore_failure":false}}}
```

As per ElasticSearch docs, the scripts are compiled and cached.

https://www.elastic.co/guide/en/elasticsearch/reference/7.3/modules-scripting-using.html#prefer-params

> If you need to pass variables into the script, you should pass them in as named params instead of hard-coding values into the script itself. For example, if you want to be able to multiply a field value by different multipliers, don’t hard-code the multiplier into the script:
  "source": "doc['my_field'] * 2"

> Instead, pass it in as a named parameter:

  "source": "doc['my_field'] * multiplier",
  "params": {
    "multiplier": 2
  }
> The first version has to be recompiled every time the multiplier changes. The second version is only compiled once.

[mine] Even more, there are potential unexpected issues

 **If you compile too many unique scripts within a small amount of time, Elasticsearch will reject the new dynamic scripts with a circuit_breaking_exception error. By default, up to 15 inline scripts per minute will be compiled. You can change this setting dynamically by setting script.max_compilations_rate.**

galkk:
The randomization was replaced with sequential id generation

PR #171, opened 2019-08-29
Adds ability to have aliases for ORDER BY and GROUP BY expressions

galkk:
*Issue #158 *

### Description of changes:

1. Enables grouping by and ordering by aliased field, allowing the following query:

```
SELECT date_format(utc_time, 'dd-MM-YYYY') date 
FROM kibana_sample_data_logs 
GROUP BY date 
ORDER BY date DESC
```

to generate proper DSL
```
GET /kibana_sample_data_logs/_search
{
  "from": 0,
  "size": 0,
  "_source": {
    "includes": [
      "script"
    ],
    "excludes": []
  },
  "stored_fields": "date",
  "script_fields": {
    "date": {
      "script": {
        "source": "def date_format_1 = DateTimeFormatter.ofPattern('dd-MM-YYYY').withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(doc['utc_time'].value.getMillis()));return date_format_1;",
        "lang": "painless"
      },
      "ignore_failure": false
    }
  },
  "aggregations": {
    "date": {
      "terms": {
        "script": {
          "source": "def date_format_1 = DateTimeFormatter.ofPattern('dd-MM-YYYY').withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(doc['utc_time'].value.getMillis()));return date_format_1;",
          "lang": "painless"
        },
        "size": 200,
        "min_doc_count": 1,
        "shard_min_doc_count": 0,
        "order": {
          "_key": "desc"
        }
      }
    }
  }
}

```

2. Introduced **deterministic** identifier generation into SQLFunctions/FieldGeneration, instead of random field generation (see **def date_format_1** in the DSL): #149 

3. This change reversed previous approach with OrderByInliner and moves inlining/analysis into later stages of DSL generation. 

### Testing
Unit tests, integration tests, manual tests via kibana


src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
```diff
@@ -247,6 +245,12 @@ private AggregationBuilder getGroupAgg(Field field, Select select) throws SqlPar
         }
 
         if (!refrence) {
+            for (Field f: select.getFields()) {
+                if (f.getAlias()!=null && f.getAlias().equals(field.getName())) {
```
dai-chen: Could we save the null check by switch 2 sides like `field.getName().equals(f.getAlias())`. I assume field name is required?

Btw, nitpick about the space. I think we need to add this rule to our checkstyle config. :)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CastParser.java
```diff
@@ -48,7 +49,7 @@ public String parse(boolean isReturn) throws SqlParseException {
 
         String dataType = castExpr.getDataType().getName().toUpperCase();
         String fileName = String.format("doc['%s'].value", Util.expr2Object(castExpr.getExpr()));
-        String name = SQLFunctions.randomize("field");
+        String name = "field" + (++fieldId);
```
dai-chen: Is it possible this name conflict with what generated in SQLFunctions? Because I notice we do have a method called field.

dai-chen: I think we need to make sure either they have different namespace (prefix) or ideally maybe a global incremental counter per request.

galkk: I wasn't able to come up with the example that could break it, and, in reality, the risk was always there, because we do not track generated identifiers.

galkk: Technically it's already close to it, because we have one SqlParser per request (see my answer to @penghuo). I like the suggestion of different prefix, implemented

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
galkk: Changed entire method to be more readable

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CastParser.java
dai-chen: I think my question is a little different. I can understand we create new instance per request. But essentially you're using 2 different counters (both starts from 0) with potential conflicting prefixes ("field" and method name which could be "field" too).

galkk: >  potential conflicting prefixes ("field" and method name which could be "field" too).

That is fixed in last change

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -254,12 +253,10 @@ public static boolean isFunctionTranslatedToScript(String function) {
...
     }
 
-    private static String random() {
-        return Math.abs(new Random().nextInt()) + "";
-    }
+    private int generatedId = 0;
 
-    public static String randomize(String methodName) {
-        return methodName + "_" + random();
+    public String nextId(String methodName) {
```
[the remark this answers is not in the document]
penghuo: Sounds good.

Issue #176, opened 2019-09-05
ORDER BY expects function names written only in lowercase


galkk:
This 
```
GET /_opendistro/_sql
{
  "query": """
    SELECT DATE_FORMAT(utc_time,'YYYY-MM-DD') from kibana_sample_data_logs
    ORDER BY date_format(utc_time,'YYYY-MM-DD')
  """
}
```
works, but this

```
GET /_opendistro/_sql
{
  "query": """
    SELECT DATE_FORMAT(utc_time,'YYYY-MM-DD') from kibana_sample_data_logs
    ORDER BY DATE_FORMAT(utc_time,'YYYY-MM-DD')
  """
}
```

fails with 

```
{
  "error": {
    "root_cause": [
      {
        "type": "unsupported_operation_exception",
        "reason": "The following method is not supported in Schema: DATE_FORMAT"
      }
    ],
    "type": "unsupported_operation_exception",
    "reason": "The following method is not supported in Schema: DATE_FORMAT"
  },
  "status": 500
}
```

Issue #175, opened 2019-09-05
ORDER BY doesn't respect table aliases


galkk:
The following query produces wrong DSL and fails

```
GET /_opendistro/_sql/_explain
{
  "query": """
  SELECT * 
  FROM kibana_sample_data_logs k
  ORDER BY k.host
  """
}

```

Produces the following DSL

```
{"from":0,"size":200,"sort":[{"k.host":{"order":"asc"}}]}
```

and throws an exception

```
{
  "error": {
    "root_cause": [
      {
        "type": "query_shard_exception",
        "reason": "No mapping found for [k.host] in order to sort on",
        "index_uuid": "epE-j4sESu-65Qakx_rDgw",
        "index": "kibana_sample_data_logs"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "kibana_sample_data_logs",
        "node": "smjPOjlXQ6enw_cfj9Lhug",
        "reason": {
          "type": "query_shard_exception",
          "reason": "No mapping found for [k.host] in order to sort on",
          "index_uuid": "epE-j4sESu-65Qakx_rDgw",
          "index": "kibana_sample_data_logs"
        }
      }
    ]
  },
  "status": 400
}```

Issue #128, opened 2019-07-23
Add support for case-changing functions


galkk:
Implement support for case-changing functions

As per [#123](https://github.com/opendistro-for-elasticsearch/sql/issues/123), there is no support for case change functions in SQL.

Implement custom script functions that will do that (SQLFunctions.java)

Reference

* https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-functions-string.html#sql-functions-string-lcase

* https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-functions-string.html#sql-functions-string-ucase

galkk:
This is a sketch of how the potential solution might look like

```
POST /kibana_sample_data_logs/_search
{
  "from": 0,
  "size": 200,
  "query": {
    "bool": {
      "filter": [
        {
          "bool": {
            "must": [
              {
                "script": {
                  
                  "script": {
                    "source": "doc['host.keyword'].value == params.v0.toLowerCase()",
                    "lang": "painless",
                    "params" : {
                      "v0": "WWW.ELASTIC.CO"
                    }
                  },
                  "boost": 1
                }
              }
            ],
            "adjust_pure_negative": true,
            "boost": 1
          }
        }
      ],
      "adjust_pure_negative": true,
      "boost": 1
    }
  },
  "_source": {
    "includes": [
      "host"
    ],
    "excludes": []
  }
}
```

PR #177, opened 2019-09-05
Case functions support: LOWER, UPPER

galkk:
Issue #128 

*Description of changes:*
* Added 2 new functions LOWER and UPPER that receive field name as a first parameter and, potentially, locale 

* Changed identifier generation strategy to id per function name instead of global id

While I feel that SQLFunctions should be refactored, to expedite the process as of now I'm adding just functions implemented by analogy with trim() and substring() functions

### Testing
integration testing, testing via kibana

Itegration tests include ORDER BY and GROUP BY clauses to ensure that everything is working.
Found 2 new issues that will be addressed separately: #176 , #175 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
@@ -28,7 +28,9 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import org.elasticsearch.common.collect.Tuple;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -50,7 +52,7 @@ public class SQLFunctions {
     );
 
     private static final Set<String> stringOperators = Sets.newHashSet(
-            "split", "concat_ws", "substring", "trim"
+            "split", "concat_ws", "substring", "trim", "lower", "upper"
     );
 
     private static final Set<String> binaryOperators = Sets.newHashSet(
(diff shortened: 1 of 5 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -109,6 +104,59 @@ public void functionAlias() throws Exception {
...
+    }
+
+    @Test
+    public void caseChangeWithAggregationTest() throws IOException {
+        String query = "SELECT UPPER(e.firstname), COUNT(*)" +
+                "FROM elasticsearch-sql_test_index_account/account e " +
+                "WHERE LOWER(e.lastname)='duke' " +
+                "GROUP BY UPPER(e.firstname) ";
+
+        System.out.println(executeQuery(query));
```
penghuo: remove this.

galkk: Done

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -74,6 +76,20 @@
             utilityFunctions)
             .flatMap(Set::stream).collect(Collectors.toSet());
 
+    private HashMap<String, Integer> generatedIds = new HashMap<>();
```
dai-chen: Use interface `Map` instead?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -74,6 +76,20 @@
...
+
+    /**
+     * Generates next id for given method name. The id's are increasing for each method name, so
+     * nextId("a"), nextId("a"), nextId("b") will return a_1, a_2, b_1
+     */
+    public String nextId(String methodName) {
```
penghuo: I think currently, the methodName is cached in request level. If i am right, does this cached method could be shared by different request?

galkk: No. The SqlFunctions is per-request, and this is instance field

dai-chen: Not sure if this look clear to you but it seems work for your purpose:

```
  private String nextId(String methodName) {
    return methodName + generatedIds.merge(methodName, 1, Integer::sum);
  }
```

galkk: Nice, i like it, thanks, will do

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
galkk: This is private field, so knowing detailed type could be beneficial in common case, but here Map will work as well, changed.

PR #180, opened 2019-09-10
Add parent for SQLExpr in AST if missing

penghuo:
*Issue #, if available:*

*Description of changes:*
Add parent for SQLExpr in AST if missing. In current implementation, this logic is in NestedFiledRewriter, this CR trying to seperate this logic out of it.


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/parent/SQLExprParentSetterRuleTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertTrue;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SQLExprParentSetterRuleTest {
+
+    @Mock
+    SQLQueryExpr queryExpr;
```
dai-chen: np: could be `private` and same for `rule`

Issue #188, opened 2019-09-19
Having doesn't working on nested field.


penghuo:
### SQL
```
    SELECT e.name, MAX(p.started_year) as max
    FROM employee_nested_it AS e, e.projects AS p
    WHERE p.name LIKE '%security%'
    GROUP BY e.name
    HAVING max > 0
```
### Test Data
https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/test/resources/employee_nested.json

### Expected Result
```
      "buckets" : [
        {
          "key" : "Bob Smith",
          "doc_count" : 1,
          "projects.started_year@NESTED" : {
            "doc_count" : 3,
            "max" : {
              "value" : 2015.0
            }
          }
        },
        {
          "key" : "Jane Smith",
          "doc_count" : 1,
          "projects.started_year@NESTED" : {
            "doc_count" : 2,
            "max" : {
              "value" : 2015.0
            }
          }
        }
      ]
```

### Actual Result
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [max]"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "employee_nested_it",
        "node": "_2_opYNCR0a53f0tlIlm4Q",
        "reason": {
          "type": "illegal_argument_exception",
          "reason": "No aggregation found for path [max]"
        }
      }
    ],
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "No aggregation found for path [max]",
      "caused_by": {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [max]"
      }
    }
  },
  "status": 400
}
```

Issue #187, opened 2019-09-19
Aggregation on Nested Array Field return unexpected value


penghuo:
The aggregatioin on nested array field return unexpected value.
SQL
```
    SELECT e.name, COUNT(p) as count
    FROM employee_nested_it AS e, e.projects AS p
    GROUP BY e.name
```

Test data: https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/test/resources/employee_nested.json

Expected Result
```
      "buckets" : [
        {
          "key" : "Bob Smith",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 3,
            "count" : {
              "value" : 3
            }
          }
        },
        {
          "key" : "Jane Smith",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 2,
            "count" : {
              "value" : 2
            }
          }
        },
        {
          "key" : "John Doe",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 0,
            "count" : {
              "value" : 0
            }
          }
        },
        {
          "key" : "Susan Smith",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 0,
            "count" : {
              "value" : 0
            }
          }
        }
      ]
```

Actual Result
```
      "buckets" : [
        {
          "key" : "Bob Smith",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 3,
            "count" : {
              "value" : 0
            }
          }
        },
        {
          "key" : "Jane Smith",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 2,
            "count" : {
              "value" : 0
            }
          }
        },
        {
          "key" : "John Doe",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 0,
            "count" : {
              "value" : 0
            }
          }
        },
        {
          "key" : "Susan Smith",
          "doc_count" : 1,
          "projects@NESTED" : {
            "doc_count" : 0,
            "count" : {
              "value" : 0
            }
          }
        }
     ]
```

PR #189, opened 2019-09-19
Fix issue related to group by and having.

penghuo:
1. Issue #187 
2. Issue #188 

*Issue #, if available:*

*Description of changes:*


src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/bucketpath/Path.java
```diff
@@ -0,0 +1,27 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.domain.bucketpath;
+
+public interface Path {
+
+    String name();
+
+    String separator();
```
dai-chen: I feel methods in this abstraction doesn't take advantage of polymorphism. Because currently each subclass is more like data class without much different behavior.

I suggest we either move logic depending on it to each subclass or just use a data class for now:

```
class Path {
    name, separator, type ...
    getters ...
}
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/HavingTest.java
```diff
@@ -206,6 +222,15 @@ public void aggregationFunctionInHavingRatherThanAlias() {
     private Collection<PipelineAggregationBuilder> translate(SQLQueryExpr expr) {
         try {
             Select select = new SqlParser().parseSelect(expr);
+            select.getFields().forEach(field -> {
+                try {
+                    new AggMaker()
+                            .withWhere(select.getWhere())
+                            .makeFieldAgg((MethodField) field, AggregationBuilders.terms(""));
+                } catch (SqlParseException e) {
+                    e.printStackTrace();
```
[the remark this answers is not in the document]
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/bucketpath/Path.java
penghuo: Done.

Issue #120, opened 2019-07-18
HAVING clause is throwing "Invalid SQL Query Exception" when there's an aggregate function


Nitinrajyadav:
for example 
`SELECT
  productName,
  productCode
FROM
  productsindex
WHERE
  (productName = 'test')
GROUP BY
  productCode
HAVING
  COUNT(calculatedQATS) < 20
`

throws exception  see **HAVING COUNT(calculatedQATS) < 20**

but 

`SELECT
  productName,
  productCode
FROM
  productsindex
WHERE
  (productName = 'test')
GROUP BY
  productCode
HAVING
 calculatedQATS < 20
`

Works  see **HAVING calculatedQATS < 20**

also, this's a random query copied from w3school (replaced table-name with valid index-name)
`SELECT COUNT(CustomerID), Country FROM productsindex GROUP BY Country HAVING COUNT(CustomerID) > 5` 
this also throws **invalid query exception**

dai-chen:
Thanks for reporting the issue! We don't have support for your cases:

 + I assume you're trying to use `HAVING` on fields that have nothing to do with fields in `SELECT`.
 + Alias for aggregate expressions present in `HAVING` is required in our plugin. Please try query like this: `SELECT COUNT(CustomerID) AS cnt, Country FROM productsindex GROUP BY Country HAVING cnt > 5`

You can check out example in our README or test cases: https://github.com/opendistro-for-elasticsearch/sql#sql-usage

And also I've only seen same aggregate expression(s) present in `HAVING` and `SELECT`. I'm wondering is query in your first 2 examples valid in standard SQL?

Nitinrajyadav:
@dai-chen thanks for the clarification. 

> and to answer your question, I believe the first query in my question is to be valid in standard SQL syntax you can check here https://www.eversql.com/sql-syntax-check-validator/ I've tested that in Sequel-PRO.

> and the second Query, I was playing with the **/_explain** API endpoint, and it was excepted as valid SQL syntax, however, in Standard SQL, it's not valid, it should have **calculatedQATS** in SELECT statement as a column or alias.

dai-chen:
> @dai-chen thanks for the clarification.
> 
> > and to answer your question, I believe the first query in my question is to be valid in standard SQL syntax you can check here https://www.eversql.com/sql-syntax-check-validator/ I've tested that in Sequel-PRO.
> 
[...]

PR #195, opened 2019-09-25
Support aggregation function in Having

penghuo:
*Issue #, if available:* #120 

*Description of changes:*
The HavingParser which (1) rewrite the aggregation function as identifier. (2) construct the Field domain object from the aggregation function.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Having.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Having.java
@@ -19,16 +19,20 @@ import com.alibaba.druid.sql.ast.SQLExpr;
 import com.alibaba.druid.sql.ast.statement.SQLSelectGroupByClause;
 import com.alibaba.druid.util.StringUtils;
 import com.amazon.opendistroforelasticsearch.sql.exception.SqlParseException;
+import com.amazon.opendistroforelasticsearch.sql.parser.HavingParser;
 import com.amazon.opendistroforelasticsearch.sql.parser.NestedType;
 import com.amazon.opendistroforelasticsearch.sql.parser.WhereParser;
 import org.elasticsearch.script.Script;
 import org.elasticsearch.search.aggregations.AggregationBuilder;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.joining;
 import static org.elasticsearch.search.aggregations.PipelineAggregatorBuilders.bucketSelector;
@@ -52,6 +56,12 @@ public class Having {
      */
     private final List<Where> conditions;
 
+    private List<Field> havingFields;
+
+    public List<Field> getHavingFields() {
+        return havingFields;
+    }
+
     /**
      * Construct by HAVING expression
      *
@@ -60,7 +70,8 @@ public class Having {
      * @throws SqlParseException exception thrown by where parser
      */
     public Having(SQLExpr havingExpr, WhereParser parser) throws SqlParseException {
-        this.conditions = parseHavingExprToConditions(havingExpr, parser);
+        this.havingFields = new ArrayList<>();
+        this.conditions = parseHavingExprToConditions(havingExpr, new HavingParser(parser, havingFields));
     }
 
     /**
@@ -87,12 +98,16 @@ public class Having {
             return;
         }
 
+        // parsing the fields from SELECT and HAVING clause
+        List<Field> combinedFields = Stream.of(fields, getHavingFields())
+                .flatMap(List::stream)
+                .collect(Collectors.toList());
         groupByAgg.subAggregation(bucketSelector(BUCKET_SELECTOR_NAME,
-                contextForFieldsInSelect(fields),
+                contextForFieldsInSelect(combinedFields),
                 explainConditions()));
     }
 
-    private List<Where> parseHavingExprToConditions(SQLExpr havingExpr, WhereParser parser)
+    private List<Where> parseHavingExprToConditions(SQLExpr havingExpr, HavingParser parser)
             throws SqlParseException {
         if (havingExpr == null) {
             return Collections.emptyList();
@@ -215,5 +230,4 @@ public class Having {
         }
         return alias;
     }
-
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Having.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Having.java
@@ -26,7 +26,6 @@ import com.google.common.collect.Iterables;
 import org.elasticsearch.script.Script;
 import org.elasticsearch.search.aggregations.AggregationBuilder;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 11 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
```diff
@@ -189,6 +189,30 @@ public void groupByTest() throws Exception {
         Assert.assertThat(gender.query(femaleBucketPrefix + "/COUNT(*)/value"), equalTo(493));
     }
 
+    @Test
+    public void groupByHavingTest() throws Exception {
+
+        JSONObject result = executeQuery(String.format("SELECT gender " +
```
davidcui-amzn: Should it be StringUtils.format() vs String.format()?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/JSONRequestTest.java
```diff
@@ -125,6 +125,139 @@ public void aggWithWhereOnNestedOrNested() {
...
+    }
+
+    @Test
+    public void aggInHavingWithWhereOnParentAndNested() {
+        JSONObject explainSQL = explainSQLToJson("SELECT name " +
+                                                 "FROM employee " +
+                                                 "WHERE name LIKE '%smith%' AND nested(projects.name, 'projects') LIKE '%security%' " +
+                                                 "GROUP BY name " +
+                                                 "HAVING COUNT(nested(projects, 'projects')) > 1");
+        assertThat(
```
davidcui-amzn: This formatting looks inconsistent

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
```diff
@@ -189,6 +189,30 @@ public void groupByTest() throws Exception {
         Assert.assertThat(gender.query(femaleBucketPrefix + "/COUNT(*)/value"), equalTo(493));
     }
 
+    @Test
+    public void groupByHavingTest() throws Exception {
```
dai-chen: Could you also verify if JDBC or CSV formatter works?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/HavingParser.java
```diff
@@ -0,0 +1,118 @@
...
+    private final HavingConditionRewriter havingConditionRewriter;
+
+    public HavingParser(WhereParser whereParser, List<Field> havingFields) {
+        this.whereParser = whereParser;
+        this.havingFields = havingFields;
+        this.havingConditionRewriter = new HavingConditionRewriter();
+    }
+
+    public void parseWhere(SQLExpr expr, Where where) throws SqlParseException {
+        expr.accept(havingConditionRewriter);
```
dai-chen: How about we separate the rewrite and parseWhere? I'm thinking in this case this class doesn't need to share the `havingFields` with outside, for example:

```
    public List<Field> rewriteHavingClause() {
        // create rewriter and rewrite
        return Collections.unmodifiableList(rewriter.havingFields);
    }

    private static class HavingConditionRewriter ... {
        List<Field> havingFields = new ArrayList<>();
    }

File: src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Having.java [hunk 5]
```diff
@@ -87,12 +98,16 @@ public void explain(AggregationBuilder groupByAgg, List<Field> fields) throws Sq
             return;
         }
 
+        // parsing the fields from SELECT and HAVING clause
+        List<Field> combinedFields = Stream.of(fields, getHavingFields())
```
dai-chen: not necessary, if you like, we can avoid creating a new list by Guava's `Iterables.concat(list1, list2)` and pass result as Iterable view to `contextForFieldsInSelect`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/HavingParser.java
```diff
@@ -0,0 +1,118 @@
...
+        whereParser.parseWhere(expr, where);
+    }
+
+    private class HavingConditionRewriter extends MySqlASTVisitorAdapter {
```
dai-chen: I think we should make this inner class `static`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
penghuo: I think internally they have same implementation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/HavingParser.java
penghuo: I want to keep it as non static because it need to access the havingFields list in the HavingParser.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/JSONRequestTest.java
penghuo: Any formatting suggestion? I format the query by making the SQL clause aligend.

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Having.java
penghuo: Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/JSONRequestTest.java
davidcui-amzn: Sorry, should've been more clear- I was referring to differences such as the ones between lines 202-203 and line 207. On line 202, it appears as if query() lists its arguments on separate lines, whereas on line 207, they are all on the same line.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/parent/SQLExprParentSetterTest.java
```diff
@@ -34,12 +35,27 @@ public void testSQLInSubQueryExprHasParent() {
         expr.accept(new SQLExprParentExistsValidator());
     }
 
+    @Test
+    public void testSQLInListExprHasParent() {
+        String sql =
+                "SELECT * FROM TbA " +
+                "WHERE a in (10, 20)";
```
abbashus: np:`in` --> `IN`

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/JSONRequestTest.java
penghuo: Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
penghuo: Done. Add the IT aggregationFunctionInHaving in PrettyFormatResponseIT.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/parent/SQLExprParentSetterTest.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/HavingParser.java
penghuo: The idea to implement HavingParser is to encapsulate the procedure and differentate with WhereParser.

Issue #151, opened 2019-08-13
Improve query verification and exception handling


dai-chen:
Currently SQL Plugin doesn’t handle various exception cases well. This is partly due to the loose check in Druid parser in use, though majorly because of our own code.

For example, some bad query ends up throwing exception like null pointer or index out of bound, while others return confusing error message or exposing implementation details (because the message is generated from class automatically). Meanwhile, some query that meant to fail succeeds return empty response instead because of missing pre-verification.

dai-chen:
To mitigate or solve the problem, we're planning to add more verification for lexical, syntactic and semantic error cases and return much clearer message accordingly to help user understand what's going wrong and how to fix it.

dai-chen:
Code changes for syntax check with new ANTLR parser is already merged. Next phase for semantic check is still in progress.

Issue #82, opened 2019-06-14
Pre-verification before actual execution of query


dai-chen:
More verification should be performed before the actual execution of query in ES by useful information in metadata.

One example is run Nested Loop join on fields of text type which nothing matches from both sides. In this case, we should exit before the execution with useful message to customer rather than confusing empty result.

dai-chen:
The new semantic analysis is already merged with many pre-verifications. The Nested Loop example above should be covered elsewhere.

PR #201, opened 2019-09-30
Improve query verification by adding semantic analyzer

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/151, https://github.com/opendistro-for-elasticsearch/sql/issues/82

*Description of changes:* This PR covers the second part of query verification improvement which include several parts:

 1. **Semantic analyzer**: consists of Semantic Context and Type System.
 2. **Parse tree visitor**: drive the entire analysis process.
 3. **UT and IT**: new test cases.
 4. **Fix broken IT**: found quite a few bad IT. fixed them to pass IT.

The majority of code changes is completed though there are still some to-do items: skip semantic check in unit test, add scope checking, performance impact etc. Sending out this PR to get some feedback earlier.

*For reviewers*: please read `docs/dev/SemanticAnalysis.md` before looking into this big PR. The design doc is not complete yet but should give you many context for the code review.


src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/scope/SemanticContext.java
```diff
@@ -0,0 +1,78 @@
...
+        environment = curEnv.getParent();
+        return curEnv;
+    }
+
+    /**
+     * Return field mappings for an index
+     * @param indexName     index name
+     * @return              field mappings
+     */
+    public FieldMappings getMapping(String indexName) {
```
penghuo: It seems SemanticContext works as the Adapter of LocalClusterState which has hard depedency on ES. In my understanding, SemanticContext should have logically abstraction which not related to physical implementation. Is it possible to abstract the IndexMappings as MetaSchema to represent the restriction and attribute related to the table/index.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
```diff
@@ -0,0 +1,279 @@
...
+        return result;
+    }
+
+    @Override
+    public T visitAtomTableItem(AtomTableItemContext ctx) {
+        String alias = (ctx.alias == null) ? "" : ctx.alias.getText();
+        T result;
+        if (ctx.tableName() instanceof SimpleTableNameContext) {
+            String tableName = ctx.tableName().getText();
+            if (isPath(tableName)) {
```
penghuo: The isPath can't not handle the case if the index name include ".". I am wondering do we need to differentiate nested index when doing semantic analysis?

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/AggregateFunction.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Aggregate function
+ */
+public enum AggregateFunction implements TypeExpression {
+    COUNT(
+        func().to(INTEGER), // COUNT(*)
+        func(ES_TYPE).to(INTEGER)
+    ),
+    MAX(func(T(NUMBER)).to(T)),
```
penghuo: I think the purpose of func(T(NUMBER)).to(T) is to bind the return type to the actual paramater type. could we doing this by func(NUMBER).to(ArgType(0))?

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/Type.java
```diff
@@ -0,0 +1,89 @@
...
+ * Type interface which represents any type of symbol in the SQL.
+ */
+public interface Type extends Reducible {
+
+    /**
+     * Hide generic type ugliness and error check here in one place.
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    default <T extends Reducible> T reduce(List<T> others) {
```
penghuo: I think the reduce operation only applied to TypeExpression. Is it possible to limit it to TypeExpression only?

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
dai-chen: Good point. I wasn't aware of this. Need to figure out because name with dot can mess up the entire analysis if it's allowed in our plugin.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/AggregateFunction.java
dai-chen: Yes, let me come up with a better form for generic type.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/Type.java
dai-chen: Essentially `Reducible` and `reduce()` provide a generic interface to visitor so we can implement different logic. For example, we could have a AST generate visitor with AST implement `Reducible`. And we could have another implementation for test case or documentation generating. In one word, anything reducible can implement it here rather than limited to type expression.

docs/dev/SemanticAnalysis.md
```diff
@@ -0,0 +1,247 @@
...
+Previously SQL plugin didn't do semantic analysis, for example field doesn't exist, function call on field with wrong type etc. So it had to rely on Elasticsearch engine to perform the "check" which is actual execution. 
+This led to bad user experience because of missing careful verification, cost of actual execution and confusing error message. In this work, we built a new semantic analyzer based on the new ANTLR generated parser introduced recently.
+With the new semantic analyzer, we manage to perform various verification in terms of meaning of the query and return clear and helpful message to user for troubleshoot.
+
+---
+## 2.Use Cases
+
+Here are a few examples of semantic check with our analyzer implemented. You could get a taste of what is the benefit of new semantic analyzer:
+
+### 2.1 Field Does Not Exist
```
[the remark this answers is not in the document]
dai-chen: Sure, will change the title. Thanks!

docs/dev/SemanticAnalysis.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+### 2.5 Operator Use on Wrong Field Type
+
+```
+POST _opendistro/_sql
+{
+  "query": "SELECT * FROM accounts WHERE lastname IS FALSE"
+}
+
```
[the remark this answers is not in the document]
dai-chen: I missed TODO here. Will include the complete dev doc in next check in.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/ComparisonOperator.java
```diff
@@ -0,0 +1,78 @@
...
+public enum ComparisonOperator implements Type {
+
+    EQUAL("="),
+    NOT_EQUAL("<>"),
+    NOT_EQUAL2("!="),
+    GREATER_THAN(">"),
+    GREATER_THAN_OR_EQUAL_TO(">="),
+    SMALLER_THAN("<"),
+    SMALLER_THAN_OR_EQUAL_TO("<="),
+    IS("IS");
```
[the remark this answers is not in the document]
dai-chen: I convert function and operator name to uppercase in `SemanticAnalyzer`: https://github.com/dai-chen/sql/blob/improve-semantic-check/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzer.java#L224

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerAggregateFunctionTest.java
```diff
@@ -0,0 +1,154 @@
...
+        expectValidationFailWithErrorMessages(
+            "SELECT SUM(city) FROM semantics",
+            "Function [SUM] cannot work with [KEYWORD].",
+            "Usage: SUM(NUMBER T) -> T"
+        );
+    }
+
+    @Test
+    public void useAvgFunctionCallAliasInHavingShouldPass() {
+        validate("SELECT city, AVG(age) AS avg FROM semantics GROUP BY city HAVING avg > 10");
```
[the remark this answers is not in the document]
dai-chen: You mean format to multi-line right? Last time @galkk made a point that it's clear and easy to copy-paste it for troubleshooting in this way. Either is fine by me. Maybe we can switch all to multi-line style when we can use multi-line literal feature in Java 12: https://www.vojtechruzicka.com/raw-strings/

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java
```diff
@@ -44,8 +44,9 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static com.amazon.opendistroforelasticsearch.sql.esdomain.LocalClusterState.FieldMappings;
-import static com.amazon.opendistroforelasticsearch.sql.esdomain.LocalClusterState.IndexMappings;
+import com.amazon.opendistroforelasticsearch.sql.esdomain.mapping.FieldMappings;
+
```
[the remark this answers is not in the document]
dai-chen: interesting, not sure how this happened. fixed by reorganizing all imports by IDE. Thanks

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzer.java
```diff
@@ -0,0 +1,315 @@
...
+    /** Define by replace with alias */
+    private void redefineFieldNameByPrefixingAlias(String indexName, String alias) {
+        Map<String, Type> typeByFullName = environment().resolveByPrefix(new Symbol(Namespace.FIELD_NAME, indexName));
+        typeByFullName.forEach(
+            (fieldName, fieldType) -> defineFieldName(fieldName.replace(indexName, alias), fieldType)
+        );
+    }
+
+    private void defineFieldName(String fieldName, String type) {
+        defineFieldName(fieldName, BaseType.typeOf(type));
```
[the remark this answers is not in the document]
dai-chen: Just looked if it could be inlined. But it seems more compact with this overloaded method used here and next line followed: https://github.com/dai-chen/sql/blob/improve-semantic-check/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzer.java#L152

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/OpenDistroSqlAnalyzer.java
```diff
@@ -17,28 +17,66 @@
...
+import org.apache.logging.log4j.Logger;
 
 /**
  * Entry point for ANTLR generated parser to perform strict syntax and semantic analysis.
  */
 public class OpenDistroSqlAnalyzer {
 
+    private static final Logger LOG = LogManager.getLogger();
+
+    /** Original sql query */
```
galkk: the comment feels redundant. What about renaming field to originalSqlQuery and removing the comment?

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/StringSimilarity.java
```diff
@@ -0,0 +1,57 @@
...
+    private final StringDistance algorithm = new LevenshteinDistance();
+
+    private final Collection<String> candidates;
+
+    public StringSimilarity(Collection<String> candidates) {
+        this.candidates = Collections.unmodifiableCollection(candidates);
+    }
+
+    /**
+     * Calculate similarity distance between target and candidate strings.
```
galkk: I think that everything here is actively deceiving reader :(

1. The javadoc says that the method calculates similartiy distance. In reality it returns string the most similar to supplied
2. The javadoc says "one or more most similar), but it will always return only one, last string that is most similar (e.g. if there will be 2 strings with the same distance, algorithm will still return only one string

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/StringSimilarity.java
```diff
@@ -0,0 +1,57 @@
...
+        for (String name : candidates) {
+            float dist = algorithm.getDistance(target, name);
+            if (dist > max) {
+                result = name;
+                max = dist;
+            }
+        }
+        return Arrays.asList(result);
+    }
+
```
galkk: consider removing that empty space here and everywhere

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
dai-chen: It seems many other code will also break. Maybe we should consider how to avoid it later.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/AggregateFunction.java
dai-chen: Agree it's not the best way. But keep it as it is and add comments for now. I'm planning to refactor it in next phase along with documentation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/StringSimilarity.java
dai-chen: done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/StringSimilarity.java
dai-chen: Good catch! Update all confusing JavaDoc. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/StringSimilarity.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Find most similar string in candidates by calculating similarity distance
+     * among target and candidate strings.
+     *
+     * @param target    string to match
+     * @return          most similar string to the target
+     */
+    public String similarTo(String target) {
+        float max = -1;
```
galkk: in this case we could replace that with one-liner

public String closestString(String target) {
    return candidates.stream().max(name->algorithm.getDistance(name, target)).get();
}

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/scope/SemanticContext.java
dai-chen: Sure. I've split `SemanticAnalyzer` into 2 sub visitors: `ESMappingLoader` and `TypeChecker`. The first one is responsible for loading field types into context by `LocalClusterState`. And the second one is focused on type checking.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/OpenDistroSqlAnalyzer.java
dai-chen: Sure!

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/StringSimilarity.java
dai-chen: Just saw your comment. Will do. Thanks!

dai-chen: Done.

Issue #166, opened 2019-08-26
pretty option doesn't work for explain request


arsen-es:
The resulting json of a request made to `_opendistro/_sql/_explain` endpoint is not properly formatted when `?pretty` query parameter is specified.

PR #203, opened 2019-10-01
Issue#166 _explain response not pretty formatted

chloe-zh:
*Issue #, if available:*
* [Issue #166 ](https://github.com/opendistro-for-elasticsearch/sql/issues/166)

*Description of changes:*
* Added a json pretty formatter, the explain responses call the formatter before sent to the response channel.
* Added IT and UT for the formatter.


src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/JsonPrettyFormatter.java
```diff
@@ -0,0 +1,39 @@
...
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.executor.format;
+
+import org.elasticsearch.common.Strings;
+import org.elasticsearch.common.xcontent.*;
```
davidcui-amzn: It's generally good practice to list the individual libraries needed and avoid using *

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.nio.charset.StandardCharsets;
+import java.sql.SQLFeatureNotSupportedException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class ExplainTest {
+
+    /*
+    executeSqlRequest(final RestRequest request, final QueryAction queryAction,
```
davidcui-amzn: Should remove this comment

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/JsonPrettyFormatter.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.io.IOException;
+
+public class JsonPrettyFormatter {
+
+    public JsonPrettyFormatter() {
+    }
+
+    public String formatter(String jsonString) throws IOException {
```
abbashus: The function name should be a verb, `format` sounds more appropriate.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/JsonPrettyFormatter.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.executor.format;
+
+import org.elasticsearch.common.Strings;
+import org.elasticsearch.common.xcontent.*;
+
+import java.io.IOException;
+
+public class JsonPrettyFormatter {
```
penghuo: It seems JsonPrettyFormatter could be utility class.

chloe-zh:
@penghuo  yes the only entrance of _explain for our own service is within com.amazon.opendistroforelasticsearch.sql.plugin.RestSqlAction.executeSqlRequest; besides, I have tested JOIN queries on es engine and it's good now.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -141,7 +142,8 @@ private void executeSqlRequest(final RestRequest request, final QueryAction quer
                                    final Client client, final RestChannel channel) throws Exception {
         if (isExplainRequest(request)) {
             final String jsonExplanation = queryAction.explain().explain();
-            channel.sendResponse(new BytesRestResponse(OK, "application/json; charset=UTF-8", jsonExplanation));
+            final String jsonPretty = (new JsonPrettyFormatter()).format(jsonExplanation);
```
dai-chen: I wonder with this change do we pretty format all the time or only when `pretty` param given in url?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/ExplainIT.java
```diff
@@ -211,4 +212,19 @@ public void testContentTypeOfExplainRequestShouldBeJson() throws IOException {
...
+    @Test
+    public void testExplainRequestPrettyFormatted() throws IOException {
+
+        String expectedOutputFilePath = TestUtils.getResourceFilePath(
+                "src/test/resources/expectedOutput/explain_format_pretty.json");
+        String expectedOutput = Files.toString(new File(expectedOutputFilePath), StandardCharsets.UTF_8);
+
+        String query = "SELECT firstname FROM " + TEST_INDEX_ACCOUNT;
+        String explain = explainQuery(query);
+        String result = (new JsonPrettyFormatter()).format(explain);
```
dai-chen: np: the parenthesis around `new JsonPrettyFormatter()` could be removed.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
chloe-zh: Actually all the time because the output would be a oneline string, which I think no need to remain. But I guess we could add to a couple of other format options with `?pretty` param in url, if this specific enhancement is needed.

dai-chen: > Actually all the time because the output would be a oneline string, which I think no need to remain. But I guess we could add to a couple of other format options with `?pretty` param in url, if this specific enhancement is needed.

Yeah, I think it would be great if we can do this only when `pretty` given which is consistent with behavior of other ES API, for example `_cluster/health`.

dai-chen: It seems ES SQL doesn't provide pretty option for translate API neither. So we're good for now.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -141,7 +142,8 @@ private void executeSqlRequest(final RestRequest request, final QueryAction quer
                                    final Client client, final RestChannel channel) throws Exception {
         if (isExplainRequest(request)) {
             final String jsonExplanation = queryAction.explain().explain();
-            channel.sendResponse(new BytesRestResponse(OK, "application/json; charset=UTF-8", jsonExplanation));
+            final String jsonPretty = new JsonPrettyFormatter().format(jsonExplanation);
```
galkk: There seem to be quite a lot of code to have formatted json. 
WIll 

    new JSONObject(jsonExplanation).toString(4) work here? It could simplify code a lot

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/JsonPrettyFormatter.java
```diff
@@ -0,0 +1,44 @@
...
+import org.elasticsearch.common.xcontent.XContentType;
+
+import java.io.IOException;
+
+public class JsonPrettyFormatter {
+
+    public JsonPrettyFormatter() {
+    }
+
+    public String format(String jsonString) throws IOException {
```
galkk: See my comment in RestSqlAction, but this method could be made static - there is no need to have instance of JsonPrettyFormatter at all

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
chloe-zh: That's what I've tried in the first place. The the outcome was not standard json format, with some missing returns. So I turned to the internal lib Xcontent.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Test
+    public void assertExplainRequestShouldBeJson() throws IOException {
+
+        String expectedOutputFilePath = TestUtils.getResourceFilePath(
+                "src/test/resources/expectedOutput/explain_format_pretty.json");
+        String expectedOutput = Files.toString(new File(expectedOutputFilePath), StandardCharsets.UTF_8)
+                .replaceAll("\r", "");
+
+        String explainResult = explain(String.format("{\"query\":\"" +
```
galkk: Couple comments here:

1. This is not integration test, so no need to use index names here from test constants here. You could just put any index name

2. Without need of using test constants you don't need string formatting

3. In case if you **need** to use string formatting, you need to use StringUtils.format that will use proper locale

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.sql.SQLFeatureNotSupportedException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class ExplainTest {
+
+    @Test
+    public void assertExplainRequestShouldBeJson() throws IOException {
+
```
galkk: consider removing those empty lines

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
```diff
@@ -0,0 +1,128 @@
...
+                "LEFT JOIN %s b2 " +
+                "ON b1.age = b2.age AND b1.state = b2.state\"}", bank, bank));
+        String result = new JsonPrettyFormatter().format(explainResult);
+        removeBlank(result);
+        replaceReturn(result);
+
+        assertThat(result, equalTo(expectedOutput));
+    }
+
+    public void removeBlank(String string) {string.replaceAll("\\s", "");}
```
galkk: string is immutable, so this method does nothing (same for replaceReturn), because replaceAll returns new string. You can delete those 2 methods and your tests expect to pass (and I feel it a bit suspicious)

src/test/resources/expectedOutput/explain_format_pretty.json
```diff
@@ -0,0 +1,10 @@
+{
```
galkk: I like that you've extracted that into files instead of hardcoding, good job.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
galkk: @chloe-zh could you provide some examples?

chloe-zh: Yes sure!
An example output of `new JSONObject(jsonExplanantion).toString(4)`:
```
{
    "size": 0,
    "stored_fields": "add_1",
    "script_fields": {"add_1": {
        "ignore_failure": false,
        "script": {
            "source": "def add_2 = doc['time_operating_in_hours'].value + 1;return add_2;",
            "lang": "painless"
        }
    }},
    "from": 0,
    "_source": {
        "excludes": [],
        "includes": ["script"]
    },
    "aggregations": {"equipment_id.keyword": {"terms": {
        "shard_min_doc_count": 0,
        "field": "equipment_id.keyword",
        "size": 200,
        "min_doc_count": 1,
        "order": [
            {"_count": "desc"},
            {"_key": "asc"}
        ]
    }}}
}
```

An example output of JsonPrettyFormatter (XContent):
```
{
  "from" : 0,
  "size" : 0,
  "_source" : {
    "includes" : [
      "script"
    ],
    "excludes" : [ ]
  },
  "stored_fields" : "add_1",
  "script_fields" : {
    "add_1" : {
      "script" : {
        "source" : "def add_2 = doc['time_operating_in_hours'].value + 1;return add_2;",
        "lang" : "painless"
      },
      "ignore_failure" : false
    }
  },
  "aggregations" : {
    "equipment_id.keyword" : {
      "terms" : {
        "field" : "equipment_id.keyword",
        "size" : 200,
        "min_doc_count" : 1,
        "shard_min_doc_count" : 0,
        "show_term_doc_count_error" : false,
        "order" : [
          {
            "_count" : "desc"
          },
          {
            "_key" : "asc"
          }
        ]
      }
    }
  }
}
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
chloe-zh: Thank you for pointing out. I'll modify the IT by today.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
chloe-zh: You are right. This method should return a String type, which returns a new string.

chloe-zh: I've made some modifications in the unittest. `\s` should be removed before replace `\n`. Then by comparing the `\n` could tell if the response still one-line or has the same number&position with expected.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
galkk: Thanks for providing an example, that definitely explains everything. Let's keep the code as is now

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/ExplainIT.java
```diff
@@ -211,4 +212,19 @@ public void testContentTypeOfExplainRequestShouldBeJson() throws IOException {
...
+
+        String expectedOutputFilePath = TestUtils.getResourceFilePath(
+                "src/test/resources/expectedOutput/explain_format_pretty.json");
+        String expectedOutput = Files.toString(new File(expectedOutputFilePath), StandardCharsets.UTF_8);
+
+        String query = "SELECT firstname FROM " + TEST_INDEX_ACCOUNT;
+        String explain = explainQuery(query);
+        String result = new JsonPrettyFormatter().format(explain);
+
+        Assert.assertThat(result.replaceAll("\\s", "").replaceAll("\\n", "*"),
```
galkk: Why we need all those replaceAll, if we read file and do translation? Why can't we just do 

    assertThat(result, isEqual(expectedOutput))?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
chloe-zh: I've removed the replaceAll. It would be enough to assert if they are equal.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
chloe-zh: I've added one more IT and rewrote the UT. Instead of testing the `_explain` request, I went more specific into testing the formatter class, including the functionality and the exception.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/ExplainIT.java
```diff
@@ -15,24 +15,31 @@
...
 import java.nio.charset.StandardCharsets;
+import java.sql.SQLFeatureNotSupportedException;
 
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_ACCOUNT;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_DOG;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_LOCATION;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_NESTED_TYPE;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_PEOPLE;
-import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_PHRASE;
+import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.*;
```
dai-chen: I guess your IDE did this automatically. Please refer to this: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports.

PR #207, opened 2019-10-04
Modified IntgTests Adapting to The Json Formatter

chloe-zh:
*Issue #, if available:*

*Description of changes:*
Modified the integration tests to make it adapt to the json formatter since the output of all the explain responses in ITs have been changed to pretty formatted.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
@@ -1127,7 +1127,8 @@ public class AggregationIT extends SQLIntegTestCase {
         String result = explainQuery(query);
 
         Assert.assertThat(result, containsString("Math.ceil(doc['timestamp'].value);"));
-        Assert.assertThat(result, containsString("\"aggregations\" : {"));
+        Assert.assertThat(result.replaceAll("\\s", ""),
+                containsString("\"script\":{\"source\""));
     }
 
     @Test
@@ -1137,7 +1138,8 @@ public class AggregationIT extends SQLIntegTestCase {
         String result = explainQuery(query);
 
         Assert.assertThat(result, containsString("Math.pow(doc['field'].value, 1)"));
-        Assert.assertThat(result, containsString("\"aggregations\" : {"));
+        Assert.assertThat(result.replaceAll("\\s", ""),
+                containsString("\"script\":{\"source\""));
     }
 
 
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
@@ -1116,7 +1116,7 @@ public class AggregationIT extends SQLIntegTestCase {
                 "'interval'='20d ', 'format'='yyyy-MM-dd') limit 1000", TEST_INDEX_ONLINE);
         String result = explainQuery(query);
 
-        Assert.assertThat(result, containsString("\"script\":{\"source\""));
+        Assert.assertThat(result, containsString("\"aggregations\" : {"));
         Assert.assertThat(result, containsString("substring(0, 1)"));
     }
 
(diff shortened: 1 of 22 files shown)


penghuo:
Could you add more detail description for the issue title. "test" is too simple.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
```diff
@@ -1116,7 +1116,7 @@ public void termsWithScript() throws Exception {
                 "'interval'='20d ', 'format'='yyyy-MM-dd') limit 1000", TEST_INDEX_ONLINE);
         String result = explainQuery(query);
 
-        Assert.assertThat(result, containsString("\"script\":{\"source\""));
+        Assert.assertThat(result, containsString("\"aggregations\" : {"));
```
penghuo: I think the json formatter only impact the output format, why the content is changed?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    public void assertExplainJoinRequestPrettyFormatted() throws IOException {
+
+        String expectedOutPutFilePath = TestUtils.getResourceFilePath(
+                "src/test/resources/expectedOutput/explain_join_format_pretty.json");
+        String expectedOutput = Files.toString(new File(expectedOutPutFilePath), StandardCharsets.UTF_8);
+        removeBlank(expectedOutput);
+        replaceReturn(expectedOutput);
```
penghuo: why these post processing is needed for String compare?

chloe-zh: No need to post process

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
chloe-zh: Yes it only changes the format. The content keeps the same.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/ExplainIT.java
```diff
@@ -187,16 +187,19 @@ public void termsIncludeExcludeExplainTest() throws IOException {
 
         String result = explainQuery(queryPrefix + " terms(field='correspond_brand_name',size='10'," +
                 "alias='correspond_brand_name',include='\\\".*sport.*\\\"',exclude='\\\"water_.*\\\"')");
-        Assert.assertThat(result, containsString(expected1));
+        Assert.assertThat(result.replaceAll("\\s", ""),
```
galkk: I'm strongly against those replacements. Why we do need them?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MethodQueryIT.java
```diff
@@ -47,8 +47,8 @@ public void queryTest() throws IOException {
         final String result = explainQuery(String.format(Locale.ROOT,
                 "select address from %s where q= query('address:880 Holmes Lane') limit 3",
                 TestsConstants.TEST_INDEX_ACCOUNT));
-        Assert.assertThat(result,
-                containsString("query_string\":{\"query\":\"address:880 Holmes Lane"));
+        Assert.assertThat(result.replaceAll("\\s", ""),
```
galkk: I recommend instead of doing substring searches, just parse result into JSONObject and get value using patch and compare it (like it is done in most of the codebase). Those removals of spaces are more than messy

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/ExplainTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+            return translate(sql, jsonRequest);
+        } catch (SqlParseException | SQLFeatureNotSupportedException e) {
+            throw new ParserException("Illegal sql expr in request: " + request);
+        }
+    }
+
+    private String translate(String sql, JSONObject jsonRequest) throws SQLFeatureNotSupportedException, SqlParseException {
+        Client mockClient = Mockito.mock(Client.class);
+        CheckScriptContents.stubMockClient(mockClient);
```
galkk: Check if this line is actually needed. In most of cases it was working without it. Also I'm sure that I saw similar code in other tests as well - consider reusing it instead of copy-pasting

Issue #121, opened 2019-07-18
Dot/period at start of index name fails to parse


aetter:
```json
POST _opendistro/_sql
{
  "query": "SELECT * FROM .kibana LIMIT 50"
}
```

This query fails with the following error:

```json
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "ERROR. token : DOT, pos : 15",
    "type": "ParserException"
  },
  "status": 400
}
```

Indices starting with `.` is perfectly valid in Elasticsearch, so unless there's a good reason, I think we should support this operation.

abbashus:
Thanks @aetter , looks like a valid case. Even wrapping `.kibana` in back-ticks and single quotes is not working. We will look into it and come with a fix.

dai-chen:
> Thanks @aetter , looks like a valid case. Even wrapping `.kibana` in back-ticks and single quotes is not working. We will look into it and come with a fix.

I remember we ran into the similar issue for identifiers in back-ticks when evaluating support for Tableau. @qreshi may have more context.

dai-chen:
New issue related: https://github.com/opendistro-for-elasticsearch/sql/issues/212

PR #216, opened 2019-10-12
Support to parse index name with dot/period at start (issue#121)

chloe-zh:
*Issue #, if available:*

* [#121 Dot/period at start of index name fails to parse](https://github.com/opendistro-for-elasticsearch/sql/issues/121)

*Description of changes:*

* Support to parse index name with dot/period at start.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
@@ -546,6 +546,14 @@ public class SqlParserTest {
     }
 
     @Test
+    public void systemIndexNameTest() throws SqlParseException {
+        String query = "SELECT * FROM .kibana";
+        SQLExpr sqlExpr = queryToExpr(query);
+        Select select = parser.parseSelect((SQLQueryExpr) sqlExpr);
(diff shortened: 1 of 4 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -545,6 +545,14 @@ public void indexWithDotsAndHyphen() throws SqlParseException {
         Assert.assertEquals("data-2015.08.22", select.getFrom().get(0).getIndex());
     }
 
+    @Test
+    public void systemIndexNameTest() throws SqlParseException {
```
davidcui-amzn: I think we could use a more descriptive test name here- something to indicate that we're testing an index with a dot character at the front of its name

chloe-zh: @davidcui-amzn I've changed it into `indexNameWithDotAtTheStart()`

dai-chen:
Your UT looks good and is making sure dot index can pass both and new parser. Could you also add a simple IT to see if other parts of our codebase works too? For example, index a document to create a dot index, then run a query with simple SELECT-FROM-WHERE and assert its result.

PR #246, opened 2019-10-23
Support to parse backticks quoted identifiers for larger coverage

chloe-zh:
*Issue #, if available:*

*Description of changes:*

* Deleted the visit SQLPropertyExpr method from the unquote identifier rewriter rule, and modified the ElasticSqlExprParser to support a larger coverage of parsing quoted identifier cases.


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -1686,6 +1686,16 @@ public void backticksQuotedIndexNameTest() throws Exception {
                 executeQuery("SELECT lastname FROM bank", "jdbc"),
                 executeQuery("SELECT `bank`.`lastname` FROM `bank`", "jdbc")
         );
+
+        assertEquals(
+                executeQuery(
+                        "SELECT `b`.`age` AS `hello world`, AVG(`b`.`balance`) FROM `bank` AS `b` " +
+                                "WHERE ABS(`b`.`age`) > 20 GROUP BY `b`.`gender` ORDER BY `b`.`age`",
+                "jdbc"),
+                executeQuery("SELECT age , AVG(balance) FROM bank " +
```
dai-chen: Because field `age` doesn't have the alias (hello world), could you confirm if the assertion works as expected?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/ElasticSqlExprParser.java
```diff
@@ -622,6 +624,14 @@ public final SQLExpr primaryRest(SQLExpr expr) {
                     + lexer.pos());
         }
 
+        if (lexer.token() == Token.DOT && expr instanceof SQLIdentifierExpr) {
```
dai-chen: Please add a comment here to explain the reason behind, ex. Druid parser creates SQLPropertyExpr (owner=xxx, name=xxx) later (dotRest?) for SQLIdentifier followed by DOT which causes xxx problem ...

chloe-zh: @dai-chen Ok will do. Thanks!

PR #247, opened 2019-10-23
Support to parse backticks quoted identifiers for larger coverage

chloe-zh:
*Issue #, if available:*

*Description of changes:*

* Deleted the visit SQLPropertyExpr rewritten method from UnquoteIdentifierRule, and modified the ElasticSqlExprParser to support a larger coverage of parsing quoted identifier cases.

* Added UT and IT for these complex cases.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/identifier/UnquoteIdentifierRuleTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/identifier/UnquoteIdentifierRuleTest.java
@@ -61,6 +61,17 @@ public class UnquoteIdentifierRuleTest {
         ).shouldBeAfterRewrite("SELECT bank.lastname FROM bank");
     }
 
+    @Test
+    public void queryWithQuotedAggrAndFunc() {
+        query("" +
+                "SELECT `b`.`lastname` AS `name`, AVG(`b`.`balance`) FROM `bank` AS `b` " +
+                "WHERE ABS(`b`.`age`) > 20 GROUP BY `b`.`gender` ORDER BY `b`.`age`"
+        ).shouldBeAfterRewrite(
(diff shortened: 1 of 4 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/identifier/UnquoteIdentifierRuleTest.java
```diff
@@ -61,6 +61,17 @@ public void selectSpecificFieldsUsingQuotedTableNamePrefix() {
         ).shouldBeAfterRewrite("SELECT bank.lastname FROM bank");
     }
 
+    @Test
+    public void queryWithQuotedAggrAndFunc() {
+        query("" +
+                "SELECT `b`.`lastname` AS `name`, AVG(`b`.`balance`) FROM `bank` AS `b` " +
+                "WHERE ABS(`b`.`age`) > 20 GROUP BY `b`.`gender` ORDER BY `b`.`age`"
```
dai-chen: Please correct both queries by using `` `b`.`lastname` `` in `GROUP BY` and `ORDER BY`. In case this would fail if we add this check in semantic analyzer in future.

chloe-zh: It's done now, thanks! @dai-chen @abbashus

abbashus: I still see a problem here, looks like it not fixed. It should be `lastname` in `GROUP BY` and `ORDER B` instead of `age`, because as per SQL standard we cannot use a non-group by field in `SELECT`

chloe-zh: @abbashus You are right. fix now. thanks!

Issue #215, opened 2019-10-11
Support ordinal in GROUP/ORDER BY clause


dai-chen:
In issue https://github.com/opendistro-for-elasticsearch/sql/issues/213, it seems BI tool like Tableau will generates `GROUP BY` (and probably `ORDER BY` too) using ordinal.

Take query `SELECT age FROM accounts GROUP BY 1` for example, the ordinal 1 here is referring to field `age`. So actually this query is Equivalent to ` SELECT age FROM accounts GROUP BY age`. Although it's said this is not a good practice, we need to support this for integration with BI tool.

Reference: 
1. https://stackoverflow.com/questions/7392730/what-does-sql-clause-group-by-1-mean
2. https://dba.stackexchange.com/questions/86609/why-do-we-use-group-by-1-and-group-by-1-2-3-in-sql-query

abbashus:
Fixed by #248.

PR #248, opened 2019-10-23
Support ordinal aliases in GROUP and ORDER BY clauses

abbashus:
*Issue #215*

*Description of changes:*
- Added a new Rewriter to map GROUP BY and ORDER By aliases to SELECT column fields

*Testing*
- Passing Gradle build with existing unit and integration tests
- Added new unit tests and integration tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
@@ -26,9 +26,7 @@ import com.alibaba.druid.sql.ast.statement.SQLSelectQuery;
 import com.alibaba.druid.sql.dialect.mysql.ast.expr.MySqlSelectGroupByExpr;
 import com.alibaba.druid.sql.dialect.mysql.ast.statement.MySqlSelectQueryBlock;
 import com.alibaba.druid.sql.dialect.mysql.visitor.MySqlASTVisitorAdapter;
-import com.alibaba.druid.sql.parser.ParserException;
 import com.alibaba.druid.sql.parser.SQLExprParser;
-import com.alibaba.druid.sql.parser.Token;
 import com.amazon.opendistroforelasticsearch.sql.parser.ElasticSqlExprParser;
 import com.amazon.opendistroforelasticsearch.sql.rewriter.RewriteRule;
 import com.amazon.opendistroforelasticsearch.sql.rewriter.RewriteRuleExecutor;
@@ -57,7 +55,7 @@ public class OrdinalRewriterRule implements RewriteRule<SQLQueryExpr> {
     }
 
     @Override
-    public boolean match(SQLQueryExpr root) throws SQLFeatureNotSupportedException {
+    public boolean match(SQLQueryExpr root) {
         SQLSelectQuery sqlSelectQuery = root.getSubQuery().getQuery();
          if (!(sqlSelectQuery instanceof MySqlSelectQueryBlock)) {
              // it could be SQLUnionQuery
@@ -70,12 +68,18 @@ public class OrdinalRewriterRule implements RewriteRule<SQLQueryExpr> {
             return false;
         }
 
-        // we cannot clone SQLSelectItem, so we need similar objects to assign to GroupBy and OrderBy items
-        sqlExprGroupCopy = generateAST();
-        sqlExprOrderCopy = generateAST();
+        return true;
+    }
 
-        collectOrdinalAliases(sqlExprGroupCopy, groupOrdinalToExpr);
-        collectOrdinalAliases(sqlExprOrderCopy, orderOrdinalToExpr);
+    @Override
+    public void rewrite(SQLQueryExpr root) throws SQLFeatureNotSupportedException {
+
+        // we cannot clone SQLSelectItem, so we need similar objects to assign to GroupBy and OrderBy items
+        sqlExprGroupCopy = toSqlExpr();
+        sqlExprOrderCopy = toSqlExpr();
+
+        mapOrdinalIndexesToSelectItems(sqlExprGroupCopy, groupOrdinalToExpr);
+        mapOrdinalIndexesToSelectItems(sqlExprOrderCopy, orderOrdinalToExpr);
 
         if (groupOrdinalToExpr.isEmpty()) {
             // orderOrdinalToExpr and groupOrdinalToExpr should be identical
@@ -91,53 +95,9 @@ public class OrdinalRewriterRule implements RewriteRule<SQLQueryExpr> {
             .build();
 
         ruleExecutor.executeOn(sqlExprGroupCopy);
-        ruleExecutor.executeOn(sqlExprGroupCopy);
-        return true;
-    }
+        ruleExecutor.executeOn(sqlExprOrderCopy);
 
-    @Override
-    public void rewrite(SQLQueryExpr root) {
         changeOrdinalAliasInGroupAndOrderBy(root);
-    }
-
-    private boolean hasGroupByWithOrdinals(MySqlSelectQueryBlock query) {
-        if (query.getGroupBy() == null) {
-            return false;
-        } else if (query.getGroupBy().getItems().isEmpty()){
-            return false;
-        }
-
-        return query.getGroupBy().getItems().stream().anyMatch(x ->
-            x instanceof MySqlSelectGroupByExpr && ((MySqlSelectGroupByExpr) x).getExpr() instanceof SQLIntegerExpr
-        );
-    }
-
-    private boolean hasOrderByWithOrdinals(MySqlSelectQueryBlock query) {
-        if (query.getOrderBy() == null) {
-            return false;
-        } else if (query.getOrderBy().getItems().isEmpty()){
-            return false;
-        }
-
-        return query.getOrderBy().getItems().stream().anyMatch(x ->
-            x.getExpr() instanceof SQLIntegerExpr
-            || (
-                x.getExpr() instanceof SQLBinaryOpExpr
-                && ((SQLBinaryOpExpr) x.getExpr()).getLeft() instanceof SQLIntegerExpr
-            )
-        );
-    }
-
-    private void collectOrdinalAliases(SQLQueryExpr root, Map<Integer, SQLExpr> map) {
-        root.accept(new MySqlASTVisitorAdapter() {
-            private Integer count = 0;
-
-            @Override
-            public boolean visit(SQLSelectItem select) {
-                map.put(++count, select.getExpr());
-                return false;
-            }
-        });
     }
 
     private void changeOrdinalAliasInGroupAndOrderBy(SQLQueryExpr root) {
@@ -170,8 +130,8 @@ public class OrdinalRewriterRule implements RewriteRule<SQLQueryExpr> {
                     checkInvalidOrdinal(newExpr, ordinalValue, orderException);
                     orderByItem.setExpr(newExpr);
                     newExpr.setParent(orderByItem);
-                }else if (expr instanceof SQLBinaryOpExpr
-                          && ((SQLBinaryOpExpr) expr).getLeft() instanceof SQLIntegerExpr) {
+                } else if (expr instanceof SQLBinaryOpExpr
+                    && ((SQLBinaryOpExpr) expr).getLeft() instanceof SQLIntegerExpr) {
                     // support ORDER BY IS NULL/NOT NULL
                     SQLBinaryOpExpr binaryOpExpr = (SQLBinaryOpExpr) expr;
                     SQLIntegerExpr integerExpr = (SQLIntegerExpr) binaryOpExpr.getLeft();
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        // we need to rewrite the generated syntax tree to match the original syntax tree Select clause
+        ruleExecutor = RewriteRuleExecutor.builder()
+            .withRule(new UnquoteIdentifierRule())
+            .withRule(new TableAliasPrefixRemoveRule())
+            .build();
+
+        ruleExecutor.executeOn(sqlExprGroupCopy);
+        ruleExecutor.executeOn(sqlExprGroupCopy);
```
galkk: this looks suspicious (the same execution twice)

abbashus: Good catch , the later one should be `sqlExprOrderCopy`.

galkk:
I like this approach. In theory, it could be used to inline/fix aliases as well, in order by/group by expressions, instead of hackier approach that we have now (something that I've tried to do in https://github.com/opendistro-for-elasticsearch/sql/pull/168/files, that I needed to revert later)

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+
+        if (groupOrdinalToExpr.isEmpty()) {
+            // orderOrdinalToExpr and groupOrdinalToExpr should be identical
+            // because they are generated from similar syntax trees
+            // checking for groupOrdinalToExpr or orderOrdinalToExpr is enough
+            return false;
+        }
+
+        // we need to rewrite the generated syntax tree to match the original syntax tree Select clause
+        ruleExecutor = RewriteRuleExecutor.builder()
```
galkk: I think that this should be moved to rewrite, as it already mutates the expression.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+    public boolean match(SQLQueryExpr root) throws SQLFeatureNotSupportedException {
+        SQLSelectQuery sqlSelectQuery = root.getSubQuery().getQuery();
+         if (!(sqlSelectQuery instanceof MySqlSelectQueryBlock)) {
+             // it could be SQLUnionQuery
+             return false;
+         }
+
+        MySqlSelectQueryBlock query = (MySqlSelectQueryBlock) sqlSelectQuery;
+
+        if (!hasGroupByWithOrdinals(query) && !hasOrderByWithOrdinals(query)) {
```
galkk: I don't understand why we've supposedly have the same check twice, could you add more comments explaining what's going on? From the code it looks like this checks if we have ordinals, and the code below does the same, but in different way. Could only one check work?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+            .build();
+
+        ruleExecutor.executeOn(sqlExprGroupCopy);
+        ruleExecutor.executeOn(sqlExprGroupCopy);
+        return true;
+    }
+
+    @Override
+    public void rewrite(SQLQueryExpr root) {
+        changeOrdinalAliasInGroupAndOrderBy(root);
```
galkk: that's matter of taste and personal preference, but I'd recommend to move method changeOrdinalAliasInGroupAndOrderBy after that method, or even inline it

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+            return false;
+        } else if (query.getOrderBy().getItems().isEmpty()){
+            return false;
+        }
+
+        return query.getOrderBy().getItems().stream().anyMatch(x ->
+            x.getExpr() instanceof SQLIntegerExpr
+            || (
+                x.getExpr() instanceof SQLBinaryOpExpr
+                && ((SQLBinaryOpExpr) x.getExpr()).getLeft() instanceof SQLIntegerExpr
```
galkk: Could you comment here which type of ORDER BY could be matched by that expression?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+            )
+        );
+    }
+
+    private void collectOrdinalAliases(SQLQueryExpr root, Map<Integer, SQLExpr> map) {
```
galkk: I feel that the name is misleading and something like

    mapSelectItemsToOrdinalIndexes 

or something along those lines will reflect meaning much better.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: `ORDER BY a IS NULL/NOT NULL` , will add a comment.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Sure, will change that, I think it should be other way around to `mapOrdinalIndexesToSelectItems`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+
+            @Override
+            public boolean visit(SQLSelectOrderByItem orderByItem) {
+                SQLExpr expr = orderByItem.getExpr();
+                Integer ordinalValue;
+
+                if (expr instanceof SQLIntegerExpr) {
+                    ordinalValue = ((SQLIntegerExpr) expr).getNumber().intValue();
+                    SQLExpr newExpr = orderOrdinalToExpr.get(ordinalValue);
+                    checkInvalidOrdinal(newExpr, ordinalValue, orderException);
```
galkk: this is a bit repetitive, you can extract this to method like

    getExpressionByOrdinal

and throw an exception inside (will simplify code a bt)

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+                    orderByItem.setExpr(newExpr);
+                    newExpr.setParent(orderByItem);
+                }else if (expr instanceof SQLBinaryOpExpr
```
galkk: surpised that it's passed checkstyle (no space between } and if)

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/ordinal/OrdinalRewriterRuleTest.java
```diff
@@ -0,0 +1,159 @@
...
+import java.sql.SQLFeatureNotSupportedException;
+
+import static org.hamcrest.Matchers.containsString;
+
+/**
+ * Test cases for ordinal aliases in GROUP BY and ORDER BY
+ */
+
+public class OrdinalRewriterRuleTest {
+
```
galkk: I have potentially funny test case 

    SELECT lastname AS 1 from bank GROUP BY 1

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Fixed it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+            return false;
+        }
+
+        // we need to rewrite the generated syntax tree to match the original syntax tree Select clause
+        ruleExecutor = RewriteRuleExecutor.builder()
+            .withRule(new UnquoteIdentifierRule())
+            .withRule(new TableAliasPrefixRemoveRule())
+            .build();
+
+        ruleExecutor.executeOn(sqlExprGroupCopy);
```
dai-chen: I'm wondering any chance we run this rewriter before `UnquoteIdentifierRule` and `TableAliasPrefixRemoveRule`? Because in my understanding, the only thing we need to do in this rewriter is to replace ordinal by AST node. If it's possible, we don't need to run all rewrite rule again here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
dai-chen: I checked that `selectList` in Druid (`MySqlSelectQueryBlock`) is an `ArrayList` with getter. So I think this visitor along with a new `Map` is not necessary.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
```diff
@@ -0,0 +1,210 @@
...
+        return (SQLQueryExpr) toSqlExpr(sql);
+    }
+
+    private static SQLExpr toSqlExpr(String sql) {
+        SQLExprParser parser = new ElasticSqlExprParser(sql);
+        SQLExpr expr = parser.expr();
+
+        if (parser.getLexer().token() != Token.EOF) {
+            throw new ParserException("Illegal SQL expression : " + sql);
+        }
```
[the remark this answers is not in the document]
abbashus: Sure, will remove this.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Sure, will check if that is possible, and will remove it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Moved `changeOrdinalAliasInGroupAndOrderBy ` after `rewrite()` method for better readability. Not inlining.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Refactored it into `checkAndGet` method.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
dai-chen: Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/ordinal/OrdinalRewriterRule.java
abbashus: Added some comments to help understand the code better. Removed the redundant check.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/ordinal/OrdinalRewriterRuleTest.java
```diff
@@ -0,0 +1,159 @@
...
+        void shouldMatchRule() throws SQLFeatureNotSupportedException {
+            Assert.assertTrue(match());
+        }
+
+        void shouldNotMatchRule() throws SQLFeatureNotSupportedException {
+            Assert.assertFalse(match());
+        }
+
+        void shouldThrowException(int ordinal) throws SQLFeatureNotSupportedException {
```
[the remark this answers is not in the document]
abbashus: Added unit tests to check for invalid ordinals.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/ordinal/OrdinalRewriterRuleTest.java
abbashus: Checked the MySQL standard, integers cannot be valid field alias.

Issue #233, opened 2019-10-18
Function names are case-sensitive


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT SUBSTRING(OriginWeather, 1, 2)  FROM kibana_sample_data_flights

**Result:**
```
Internal Server Error. Reason: There was internal problem at backend. Type: UnsupportedOperationException. Details: The following method is not supported in Schema: SUBSTRING
	at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.executeQueryRequest(StatementImpl.java:78)
	at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.executeQueryX(StatementImpl.java:61)
	at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.execute(StatementImpl.java:170)
```

**Note:**
`SELECT substring(OriginWeather, 1, 2)...` returns results as intended.

**Expectation:**
Tableau will use UPPERCASE function names in almost all cases. Casing should not matter for function names in a query.

dai-chen:
Actually both our old and new parser are case-insensitive to function name. This issue is similar as https://github.com/opendistro-for-elasticsearch/sql/issues/224 and thrown from our JDBC formatter.

dai-chen:
Just checked our formatter seems match function name in case sensitive way (simply look it up in hash set).

galkk:
related to #176

PR #249, opened 2019-10-23
Fix the issue that function name is case sensitive in JDBC formatter

chloe-zh:
*Issue #, if available:*

* **[Issue #233 Function names are case-sensitive](https://github.com/opendistro-for-elasticsearch/sql/issues/233)**

*Description of changes:*

* Fixed the bug that the function names are case sensitive when formatting the response to JDBC.
* Added IT.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
@@ -127,9 +127,13 @@ public class JdbcTestIT extends SQLIntegTestCase {
     public void dateFunctionNameCaseInsensitiveTest() {
         assertEquals(
                 executeQuery("SELECT insert_time FROM elasticsearch-sql_test_index_online/online " +
-                        "WHERE date_FORMAT(insert_time, 'yyyy-MM-dd', 'UTC') > '2014-01-01'", "jdbc"),
+                        "WHERE date_FORMAT(insert_time, 'yyyy-MM-dd', 'UTC') > '2014-01-01' " +
+                        "group by DAte_format(insert_time, 'yyyy-MM-dd', 'UTC') " +
+                        "oRDEr By date_forMAT(insert_time, 'yyyy-MM-dd', 'UTC')", "jdbc"),
                 executeQuery("SELECT insert_time FROM elasticsearch-sql_test_index_online/online " +
-                        "WHERE date_format(insert_time, 'yyyy-MM-dd', 'UTC') > '2014-01-01'", "jdbc")
+                        "WHERE date_format(insert_time, 'yyyy-MM-dd', 'UTC') > '2014-01-01' " +
+                        "GROUP BY date_format(insert_time, 'yyyy-MM-dd', 'UTC') " +
+                        "ORDER BY date_format(insert_time, 'yyyy-MM-dd', 'UTC')", "jdbc")
         );
     }
 }
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
@@ -114,16 +114,6 @@ public class JdbcTestIT extends SQLIntegTestCase {
     }
 
     @Test
-    public void binaryOperatorNameCaseInsensitiveTest() {
-        assertEquals(
-                executeQuery("SELECT lastname FROM elasticsearch-sql_test_index_account/account " +
-                        "WHERE (age IS NOT NULL) AND (gender = 'M') ORDER BY age LIMIT 5", "jdbc"),
-                executeQuery("SELECT lastname FROM elasticsearch-sql_test_index_account/account " +
-                        "WHERE (age IS NOT NULL) and (gender = 'M') ORDER by age LIMIT 5", "jdbc")
-        );
-    }
-
-    @Test
     public void dateFunctionNameCaseInsensitiveTest() {
         assertEquals(
                 executeQuery("SELECT insert_time FROM elasticsearch-sql_test_index_online/online " +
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryFunctionsIT.java
```diff
@@ -62,6 +62,7 @@ protected void init() throws Exception {
         loadIndex(Index.ACCOUNT);
         loadIndex(Index.NESTED);
         loadIndex(Index.PHRASE);
+        loadIndex(Index.ONLINE);
```
dai-chen: Any special reason to load this index? Because I don't see code changes on test cases in this IT.

chloe-zh: @dai-chen Sorry about the mistake. I put the ITs into QueryFunctionsIT at the first place, and moved them to JdbcTestIT but forgot to delete the loadIndex. Will update it, thanks!

dai-chen: No worries. Thanks! 👍

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
```diff
@@ -78,4 +80,56 @@ public void testGroupByInQuery() {
...
+                executeQuery("SELECT lastname FROM elasticsearch-sql_test_index_account/account " +
+                        "WHERE (age IS NOT NULL) and (gender = 'M') ORDER by age LIMIT 5", "jdbc")
+        );
+    }
+
+    @Test
+    public void dateFunctionNameCaseInsensitiveTest() {
+        assertEquals(
+                executeQuery("SELECT insert_time FROM elasticsearch-sql_test_index_online/online " +
+                        "WHERE date_FORMAT(insert_time, 'yyyy-MM-dd', 'UTC') > '2014-01-01'", "jdbc"),
```
dai-chen: In your issue #233, I noticed that Andy also mentioned https://github.com/opendistro-for-elasticsearch/sql/issues/176. Could you change this test case by using `date_format` with different cases in `GROUP BY` and `ORDER BY` too? So we can make sure we support case insensitive function name present anywhere. Reference: https://github.com/opendistro-for-elasticsearch/sql/blob/2ef99b3e0a822aea498525a39992ec512eb25636/src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/DateFormatTest.java#L99

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
```diff
@@ -78,4 +80,56 @@ public void testGroupByInQuery() {
...
+                        "SELECT SubStrinG(lastname, 0, 2) FROM elasticsearch-sql_test_index_account/account " +
+                                "ORDER BY age LIMIT 5", "jdbc"),
+                executeQuery(
+                        "SELECT substring(lastname, 0, 2) FROM elasticsearch-sql_test_index_account/account " +
+                                "ORDER BY age LIMIT 5", "jdbc")
+        );
+    }
+
+    @Test
+    public void binaryOperatorNameCaseInsensitiveTest() {
```
abbashus: I don't think `AND` binary operator is affected by this change. If that is true, this test is redundant.

chloe-zh: @abbashus Thanks for pointing out! 

Actually amaong all types of functions: 
number operators, trigonometric functions, string operators cannot go through the jdbc formatter;
binary operators and date functions can go through the jdbc formatter.

But I reckoned it would be good to test all five types of functions here over the case sensitivity, since no tests have ever tested even the binary operators and date functions. 
But was not quite sure about it, so do you think it would be better not to test them here?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
chloe-zh: @dai-chen yes sure. will do it, thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
abbashus: Rest what you tested accept `AND` are functions, but not `AND`. The parser is already case-insensitive to it. If `AND` actually would have been a function like `AND(a, b)`, it would have made sense. So, I personally feel it is not needed.

chloe-zh: @abbashus  You are right, I went back to the code and found the `AND`-like "binary operator" is no need to test since it has nothing with the function name case sensitivity that I was to fix. It was not "function" actually. I have deleted this block of code, thanks!

chloe-zh: @abbashus You are right, I went back to the code and found the AND-like "binary operator" is no need to test since it has nothing with the function name case sensitivity that I was to fix. It was not "function" actually. I have deleted this block of code, thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
```diff
@@ -78,4 +80,50 @@ public void testGroupByInQuery() {
...
+                executeQuery(
+                        "SELECT substring(lastname, 0, 2) FROM elasticsearch-sql_test_index_account/account " +
+                                "ORDER BY age LIMIT 5", "jdbc")
+        );
+    }
+
+    @Test
+    public void dateFunctionNameCaseInsensitiveTest() {
+        assertEquals(
+                executeQuery("SELECT insert_time FROM elasticsearch-sql_test_index_online/online " +
```
dai-chen: `insert_time` should be changed to what is present in `GROUP BY` (`date_format(...)`)

chloe-zh: Field name inconsistent with the aggregation again. Sorry and will fix it. Thanks!

Issue #225, opened 2019-10-18
Support for CAST() function


rupal-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
Dataset: Flights sample data (loaded through Kibana)
SQL Plugin version: 1.1.0.1

Query: ```SELECT cast(dayOfWeek AS FLOAT) AS calculation FROM kibana_sample_data_flights```

Result: No data
```
{[\n]"
  "error": {[\n]"
    "reason": "Invalid SQL query",[\n]"
    "details": "Failed to parse query due to offending symbol [AS] at: 'SELECT cast(dayOfWeek AS' <--- HERE... More details: Expecting tokens in {'SELECT', '('}",[\n]"
    "type": "SyntaxAnalysisException"[\n]"
  },[\n]"
  "status": 400[\n]"
}
```

Expected Behaviour: Should return all values from column `dayOfWeek` as float.

dai-chen:
We removed `CAST` syntax in new parser grammar because we thought we don't support it. Actually there are UTs in codebase however UT won't run with new parser. Need to add it back.

dai-chen:
Apart from the changes in grammar, need to add ITs to cover this case.

dai-chen:
CAST syntax removed accidentally: https://github.com/antlr/grammars-v4/blob/master/mysql/Positive-Technologies/MySqlParser.g4#L2133

PR #253, opened 2019-10-25
Support Cast() function

davidcui-amzn:
*Issue #, if available:*
#225 
*Description of changes:*
Supports the cast function to cast a field to a different type (DateTime, Float, Long, String, Int, Double). 


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +160,44 @@ public void caseChangeWithAggregationTest() throws IOException {
                 hitAny("/aggregations/UPPER_1/buckets", kvString("/key", equalTo("AMBER"))));
     }
 
+    @Test
+    public void castIntFieldToDoubleWithoutAliasTest() throws IOException {
+        String query = "SELECT CAST(age AS DOUBLE) FROM " + TestsConstants.TEST_INDEX_ACCOUNT + " /account limit 1";
```
dai-chen: Could you add similar test for `FLOAT` which was reported in issue?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
```diff
@@ -152,6 +152,9 @@ public void setFields(List<Field> fields) throws SqlParseException {
                     MethodField method = (MethodField) field;
                     if (method.getName().toLowerCase().equals("script")) {
                         handleScriptField(method);
+                        if (method.getExpression() instanceof SQLCastExpr) {
```
dai-chen: I just checked other functions and this issue (return all fields) seems common for default DSL format. Could you add IT for JDBC format to verify if it works? For other functions, JDBC format looks good.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +160,44 @@ public void caseChangeWithAggregationTest() throws IOException {
...
+    public void castIntFieldToStringWithoutAliasTest() throws IOException {
+        String query = "SELECT CAST(balance AS STRING) FROM " + TestsConstants.TEST_INDEX_ACCOUNT + " /account limit 1";
+
+        final SearchHit[] hits = query(query).getHits();
+        assertTrue(hits[0].getFields().containsKey("cast_balance"));
+        assertTrue(hits[0].getFields().get("cast_balance").getValue() instanceof String);
+    }
+
+    @Test
+    public void castIntFieldToStringWithAliasTest() throws IOException {
```
abbashus: Append `FROM` into first select string. Here and other new test cases. Similar to `castIntFieldToStringWithoutAliasTest ` above.

abbashus: Then I believe checkstyle should fail for `castIntFieldToStringWithoutAliasTest()` as well.

davidcui-amzn: that String statement is actually at the maximum allowed length but any more and it would fail

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -614,4 +614,24 @@ private String upper(String property, String culture) {
                         "The following method is not supported in Schema: %s",
                         functionName));
     }
+
+    public static Schema.Type getCastFunctionReturnType(String castType) {
+        if (castType.toUpperCase().equals("FLOAT")) {
```
dai-chen: String switch has been available since Java 7: https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +160,74 @@ public void caseChangeWithAggregationTest() throws IOException {
                 hitAny("/aggregations/UPPER_1/buckets", kvString("/key", equalTo("AMBER"))));
     }
 
+    @Test
```
galkk: Could you add the typical tests that broke the code in past? 

1. Aggregation

SELECT CAST(age as INT) intage FROM index GROUP BY intage

2. ORDER BY 
SELECT CAST(age as INT) intage FROM index ORDER BY intage

3. something crazier, like nested casts

SELECT cast(CAST (1 AS varchar) as float)

(that one could be ignored, because I don't expect it to work, but at least we will see how the code behaves).

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
abbashus: Sorry if it was not clear, I meant making it `SELECT CAST(balance AS STRING) AS cast_string_alias FROM ` instead of `"SELECT CAST(balance AS STRING) AS cast_string_alias"  + "FROM"` and keep the `TestsConstants.TEST_INDEX_ACCOUNT + " /account limit 1"` on next line. This way it would be consistent with other new tests added without failing checkstyle.

PS: It's a minor change , you can skip if you like.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
davidcui-amzn: Good suggestion- definitely will add

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +161,113 @@ public void caseChangeWithAggregationTest() throws IOException {
...
+
+        final SearchHit[] hits = query(query).getHits();
+        assertTrue(hits[0].getFields().containsKey("cast_age"));
+        assertTrue(hits[0].getFields().get("cast_age").getValue() instanceof Double);
+    }
+
+    @Test
+    public void castIntFieldToDoubleWithAliasTest() throws IOException {
+        String query = "SELECT CAST(age AS DOUBLE) AS test_alias " +
+                       "FROM " + TestsConstants.TEST_INDEX_ACCOUNT + " /account limit 1";
```
galkk: if I recall correctly, that /account is old syntax, and you don't need it for test to work (just index name is enough).

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +161,113 @@ public void caseChangeWithAggregationTest() throws IOException {
...
+
+        final SearchHit[] hits = query(query).getHits();
```
galkk: inconsistency - you have final here, and don't have it line above. Feel free to remove unnecessary finals

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +161,113 @@ public void caseChangeWithAggregationTest() throws IOException {
...
+        assertTrue(hits[0].getFields().containsKey("test_alias"));
```
galkk: seems like http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/collection/IsMapContaining.html will help here - there are matchers hasKey  / hasEntry.
assertTrue/false should be last resort in general

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +161,113 @@ public void caseChangeWithAggregationTest() throws IOException {
...
+        assertEquals(response.getJSONArray("schema").get(0).toString(), float_type_cast);
+    }
+
+    @Test
+    public void castIntFieldToDoubleWithoutAliasOrderByTest() throws IOException {
+        String query = "SELECT CAST(age AS DOUBLE) FROM " + TestsConstants.TEST_INDEX_ACCOUNT + " /account " +
+                       "ORDER BY age limit 1";
+
+        final SearchHit[] hits = query(query).getHits();
+        assertTrue(hits[0].getFields().containsKey("cast_age"));
```
galkk: look into usage of kvInt / kvString in other tests

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -159,6 +164,100 @@ public void caseChangeWithAggregationTest() throws IOException {
...
+        JSONObject response = executeJdbcRequest(
+                "SELECT CAST(balance AS FLOAT) AS jdbc_float_alias " +
+                        "FROM " + TestsConstants.TEST_INDEX_ACCOUNT + " GROUP BY jdbc_float_alias LIMIT 5");
+
+        String float_type_cast = "{\"name\":\"jdbc_float_alias\",\"type\":\"float\"}";
+        assertEquals(response.getJSONArray("schema").get(0).toString(), float_type_cast);
+    }
+
+    @Test
+    public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
```
galkk: I don't see any result value checks. Could you add them?

For example, cast some float to int and check that the output has correct value (do we round, remove digits after decimal point)?

What will happen if I will try cast age as datetime? string as int?

davidcui-amzn: Yes- will add

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -614,4 +614,25 @@ private String upper(String property, String culture) {
                         "The following method is not supported in Schema: %s",
                         functionName));
     }
+
+    public static Schema.Type getCastFunctionReturnType(String castType) {
+        switch (castType) {
```
dai-chen: Should we do `StringUtils.toUpperCase`? Please check if the data type name is case sensitive or not in SQL.

davidcui-amzn: Good catch- just tested and need to add, will do that now.

Issue #254, opened 2019-10-25
Support SELECT <literal>


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset**: Flights sample data (loaded through Kibana)
**SQL Plugin version**: 1.1.0.2
**Tableau version**: 2019.3

**Query**: ``SELECT 1 as `Number of Records`, AvgTicketPrice FROM kibana_sample_data_flights``

**Result**: 
```
[2019-10-25T18:59:17,978][ERROR][c.a.o.s.e.f.PrettyFormatRestExecutor] [odfe-node1] Error happened in pretty formatter
java.lang.UnsupportedOperationException: The following method is not supported in Schema: assign
	at com.amazon.opendistroforelasticsearch.sql.utils.SQLFunctions.getScriptFunctionReturnType(SQLFunctions.java:615) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.SelectResultSet.fetchMethodReturnType(SelectResultSet.java:319) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.SelectResultSet.populateColumns(SelectResultSet.java:374) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.SelectResultSet.loadFromEsState(SelectResultSet.java:145) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.SelectResultSet.<init>(SelectResultSet.java:85) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.loadResultSet(Protocol.java:61) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.<init>(Protocol.java:48) ~[opendistro_sql-1.1.1.0.jar:1.1.1.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.PrettyFormatRestExecutor.execute(PrettyFormatRestExecutor.java:70) [opendistro_sql-1.1.1.0.jar:1.1.1.0]
[...]

PR #255, opened 2019-10-26
Support SELECT <number_literal>

chloe-zh:
*Issue #, if available:*

* **[Issue #254 Support SELECT <literal>](https://github.com/opendistro-for-elasticsearch/sql/issues/254)** 

*Description of changes:*

* Supported number literal in SELECT field as utility function "ASSIGN" to pass the jdbc formatter.
* Added IT and UT.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
@@ -634,7 +634,8 @@ public class SQLFunctions {
         }
 
         if (mathConstants.contains(functionName) || numberOperators.contains(functionName)
-                || trigFunctions.contains(functionName) || binaryOperators.contains(functionName)) {
+                || trigFunctions.contains(functionName) || binaryOperators.contains(functionName)
+                || utilityFunctions.contains(functionName)) {
             return Schema.Type.DOUBLE;
         }
 
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryFunctionsIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryFunctionsIT.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistroforelasticsearch.sql.esintgtest;
 
+import com.amazon.opendistroforelasticsearch.sql.utils.StringUtils;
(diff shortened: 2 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryFunctionsIT.java
```diff
@@ -196,6 +197,19 @@ public void multiMatchQueryWildcardField() throws IOException {
         );
     }
 
+    @Test
+    public void numberLiteralInSelectField() {
+        assertTrue(
```
galkk: Could you add check that the result actually contains 2 and 2.1? I'm ready to approve when I'll be sure that it works

chloe-zh: @galkk Will do, thanks!

chloe-zh: Done now, thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -634,7 +634,8 @@ private String upper(String property, String culture) {
         }
 
         if (mathConstants.contains(functionName) || numberOperators.contains(functionName)
-                || trigFunctions.contains(functionName) || binaryOperators.contains(functionName)) {
+                || trigFunctions.contains(functionName) || binaryOperators.contains(functionName)
+                || utilityFunctions.contains(functionName)) {
```
dai-chen: a minor concern: there is another `field` function (and probably more other functions added in future) in `utilityFunctions`. We may need to have a mapping for result type of specific function later.

chloe-zh: @dai-chen  Yeah, it is. So do you think we need to leave out the `field` and put `functionNama.equals("assign") in the condition clause instead?

dai-chen: It's okay. We can do this later because I assume you will touch this part of code soon. :)

chloe-zh: Ok, thanks!

Issue #235, opened 2019-10-18
Unable to recognize certain function names


rupal-bq:
Using Docker image amazon/opendistro-for-elasticsearch:1.1.0
Dataset: Flights sample data (loaded through Kibana)
SQL Plugin version: 1.1.0.1
Tableau version: 2019.3

Query:  `SELECT month(timestamp) AS mn_timestamp_ok FROM kibana_sample_data_flights`

Result: No data
```
{[\n]"
  "error": {[\n]"
    "reason": "Invalid SQL query",[\n]"
    "details": "Function [MONTH] cannot be found or used here. Did you mean [DAY_OF_MONTH]?",[\n]"
    "type": "SemanticAnalysisException"[\n]"
  },[\n]"
  "status": 400[\n]"
}
```
Expected Behaviour: Should return respective month for the values in column `timestamp`

List of functions which have similar issue:

Date Functions
- MONTH
- DAYOFMONTH
- DATE
- MONTHNAME
- TIMESTAMP
- MAKETIME
- NOW
- CURDATE

Logical Function
- IF

dai-chen:
Checked the code that these functions are missing or have different name. Here are the list of data functions we support for now:

```
            "date_format", "year", "month_of_year", "week_of_year", "day_of_year", "day_of_month",
            "day_of_week", "hour_of_day", "minute_of_day", "minute_of_hour", "second_of_minute"
```

Issue #226, opened 2019-10-18
SUBSTRING() function is 0-indexed


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT substring(OriginWeather, 1, 4) FROM kibana_sample_data_flights

**Results:**
```
| substring_1 | 
| unn | 
| lea | 
| ain | 
| hun | 
| ama | 
...
```

**Expectation:**
The JDBC standard defines the SUBSTRING() function as being 1-indexed (ie. the first character in the string is index 0), and the upper bound as inclusive, so the results should be as shown below:
```
| substring_1 | 
| Sunn | 
| Clea | 
| Rain | 
| Thun | 
| Dama | 
...
```

dai-chen:
Just did a quick look into x-pack. Interestingly, it supports both 0 and 1-indexed. Will check SQL and JDBC standard later.

chloe-zh:
Besides, the function format should be `substring(originalString, startIndex, substringLength)`, but our `substring` was implemented to be `substring(originalString, startIndex, endIndex)`. 

Will check the index rule to be 0-index or 1-index, and fix the problem mentioned above in the latest PR I'm going to submit as well.

chloe-zh:
I've check this case in es x-pack, it should be 1-index. When either `startIndex` or `endIndex` is out of bound, they are corrected to `startIndex = 1` or `endIndex = stringLength`, which is translated to `start_index = 0` and `end_index = string_legnth - 1` in painless language script.

chloe-zh:
> I've check this case in es x-pack, it should be 1-index. When either `startIndex` or `endIndex` is out of bound, they are corrected to `startIndex = 1` or `endIndex = stringLength`, which is translated to `start_index = 0` and `end_index = string_legnth - 1` in painless language script.

One more thing to mention, our semantic analyzer throws exception when the params like index, length etc. are set to negative.

Issue #224, opened 2019-10-18
Support for extra functions required by BI tools 


rupal-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
Dataset: Flights sample data (loaded through Kibana)
SQL Plugin version: 1.1.0.1
Tableau version: 2019.3

Tableau generated SQL (MySQL dialect):  ```SELECT POWER(`kibana_sample_data_flights`.`dayOfWeek`, 2) AS `sum_Calculation_160722252357632000_ok`\nFROM `kibana_sample_data_flights`\nGROUP BY 1.1000000000000001```

Generated SQL without quotes (#212) and Group By (#215): `SELECT power(dayOfWeek, 2) AS sum_Calculation_160722252357632000_ok FROM kibana_sample_data_flights`

Result: No data
```
{[\n]"
  "error": {[\n]"
    "reason": "There was internal problem at backend",[\n]"
    "details": "The following method is not supported in Schema: power",[\n]"
    "type": "UnsupportedOperationException"[\n]"
  },[\n]"
  "status": 500[\n]"
}
```

Expected Behaviour: Should return square of all values present in column dayOfWeek.

List of functions which have similar issue:
Number functions

- ATAN2
- COT 
- SIGN 
- POWER

String functions

- ASCII
- RTRIM
- LTRIM
- LOCATE
- LENGTH
- REPLACE
- LEFT
- RIGHT

Logical functions

- IFNULL
- ISNULL

dai-chen:
I think this is the limitation in our JDBC formatter. Currently the formatter doesn't recognize many SQL function and map it to correct result type in JDBC spec.

dai-chen:
For these functions, first we need to change grammar to pass both new and old SQL parser. Then based on their behaviors in standard SQL or MySQL, we need to figure out what's the right DSL to generate.

dai-chen:
Another function list that we need to support: https://github.com/opendistro-for-elasticsearch/sql/issues/235.

chloe-zh:
@dai-chen I will do #235  along with this issue.

chloe-zh:
PRs will be separated and submitted respectfully according to the function/operator type.

Issue #150, opened 2019-08-13
Fix script functions: remove hardcoded parameters


galkk:
This is followup for #149, with the same rationale

The following SQL
```sql
GET /_opendistro/_sql/_explain
{
  "query": """
    SELECT name, substring(name, 0, 3) 
    FROM employess_with_missing
    WHERE substring(name, 0, 3) = 'bob'
  """
}
```

produces the following DSL

```json
{
  "from": 0,
  "size": 200,
  "query": {
    "bool": {
      "filter": [
        {
          "bool": {
            "must": [
              {
                "script": {
                  "script": {
                    "source": "def substring_1707568387 = doc['name'].value.substring(0, 3);substring_1707568387 == 'bob'",
                    "lang": "painless"
                  },
                  "boost": 1
                }
              }
            ],
            "adjust_pure_negative": true,
            "boost": 1
          }
        }
      ],
      "adjust_pure_negative": true,
      "boost": 1
    }
  },
  "_source": {
    "includes": [
      "name"
    ],
    "excludes": []
  },
  "script_fields": {
    "substring_1954371849": {
      "script": {
        "source": "def substring_427843559 = doc['name'].value.substring(0, 3);return substring_427843559;",
        "lang": "painless"
      },
      "ignore_failure": false
    }
  }
}
```

As per Elastic doc https://www.elastic.co/guide/en/elasticsearch/reference/7.3/modules-scripting-using.html#prefer-params

the script should use parameters, not hardcoded fields

Basically, in generated DSL instead of this
```json
                "script": {
                  "script": {
                    "source": "def substring_1707568387 = doc['name'].value.substring(0, 3);substring_1707568387 == 'bob'",
                    "lang": "painless",
```

we should have 

```json
                  "script": {
                    "source": " doc[params.p0].value.substring(params.p1, params.p2) == params.p3",
                    "lang": "painless",
                    "params": {
                      "p0" : "name",
                      "p1": 0,
                      "p2": 3,
                      "p3": "bob"
                    }
                  },
```

for substring and any other function.

galkk:
Otherwise, theoretically, there could be even kind of DDOS attack, because Elastic limits amount of inline scripts precompiled

> If you compile too many unique scripts within a small amount of time, Elasticsearch will reject the new dynamic scripts with a circuit_breaking_exception error. By default, up to 15 inline scripts per minute will be compiled. You can change this setting dynamically by setting script.max_compilations_rate.

[...]

PR #260, opened 2019-10-30
Support string operators: ASCII, RTRIM, LTRIM, LOCATE, LENGTH, REPLACE

chloe-zh:
*Issue #, if available:*

* [**Issue #224 Support for extra functions required by BI tools**](https://github.com/opendistro-for-elasticsearch/sql/issues/224)
* [**Issue #235 Unable to recognize certain function names**](https://github.com/opendistro-for-elasticsearch/sql/issues/235)
* [**Issue #226 SUBSTRING() function is 0-indexed**](https://github.com/opendistro-for-elasticsearch/sql/issues/226)

*Description of changes:*

* Implemented string operators: ASCII, RTRIM, LTRIM, LOCATE, LENGTH, REPLACE
* Modified SUBSTRING: 1) SUBSTRING(source, start, end) --> SUBSTRING(source, start, length), 2) 0-index --> 1-index
* Added IT and UT

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -49,6 +49,7 @@ public enum ScalarFunction implements TypeExpression {
     E(func().to(DOUBLE)),
     EXP(func(T(NUMBER)).to(T)),
     EXPM1(func(T(NUMBER)).to(T)),
+    LENGTH(func(T(STRING)).to(INTEGER)),
     FLOOR(func(T(NUMBER)).to(T)),
     LOG(
         func(T(NUMBER)).to(T),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -68,6 +68,7 @@ public enum ScalarFunction implements TypeExpression {
     ),
     RADIANS(func(T(NUMBER)).to(T)),
     RANDOM(func(T(NUMBER)).to(T)),
+    REPLACE(func(STRING, STRING, STRING).to(STRING)),
     RINT(func(T(NUMBER)).to(T)),
     ROUND(func(T(NUMBER)).to(T)),
     SIGN(func(T(NUMBER)).to(T)),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -51,11 +51,11 @@ public enum ScalarFunction implements TypeExpression {
     EXP(func(T(NUMBER)).to(T)),
     EXPM1(func(T(NUMBER)).to(T)),
     FLOOR(func(T(NUMBER)).to(T)),
-    LENGTH(func(T(STRING)).to(INTEGER)
+    LENGTH(func(STRING).to(INTEGER)
 ),
     LOCATE(
-            func(STRING, T(STRING), INTEGER).to(INTEGER),
-            func(STRING, T(STRING)).to(INTEGER)
+            func(STRING, STRING, INTEGER).to(INTEGER),
+            func(STRING, STRING).to(INTEGER)
     ),
     LOG(
         func(T(NUMBER)).to(T),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -49,8 +49,14 @@ public enum ScalarFunction implements TypeExpression {
     E(func().to(DOUBLE)),
     EXP(func(T(NUMBER)).to(T)),
     EXPM1(func(T(NUMBER)).to(T)),
-    LENGTH(func(T(STRING)).to(INTEGER)),
     FLOOR(func(T(NUMBER)).to(T)),
+//    LEFT(func(T(STRING), INTEGER).to(T)),
+    LENGTH(func(T(STRING)).to(INTEGER)
+),
+    LOCATE(
+            func(STRING, T(STRING), INTEGER).to(INTEGER),
+            func(STRING, T(STRING)).to(INTEGER)
+    ),
     LOG(
         func(T(NUMBER)).to(T),
         func(T(NUMBER), NUMBER).to(T)
@@ -68,7 +74,7 @@ public enum ScalarFunction implements TypeExpression {
     ),
     RADIANS(func(T(NUMBER)).to(T)),
     RANDOM(func(T(NUMBER)).to(T)),
-    REPLACE(func(STRING, STRING, STRING).to(STRING)),
+    REPLACE(func(T(STRING), STRING, STRING).to(T)),
     RINT(func(T(NUMBER)).to(T)),
     ROUND(func(T(NUMBER)).to(T)),
     SIGN(func(T(NUMBER)).to(T)),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -67,6 +67,7 @@ public enum ScalarFunction implements TypeExpression {
         func(T(STRING)).to(T),
         func(T(STRING), STRING).to(T)
     ),
+    LTRIM(func(T(STRING)).to(T)),
     PI(func().to(DOUBLE)),
     POW, POWER(
         func(T(NUMBER)).to(T),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -51,7 +51,6 @@ public enum ScalarFunction implements TypeExpression {
     EXP(func(T(NUMBER)).to(T)),
     EXPM1(func(T(NUMBER)).to(T)),
     FLOOR(func(T(NUMBER)).to(T)),
-//    LEFT(func(T(STRING), INTEGER).to(T)),
     LENGTH(func(T(STRING)).to(INTEGER)
 ),
     LOCATE(
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -31,6 +31,7 @@ import static com.amazon.opendistroforelasticsearch.sql.antlr.semantic.types.spe
 public enum ScalarFunction implements TypeExpression {
 
     ABS(func(T(NUMBER)).to(T)), // translate to Java: <T extends Number> T ABS(T)
+    ASCII(func(T(STRING)).to(T)),
     ASIN(func(T(NUMBER)).to(T)),
     ATAN(func(T(NUMBER)).to(T)),
     ATAN2(func(T(NUMBER), NUMBER).to(T)),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
@@ -77,6 +77,7 @@ public enum ScalarFunction implements TypeExpression {
     REPLACE(func(T(STRING), STRING, STRING).to(T)),
     RINT(func(T(NUMBER)).to(T)),
     ROUND(func(T(NUMBER)).to(T)),
+    RTRIM(func(T(STRING)).to(T)),
     SIGN(func(T(NUMBER)).to(T)),
     SIGNUM(func(T(NUMBER)).to(T)),
     SIN(func(T(NUMBER)).to(T)),
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
@@ -54,7 +54,7 @@ public class SQLFunctions {
     );
 
     private static final Set<String> stringOperators = Sets.newHashSet(
-            "split", "concat_ws", "substring", "trim", "lower", "upper", "ascii", "rtrim", "locate",
+            "split", "concat_ws", "substring", "trim", "lower", "upper", "ascii", "rtrim", "ltrim", "locate",
             "length", "replace", "left", "right"
     );
 
@@ -309,6 +309,10 @@ public class SQLFunctions {
                 break;
             case "rtrim":
                 functionStr = rtrim((SQLExpr) paramers.get(0).value);
+                break;
+            case "ltrim":
+                functionStr = ltrim((SQLExpr) paramers.get(0).value);
+                break;
             default:
 
         }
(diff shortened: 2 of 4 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -242,6 +236,93 @@ public void functionPow() throws Exception {
         );
     }
 
+    @Test
+    public void operatorSubstring() {
+        assertThat(
+                executeQuery("SELECT substring('sampleName', 1, 4) FROM " + TEST_INDEX_ACCOUNT,
+                        "jdbc"),
+                containsString("samp")
```
abbashus: To make sure we are actually getting a substring, include an assert here to check it does not contains `sampleName ` i.e. 

```
executeQuery("SELECT substring('sampleName', 1, 4) FROM " + TEST_INDEX_ACCOUNT, "jdbc"),
               not( containsString("sampleName"))
```

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -242,6 +236,93 @@ public void functionPow() throws Exception {
...
+        );
+
+    }
+
+    @Test
+    public void operatorReplace() {
+        String query = "SELECT REPLACE('elastic', 'el', 'fant') FROM " + TEST_INDEX_ACCOUNT;
+        assertThat(
+                executeQuery(query, "jdbc"),
+                containsString("fantastic")
```
abbashus: Thats a good one :)

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
chloe-zh: Good point, will do it! Thanks! @abbashus

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -242,6 +236,93 @@ public void functionPow() throws Exception {
...
+                executeQuery("SELECT RTRIM(' sampleName  ') FROM " + TEST_INDEX_ACCOUNT, "jdbc"),
+                containsString(" sampleName")
+        );
+    }
+
+    @Test
+    public void ltrim() {
+        assertThat(
+                executeQuery("SELECT LTRIM(' sampleName  ') FROM " + TEST_INDEX_ACCOUNT, "jdbc"),
+                containsString("sampleName  ")
```
abbashus: Here also add a negative assert to check it does not contain ` "   sampleName "`, because `containsString` assertion will work even if it not left trimmed.

chloe-zh: @abbashus I've changed the assert in `substring()`, `ltrim()` and `rtrim()` to compare the content in /hits/fields array of the response and that would be more straightforward. Thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/StringOperatorsTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    public void substringIndexOutOfBoundTest() {
+        String query = "SELECT substring('sampleName', 0, 20) FROM accounts";
+        ScriptField scriptField = CheckScriptContents.getScriptFieldFromQuery(query);
+        assertTrue(
+                CheckScriptContents.scriptContainsString(
+                        scriptField,
+                        "'sampleName'.substring(0, 10)"
```
dai-chen: Should the second argument be 20? Did this test pass?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -620,6 +665,69 @@ private String upper(String property, String culture) {
         return property + ".toUpperCase(Locale.forLanguageTag(\"" + culture + "\"))";
     }
 
+    private Tuple<String, String> length(SQLExpr field, String valueName) {
```
dai-chen: Do you happen to know what's this `valueName` and why other functions like `locate`, `l/rtrim` don't have it?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/StringOperatorsTest.java
chloe-zh: @dai-chen the es x-pack will not throw exception when the index are out of bound execute it like the endIndex is corrected to the index of the last position in a string, but the painless language does behave in that way. I correct the out-of-bound indexes to in-bound so that it behaves like x-pack.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -620,6 +665,69 @@ private String upper(String property, String culture) {
...
+    private Tuple<String, String> locate(String pattern, SQLExpr source, int start) {
+        String name = nextId("locate");
+        String docSource = getPropertyOrStringValue(source);
+        start = start < 1 ? 0 : start - 1;
+        return new Tuple<>(name, StringUtils.format(
+                "int count=%d;int res=-1;" + "while(count< %s.length() - %s.length()) {"
+                        + "if (%s == %s.substring(count, count+%s.length())) {"
+                        + "res = count;break;} "
+                        + "else {count++;}}"
+                        + def(name, "res+1"),
```
dai-chen: If I understand right, this `locate` function is essentially `indexOf` in Java/Painless. Could we make use of `indexOf(String, int)` API directly?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -620,6 +665,69 @@ private String upper(String property, String culture) {
...
+                start, docSource, pattern, pattern, docSource, pattern)
+                );
+    }
+
+    private Tuple<String, String> rtrim(SQLExpr field) {
+        String name = nextId("rtrim");
+        String fieldString = getPropertyOrStringValue(field);
+        return new Tuple<>(name, StringUtils.format(
+                "int pos=%s.length();"
+                + "while(%s.substring(pos-1, pos)==' ' && pos > 0) {pos --;} "
```
dai-chen: Please try `charAt` because I assume `substring` will generate new string which is not efficient.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/StringOperatorsTest.java
dai-chen: Got it. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
dai-chen: Btw, another option for whitespace checking is to use `Character.isWhitespace(int)` static method.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
```diff
@@ -50,6 +51,12 @@
     EXP(func(T(NUMBER)).to(T)),
     EXPM1(func(T(NUMBER)).to(T)),
     FLOOR(func(T(NUMBER)).to(T)),
+    LENGTH(func(T(STRING)).to(INTEGER)
+),
+    LOCATE(
+            func(STRING, T(STRING), INTEGER).to(INTEGER),
+            func(STRING, T(STRING)).to(INTEGER)
```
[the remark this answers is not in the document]
chloe-zh: Done, thanks! @dai-chen

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: Actually the `nameName` is passed as null. I've deleted them.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: I've updated them to indexOf(), thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: Done, thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -53,7 +54,12 @@
     );
 
     private static final Set<String> stringOperators = Sets.newHashSet(
```
galkk: Seems like we starting to have considerable duplication here with our grammar file.
This is not a request for this PR, but if that information is somehow . For example those sets doesn't give us anything useful
```
    private static final Set<String> stringOperators = Sets.newHashSet(
            "split", "concat_ws", "substring", "trim", "lower", "upper"
    );
```

and are used only below

```
public static Schema.Type getScriptFunctionReturnType(String functionName) {
        functionName = functionName.toLowerCase();

        if (dateFunctions.contains(functionName) || stringOperators.contains(functionName)) {
            return Schema.Type.TEXT;
        }

        if (mathConstants.contains(functionName) || numberOperators.contains(functionName)
                || trigFunctions.contains(functionName) || binaryOperators.contains(functionName)
                || utilityFunctions.contains(functionName)) {
            return Schema.Type.DOUBLE;
        }

        throw new UnsupportedOperationException(
                String.format(
                        "The following method is not supported in Schema: %s",
                        functionName));
    }

```


Can we actually get rid of them and use grammar information to get function types from grammar (or vice versa, extract each function definition into some entity that can generate script, and have metadata, and bring it to grammar)?

@dai-chen what do you think?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -620,6 +665,69 @@ private String upper(String property, String culture) {
...
+    }
+
+    // es behavior: both 'start' and return value are 1-index; return 0 if pattern does not exist;
+    // support out-of-bound index
+    private Tuple<String, String> locate(String pattern, SQLExpr source, int start) {
+        String name = nextId("locate");
+        String docSource = getPropertyOrStringValue(source);
+        start = start < 1 ? 0 : start - 1;
+        return new Tuple<>(name, StringUtils.format(
+                "int count=%d;int res=-1;" + "while(count< %s.length() - %s.length()) {"
```
galkk: What is the purpose of this "+" inside of string expression?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -620,6 +665,69 @@ private String upper(String property, String culture) {
...
+                fieldString, fieldString, fieldString
+        ));
+    }
+
+    private Tuple<String, String> ltrim(SQLExpr field) {
+        String name = nextId("ltrim");
+        String fieldString = getPropertyOrStringValue(field);
+        return new Tuple<>(name, StringUtils.format(
+                "int pos=0;"
+                + "while(%s.substring(pos, pos+1)==' ' && pos < %s.length()) {pos ++;} "
```
galkk: From the doc it seems like painless exposes charAt, so you don't need to substrings

https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-api-reference-shared-java-lang.html#painless-api-reference-shared-String

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -620,6 +665,69 @@ private String upper(String property, String culture) {
...
+                + def(name, "%s.substring(pos, %s.length())"),
+                fieldString, fieldString, fieldString, fieldString
+        ));
+    }
+
+    private Tuple<String, String> ascii(SQLExpr field) {
+        String name = nextId("ascii");
```
galkk: hm. So if I'll pass "abc" to that function, it will return only ascii code of "a".
Is it supposed contract of this function?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
galkk: I think equals is more proper here

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -242,6 +236,91 @@ public void functionPow() throws Exception {
...
+        );
+
+    }
+
+    @Test
+    public void operatorReplace() {
+        String query = "SELECT REPLACE('elastic', 'el', 'fant') FROM " + TEST_INDEX_ACCOUNT;
+        assertThat(
+                executeQuery(query, "jdbc"),
+                containsString("fantastic")
```
galkk: containsString -> equals

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
dai-chen: Sure, I agree we should refactor the function in more OO way (which is convenient to integrate into our Logical Plan later). And one thing I want to mention is that the whole type system in new semantic analyzer is based on Elasticsearch data type. But here we are mapping function return type to JDBC type.

galkk:
BTW, I've created issue #150 long time ago. SInce you're implementing new functions, I recommend to change generated scripts to utilize positional parameters from elastic

So this

return new Tuple<>(name, StringUtils.format(
                "int pos=%s.length()-1;"
                + "while(pos >= 0 && Character.isWhitespace(%s.charAt(pos))) {pos --;} "
                + def(name, "%s.substring(0, pos+1)"),
                fieldString, fieldString, fieldString
        ));

May become something like

return new Tuple<>(name, StringUtils.format(
                "int pos=params[p0.length]-1;"
                + "while(pos >= 0 && Character.isWhitespace(params[p0.length].charAt(pos))) {pos --;} "
                + def(name, "%s.substring(0, pos+1)"),
                fieldString
        ));

references:
https://github.com/opendistro-for-elasticsearch/sql/issues/150
https://www.elastic.co/guide/en/elasticsearch/reference/7.3/modules-scripting-using.html#prefer-params

Since you're in the context of an issue, could you take a look how hard will it be change the implementation to utilize this? I think ASCII function is the best candidate for exploration

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
abbashus: I checked the MySQL dialect, thats how it is works.

chloe-zh:
> BTW, I've created issue #150 long time ago. SInce you're implementing new functions, I recommend to change generated scripts to utilize positional parameters from elastic
> 
> So this
> 
> return new Tuple<>(name, StringUtils.format(
> "int pos=%s.length()-1;"
> + "while(pos >= 0 && Character.isWhitespace(%s.charAt(pos))) {pos --;} "
> + def(name, "%s.substring(0, pos+1)"),
> fieldString, fieldString, fieldString
> ));
> 
> May become something like
> 
> return new Tuple<>(name, StringUtils.format(
> "int pos=params[p0.length]-1;"
> + "while(pos >= 0 && Character.isWhitespace(params[p0.length].charAt(pos))) {pos --;} "
> + def(name, "%s.substring(0, pos+1)"),
> fieldString
> ));
> 
> references:
> #150
> https://www.elastic.co/guide/en/elasticsearch/reference/7.3/modules-scripting-using.html#prefer-params
> 
> Since you're in the context of an issue, could you take a look how hard will it be change the implementation to utilize this? I think ASCII function is the best candidate for exploration

Yes sure! I looked into our code base, when making the field in `FieldMaker.java`, a linkedlist structured params is created, which stores the alias and the script code. I guess it would not be that hard if we attempt to add one more param named `params` in it, and then extract the paramers from the script. But the script is utilized in so many cases in our plugin, it's almost everywhere, so it would not be easy to finish overwriting in a short term.

I got your point to pass the params that might change in different cases, like the `multiplier` in the elastic documentation, and the `substring(params.p1, params.p2) == params.p3` in issue #150, and even the replacement of the generated variable `substring_1707568387`. But one thing I was sort of confused about: the is it necessary to set the doc name as a paramer as well since it seems like the doc is passed as a whole column rather than a single individual paramer?

Issue #239, opened 2019-10-18
Add support for CASE statement


rupal-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
Dataset: Flights sample data (loaded through Kibana)
SQL Plugin version: 1.1.0.1
Tableau version: 2019.3

Query: `` SELECT (case OriginWeather when 'Sunny' then '1' when 'Rain' then '0' else 'NA' end) AS Calculation_462181953506873347 FROM kibana_sample_data_flights``

Result: No data
```
{[\n]"
    "error": {[\n]"
    "reason": "Invalid SQL query",[\n]"
    "details": "err find condition class com.alibaba.druid.sql.ast.expr.SQLCharExpr",[\n]"
    "type": "SqlParseException"[\n]"
  },[\n]"
  "status": 400[\n]"
}
```

Expected Behaviour: Should return `1` for `Sunny`, `0` for `Rain` and `NA` for all other values in column `OriginWeather`

jordanw-bq:
Different behaviour in CASE with NOT(). This might be related, or it could be a separate issue with the NOT() function.

Query:
`SELECT (case when (AvgTicketPrice > 500) THEN 'High' when not (AvgTicketPrice > 500) then 'Low' else null end) AS Calculation_462181953506873347 FROM kibana_sample_data_flights`

Result:
```
Type: ClassCastException. Details: class com.amazon.opendistroforelasticsearch.sql.domain.MethodField cannot be cast to class com.amazon.opendistroforelasticsearch.sql.domain.ScriptMethodField (com.amazon.opendistroforelasticsearch.sql.domain.MethodField and com.amazon.opendistroforelasticsearch.sql.domain.ScriptMethodField are in unnamed module of loader java.net.FactoryURLClassLoader @3330f3ad)
```

dai-chen:
I can find 1 test case in codebase for `CASE` with a little different syntax: `SELECT address, CASE WHEN gender='0' THEN 'aaa' ELSE 'bbb' END AS a2345, count(age)`. Maybe we only support case statement written in this manner.

dai-chen:
`CASE` statement does have 2 forms: https://dev.mysql.com/doc/refman/5.7/en/case.html.

PR #262, opened 2019-10-31
Support CASE statement in one more grammer

chloe-zh:
*Issue #, if available:*

* [**Issue #239 Add support for CASE statement**](https://github.com/opendistro-for-elasticsearch/sql/issues/239)

*Description of changes:*

* Support the switch-like case statement, e.g. `SELECT (case OriginWeather when 'Sunny' then '1' when 'Rain' then '0' else 'NA' end) AS Calculation_462181953506873347 FROM kibana_sample_data_flights`.
* Support the case statement pass the JDBC formatter.
* Added UT and IT.


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -1740,6 +1740,32 @@ public void backticksQuotedAliasInJDBCResponseTest() {
         assertTrue(response.contains("\"alias\": \"name\""));
     }
 
+    @Test
+    public void caseWhenSwitchTest() throws IOException {
+        JSONObject response = executeQuery("SELECT CASE age " +
+                "WHEN '30' THEN '1' " +
+                "WHEN '40' THEN 2 " +
+                "ELSE '0' END AS cases FROM " + TEST_INDEX_ACCOUNT + " WHERE age IS NOT NULL");
```
penghuo: what is the expected result if age IS NULL?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -269,7 +269,7 @@ public void operatorReplace() {
         String query = "SELECT REPLACE('elastic', 'el', 'fant') FROM " + TEST_INDEX_ACCOUNT;
         assertThat(
                 executeQuery(query, "jdbc"),
-                equalTo("fantastic")
+                containsString("fantastic")
```
galkk: some old change slipped back?

chloe-zh: i made this change back cuz the result of `executeQuery` is the whole response output from the jdbc formatter

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -1740,6 +1740,32 @@ public void backticksQuotedAliasInJDBCResponseTest() {
...
+        JSONObject hit = getHits(response).getJSONObject(0);
+        String age = hit.query("/_source/age").toString();
+        String cases = age.equals("30") ? "1" : age.equals("40") ? "2" : "0";
+
+        assertThat(cases, equalTo(hit.query("/fields/cases/0")));
+    }
+
+    @Test
+    public void caseWhenJdbcResponseTest() {
```
davidcui-amzn: How do the tests behave when ORDER BY/GROUP BY are added?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -1116,6 +1116,36 @@ public void caseWhenTestWithouhtElseExpr() throws SqlParseException {
 
     }
 
+    @Test
+    public void caseWhenSwitchTest() throws SqlParseException {
+        String query = "SELECT CASE weather "
```
davidcui-amzn: Small thing- but increasing the indent so that all the lines of the query are inline would improve readability

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -1116,6 +1116,36 @@ public void caseWhenTestWithouhtElseExpr() throws SqlParseException {
...
+        for (Field field : select.getFields()) {
+            if (field instanceof MethodField) {
+                MethodField methodField = (MethodField) field;
+                String alias = (String) methodField.getParams().get(0).value;
+                String scriptCode = (String) methodField.getParams().get(1).value;
+                Assert.assertEquals(alias, "case");
+                Matcher docValue = Pattern.compile("doc\\['weather'].value").matcher(scriptCode);
+                Matcher number = Pattern.compile(" (\\s+'Sunny') | (\\s+'Rainy')").matcher(scriptCode);
+
+                AtomicInteger docValueCounter = new AtomicInteger();
```
galkk: why do you need atomicinteger here? nothing in this test is parallel

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -1116,6 +1116,36 @@ public void caseWhenTestWithouhtElseExpr() throws SqlParseException {
...
+    @Test
+    public void caseWhenSwitchTest() throws SqlParseException {
+        String query = "SELECT CASE weather "
+                + "WHEN 'Sunny' THEN '0' "
+                + "WHEN 'Rainy' THEN '1' "
+                + "ELSE 'NA' END AS case "
+                + "FROM t";
+        SQLExpr sqlExpr = queryToExpr(query);
+        Select select = parser.parseSelect((SQLQueryExpr) sqlExpr);
+        for (Field field : select.getFields()) {
```
galkk: I'm really suspicuous about this test. Can we get rid of logic and evaluate assertions as much as we can?

chloe-zh: Yes, both here and the atomic integer is what i referred from the previous code. I was also suspicious about the necessity to do it in parallel since we don't touch any concurrent computing in our plugin. I will modify the test.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CaseWhenParser.java
```diff
@@ -45,6 +48,15 @@ public CaseWhenParser(SQLCaseExpr caseExpr, String alias, String tableAlias) {
     public String parse() throws SqlParseException {
         List<String> result = new ArrayList<>();
 
+        if (caseExpr.getValueExpr() instanceof SQLIdentifierExpr) {
```
penghuo: what if caseExpr.getValueExpr() instanceof SQLIdentifierExpr return false?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
chloe-zh: It says "all shards failed", which is the same with the behavior in functions when the "null" rows are not excluded in the functions/operators.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CaseWhenParser.java
dai-chen: > what if caseExpr.getValueExpr() instanceof SQLIdentifierExpr return false?

If we do support for example `CASE WHEN LOG(age) = 1 ...`, maybe we can do this in negate way, `if valueExpr is not instanceof SQLBinaryOp` ?

chloe-zh: Typically the (SQLCaseExpr) caseExpr has a couple of items and a valueExpr. Each item gets a left expr, a right expr, and a binary operator expr, these 3 parts come to the conditionExpr. And valueExpr is null.  So it will go directly to traverse the items under the caseExpr.

But in this grammer `select case weather when 'sunny' then ...` the parser recognizes `weather` as the (SQLIdentifierExpr) valueExpr, but in the item the conditionExpr is kind of in a mess.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
chloe-zh: It's gonna be still working well. Thanks for reminding! I'll add them into the query.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
chloe-zh: Cool, will do it. Thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
chloe-zh: But I've no idea if it makes any sense in sql. Since the GROUP BY and ORDER BY field should also be in the SELECT field, I didn't figure out any good way to measure the response content. 

How about I creating another test to verify that it works well, since the CASE statement failed to pass the jdbc formatter and threw exceptions before I made this modification, what do you think?

opendistro-elasticsearch-sql.release-notes
```diff
@@ -5,6 +5,8 @@
 * Feature [#258](https://github.com/opendistro-for-elasticsearch/sql/issues/185): Elasticsearch 7.3.2 compatibility
 * Feature [#201](https://github.com/opendistro-for-elasticsearch/sql/pull/201): Improve query verification by adding semantic analyzer
 
+* Enhancement [#262](https://github.com/opendistro-for-elasticsearch/sql/pull/262): Support CASE statement in one more grammer
```
dai-chen: Need to add bug fix #263 here too which is just merged.

Issue #266, opened 2019-11-01
SQLFunctionsIT operator replacement test does not pass


davidcui-amzn:
The operatorReplace() integration test in SQLFunctionsIT is not functioning. Creating this issue for the quick-fix PR

PR #267, opened 2019-11-01
Fixed operatorReplace Integration Test

davidcui-amzn:
*Issue #, if available:*
#266 
*Description of changes:*
Fixed operatorReplace integration test so that gradle build will be successful. Replaced usage of equalTo() with containsString() 


davidcui-amzn:
> It shows 10 commits in this PR. Please double check before merge.

Yes- that was strange to me too. Only 1 file change though with the sole change I made.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -427,7 +427,7 @@ public void operatorReplace() {
         String query = "SELECT REPLACE('elastic', 'el', 'fant') FROM " + TEST_INDEX_ACCOUNT;
         assertThat(
                 executeQuery(query, "jdbc"),
-                equalTo("fantastic")
+                containsString("fantastic")
```
abbashus: I think we should use `equalTo` but utilize the `kvInt` like above `operatorLength ()` test.

Issue #268, opened 2019-11-01
DATETIME is not supported in Cast()


davidcui-amzn:
Extension of #225 

Queries that cast a field to type DATETIME return errors.

Example query:
```
SELECT CAST(timestamp AS DATETIME) FROM kibana_sample_data_flights LIMIT 10
```

Returns:
```
{
  "error": {
    "root_cause": [
      {
        "type": "script_exception",
        "reason": "runtime error",
        "script_stack": [
          "java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)",
          "java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110)",
          "java.base/java.lang.Double.parseDouble(Double.java:549)",
          "cast_field1 = new Date(Double.parseDouble(doc['timestamp'].value.toString()).longValue()); ",
          "                                                                ^---- HERE"
        ],
        "script": "def cast_field1 = new Date(Double.parseDouble(doc['timestamp'].value.toString()).longValue()); return cast_field1",
        "lang": "painless"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "kibana_sample_data_flights",
        "node": "3ZdEGtaiRt2uUzeNPRmNZQ",
        "reason": {
          "type": "script_exception",
          "reason": "runtime error",
          "script_stack": [
            "java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)",
            "java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110)",
            "java.base/java.lang.Double.parseDouble(Double.java:549)",
            "cast_field1 = new Date(Double.parseDouble(doc['timestamp'].value.toString()).longValue()); ",
            "                                                                ^---- HERE"
          ],
          "script": "def cast_field1 = new Date(Double.parseDouble(doc['timestamp'].value.toString()).longValue()); return cast_field1",
          "lang": "painless",
          "caused_by": {
            "type": "number_format_exception",
            "reason": "For input string: \"2019-07-22T00:00:00.000Z\""
          }
        }
      }
    ]
  },
  "status": 400
}
```

PR #271, opened 2019-11-05
Support DateTime Casting in CAST() function

davidcui-amzn:
*Issue #, if available:*
#268 
*Description of changes:*
Add support to cast to DateTime in CAST() in `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` format. 


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -312,6 +317,86 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
         }
     }
 
+    @Test
+    public void castFieldToDatetimeWithoutAliasTest() throws IOException {
+        String query = "SELECT CAST(insert_time AS DATETIME) FROM " + TestsConstants.TEST_INDEX_ONLINE +
```
dai-chen: I see that index online doesn't have predefined mapping. Could you check if field `insert_time` is already a date in ES? If so, I think we need to test against fields of different type, ex. keyword. I assume that's the point of `CAST` statement.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CastParser.java
```diff
@@ -63,15 +63,14 @@ public String parse(boolean isReturn) throws SqlParseException {
             } else if (DataType.valueOf(dataType) == DataType.STRING) {
                 result.add(String.format("def %s = %s.toString()", name, fileName));
             } else if (DataType.valueOf(dataType) == DataType.DATETIME) {
-                result.add(String.format("def %s = new Date(Double.parseDouble(%s.toString()).longValue())",
-                        name, fileName));
+                result.add(String.format("def %s = new SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")"
```
dai-chen: So the assumption is the field should be in this format. Not sure if it's easy to implement but MySQL does this in  more intelligent way.

```
>create table test (day text);
>
>insert into test values ('2019-11-06');
>insert into test values ('2019');
>insert into test values ('2019-11-06 12');
>
>select cast(day as datetime) from test;
cast(day as datetime)
2019-11-06 00:00:00
NULL
2019-11-06 12:00:00
```

davidcui-amzn: It will convert other date formats into that format as well- I tested an example with "2019-03-02" and it converted it into that SimpleDateFormat

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
davidcui-amzn: You're right- will find other fields to test the Datetime cast

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CastParser.java
dai-chen: > It will convert other date formats into that format as well- I tested an example with "2019-03-02" and it converted it into that SimpleDateFormat

Cool, thanks! Maybe you can add test cases for field with different format (after you address my first comment by creating a keyword field)

davidcui-amzn:
Closing this pull request for #310

PR #273, opened 2019-11-07
Support conditional functions: IF, IFNULL, ISNULL

chloe-zh:
*Issue #, if available:*

* [**Issue #235 Unable to recognize certain function names**](https://github.com/opendistro-for-elasticsearch/sql/issues/235)
* [**Issue #224 Support for extra functions required by BI tools**](https://github.com/opendistro-for-elasticsearch/sql/issues/224)

*Description of changes:*

* Supported conditional functions: IF, IFNULL, ISNULL.
* Added UT and IT.

*Bypassed case:*

When the query of `ISNULL` looks like `ISNULL(1/0)`, where the field is an invalid math expression, it should return 1. 

In this case, I attempted to catch arithmetic exception in the painless script, and case like `ISNULL(1*1/0)` passes the test. But case like `ISNULL(1/0)` fails. The difference of these two queries is the generated scripts of the former is more complicated. `1*1` is passed as a variable , finally it looks like `"multiply_1 = 1 * 1;divide_1 = multiply_1 / 0;return divide_1;"`, and ES catches arithmetic exception so that returns 1.  But the latter would be like `"divede_1 = 1 / 0;return divide_1;"` and ES fails to compile the script. 

In addition, this case is not among the most common cases we need to support. Thus I think we can bypass this case yet.


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -480,6 +481,58 @@ public void ascii() throws IOException {
...
+        assertThat(
+                executeQuery("SELECT IFNULL('sample', 'IsNull') AS name FROM " + TEST_INDEX_ACCOUNT),
+                hitAny(kvString("/fields/name/0", equalTo("sample")))
+        );
+//        assertThat(
+//                executeQuery("SELECT ISNULL(null, 'IsNull') AS name FROM " + TEST_INDEX_ACCOUNT),
+//                hitAny(kvString("/fields/name/0", equalTo("IsNull")))
+//        );
+    }
+
```
chloe-zh: I commented out this assert because in the query `SELECT IFNULL(null, 'IsNull') ...`, the semantic analyzer seemed to leave the `null` out and recognized the params in ISNULL as one string. But actually it does not matter since typically people will not put `null` as the condition directly into IFNULL condition field. Any suggestions?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -710,6 +746,51 @@ private String upper(String property, String culture) {
...
+            return iif(condition, expr1, expr2);
+        } else {
+            // (KVValue instance) parameters.get(0) is the parsed condition: key (String) == value (Object)
+            String key = getPropertyOrValue(paramers.get(0).key);
+            String value = getPropertyOrValue(paramers.get(0).value.toString());
+            String condition = key + " == " + value;
+            return iif(condition, expr1, expr2);
+        }
+    }
+
```
chloe-zh: Detailed explanation of cases where the condition is in the format of `a=b`:

`a` is parsed as the key (String) of a KVValue (get from `paramers.get(0)`), and `b` is parsed as the value (Object) of this KVValue.

Either `a` or `b` could be a column name, literal, or a number:
* if `isNumeric` is true --> number
* else if this string is single quoted --> literal
* else --> column name

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
penghuo: Could you add the test case of ISNULL.
1) SELECT ISNULL(1+1); expected return 0
2) SELECT ISNULL(1/0); expected return 1
https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#function_isnull

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -480,6 +481,58 @@ public void ascii() throws IOException {
...
+    @Test
+    public void ifnull() throws IOException {
+        assertThat(
+                executeQuery("SELECT IFNULL(lastname, 'unknown') AS name FROM " + TEST_INDEX_ACCOUNT
+                        + " GROUP BY name", "jdbc"),
+                containsString("\"type\": \"keyword\"")
+        );
+        assertThat(
+                executeQuery("SELECT IFNULL('sample', 'IsNull') AS name FROM " + TEST_INDEX_ACCOUNT),
```
penghuo: Could you add the test for NULL value?
1) SELECT IFNULL(NULL,10); expected 10
https://dev.mysql.com/doc/refman/8.0/en/control-flow-functions.html#function_ifnull

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
```diff
@@ -51,6 +52,22 @@
     EXP(func(T(NUMBER)).to(T)),
     EXPM1(func(T(NUMBER)).to(T)),
     FLOOR(func(T(NUMBER)).to(T)),
+    IF(
+            func(BOOLEAN, T(STRING), T(STRING)).to(T),
```
penghuo: In the MySQL mannual, the IF require there paramaters. IF(expr1,expr2,expr3)
https://dev.mysql.com/doc/refman/8.0/en/control-flow-functions.html#function_if

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
galkk: I recommed to put it as a comment into code itself :)

davidcui-amzn: +1

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/QueryFunctionsTest.java
```diff
@@ -212,6 +212,54 @@ public void numberLiteralInSelectField() {
...
+                CheckScriptContents.scriptContainsString(
+                        scriptField,
+                        "boolean cond = doc['age'].value > 35;"
+                )
+        );
+    }
+
+    @Test
+    public void ifFunctionWithEquationConditionStatement() {
+        String query = "SELECT IF(age = 35, 'elastic', 'search') AS Ages FROM accounts";
```
davidcui-amzn: Is there any way to assert that the alias `Ages`  is being properly generated?

chloe-zh: Yes like by comparing it with a proper expected output file. But I have tested the cases where the condition expression does not contain any column from a table, and the output is proper. Unit test indicates the script for the function is generated properly as well. I guess it's quite enough so the result could be exempted? Btw I also sanity tested seeveral cases. What do you think?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -480,6 +482,81 @@ public void ascii() throws IOException {
...
+//        assertThat(
+//                executeQuery("SELECT IFNULL(null, 'IsNull') AS ifnull FROM " + TEST_INDEX_ACCOUNT),
+//                hitAny(kvString("/fields/name/0", equalTo("IsNull")))
+//        );
+    }
+
+    @Test
+    public void isnull() throws IOException {
+        assertThat(
+                executeQuery("SELECT ISNULL(lastname) AS name FROM " + TEST_INDEX_ACCOUNT + "ORDER BY name", "jdbc"),
```
dai-chen: Just wonder what will happen if the field doesn't exist in some document? It is possible in ES so we may want to check if our impl can take care of this.

src/main/antlr/OpenDistroSqlParser.g4
```diff
@@ -413,6 +413,7 @@ functionNameBase
     | E | EXP | EXPM1 | FLOOR | LOG | LOG10 | LOG2 | LOWER
     | PI | POW | RADIANS | RANDOM | RINT | ROUND
     | SIN | SINH | SQRT | SUBSTRING | TAN | TRIM | UPPER | YEAR
+    | IF
```
dai-chen: any special reason not to add `IFNULL  and `ISNULL` here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -509,6 +531,16 @@ private static String getPropertyOrValue(SQLExpr expr) {
         }
     }
 
+    private static String getPropertyOrValue(String expr) {
+        if (expr.startsWith("'") && expr.endsWith("'")) {
```
dai-chen: In SQL standard, is double quote also valid string literal? Do we have the original expr to pass in here? I think check on AST expr is more reliable.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
chloe-zh: ES returns field[...] cannot be found or used here. DId you mean to precheck the field that doesn't exit before running into ES?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: The quoted string literal is parsed to SQLCharExpr regardless of being single quoted or double quoted.

src/main/antlr/OpenDistroSqlParser.g4
chloe-zh: Nope, I've not yet updated the code changes from the local repository thoroughly, I push the updates by today.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -710,6 +747,85 @@ private String upper(String property, String culture) {
...
+            String condStr = Strings.isNullOrEmpty(condition.toString()) ? null : getPropertyOrStringValue(condition);
+            return new Tuple<>(name, def(name, condStr));
+        }
+    }
+
+    private Tuple<String, String> isnull(SQLExpr expr) {
+        String name = nextId("isnull");
+        if (isProperty(expr)) {
+            return new Tuple<>(name, def(name, doc(expr) + ".size()==0 ? 1 : 0"));
+        } else {
```
galkk: you don't need that else block here, you return in previous line anyway. Remove else and decrease nestendess

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
```diff
@@ -317,8 +316,14 @@ protected T aggregateResult(T aggregate, T nextResult) {
 
     /** Named argument, ex. TOPHITS('size'=3), is under FunctionArgs -> Predicate */
     private boolean isNamedArgument(BinaryComparisonPredicateContext ctx) {
-        return ctx.getParent() != null && ctx.getParent().getParent() != null
-            && ctx.getParent().getParent() instanceof FunctionArgsContext;
+        if (ctx.getParent() != null && ctx.getParent().getParent() != null
```
galkk: opinion: that chain of getParent()'s smells funny

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -710,6 +747,85 @@ private String upper(String property, String culture) {
...
 
+    private Tuple<String, String> ifFunc(List<KVValue> paramers) {
+        String expr1 = paramers.get(1).value.toString();
+        String expr2 = paramers.get(2).value.toString();
+
+        if (paramers.get(0).value instanceof MethodField) {
+            String condition = getScriptText((MethodField) paramers.get(0).value);
+            return ifFunc(condition, expr1, expr2);
+        } else if (paramers.get(0).value instanceof SQLBooleanExpr) {
+            Boolean condition = ((SQLBooleanExpr) paramers.get(0).value).getValue();
```
galkk: could you give an example of query that will trigger that condition? I'm trying to understand how it end's up as true/false that we analyze in the method at line # 783

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
```diff
@@ -121,6 +121,15 @@ public static boolean isQuoted(String text, String quote) {
         return !Strings.isNullOrEmpty(text) && text.startsWith(quote) && text.endsWith(quote);
     }
 
+    public static boolean isNumeric(String text) {
```
galkk: we're using Guava, so this can be replaced by 

Doubles.tryParse(text) != null;

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -480,6 +482,81 @@ public void ascii() throws IOException {
         );
     }
 
+    @Test
+    public void ifTest() throws IOException {
+        assertThat(
```
galkk: split that to different tests or use parametrized tests

chloe-zh: ok will do

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
chloe-zh: Great, I'll try this out. Thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -480,6 +482,81 @@ public void ascii() throws IOException {
...
+                executeQuery("SELECT IFNULL(lastname, 'unknown') AS name FROM " + TEST_INDEX_ACCOUNT
+                        + " GROUP BY name", "jdbc"),
+                containsString("\"type\": \"keyword\"")
+        );
+        assertThat(
+                executeQuery("SELECT IFNULL('sample', 'IsNull') AS ifnull FROM " + TEST_INDEX_ACCOUNT),
+                hitAny(kvString("/fields/name/0", equalTo("sample")))
+        );
+
+        //TODO: let null pass the verification of semantic analyzer
```
galkk: either remove commented code or add it as @Ignore test with reference to github issue that will allow to fix that problem

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
dai-chen: Sorry, I meant the field does exist in index mapping but some document doesn't have it. For example, some document doesn't include `lastname` in your test case.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: @galkk Yes sure. 
When we give a query like this: `SELECT IF(true, 1, 0)`. The first param (KVValue) would get a KVValue with `key = null` and `value = (SQLBooleanExpr) "xTRUE"`, and the value of this boolean expr is true.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
chloe-zh: I've added ISNULL(1+1) returns 0, and ISNULL(1+1*1/0) returns 1 into the test. However, when we input a simple enough math expression like `1/0`, the ES throws compile error exception rather than "/ by zero exception" (arithmetic exception) so that it returns "all shards fail". We need to systematically improve our exception mechanism to catch and throw proper exception in the future, so I think this should be part of that as well.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
```diff
@@ -51,6 +52,22 @@
...
+            func(BOOLEAN, STRING, STRING).to(STRING),
+            func(BOOLEAN, NUMBER, NUMBER).to(NUMBER),
+            func(BOOLEAN, T(STRING), T(NUMBER)).to(T),
+            func(BOOLEAN, T(NUMBER), T(STRING)).to(T)
+    ),
+    IFNULL(
+            func(NUMBER, NUMBER).to(NUMBER),
+            func(STRING, STRING). to(STRING),
+            func(T(NUMBER), T(STRING)).to(T),
+            func(T(STRING), T(NUMBER)).to(T)
```
dai-chen: I just found that this `IFNULL` (as well as `IF`) is a little strange because it allows to return 2 totally different types. In this case, I think we can define its type as `ES_TYPE, ES_TYPE -> ES_TYPE` which means we only do params number check.

Note that `ES_TYPE` is the root and compatible with all other types.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
```diff
@@ -51,6 +52,22 @@
...
+    ),
+    ISNULL(
+            func(NUMBER).to(INTEGER),
+            func(STRING).to(INTEGER)
```
dai-chen: I think it should be `ES_TYPE -> INTEGER`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -710,6 +750,96 @@ private String upper(String property, String culture) {
...
+    }
+
+    private Tuple<String, String> ifFunc(Boolean condition, String expr1, String expr2) {
+        String name = nextId("if");
+        if (condition) {
+            return new Tuple<>(name, def(name, expr1));
+        } else {
+            return new Tuple<>(name, def(name, expr2));
+        }
+    }
```
dai-chen: Maybe we can inline these 2 functions into `ifFunc(List<KVValue> paramers)` just like your `ifnull` method because:

 1. They're short
 2. They share same logic `String name = nextId("if");`
 3. A little hard to follow 3 overload methods of same name.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/function/ScalarFunction.java
chloe-zh: ok cool.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -710,6 +750,96 @@ private String upper(String property, String culture) {
...
+        if (Strings.isNullOrEmpty(expr.toString())) {
+            resultStr = "1";
+        }
+        if (expr instanceof SQLCharExpr && this.generatedIds.size() > 1) {
+            // the expr is a math expression
+            String mathExpr = ((SQLCharExpr) expr).getText();
+            return new Tuple<>(name, StringUtils.format(
+                    "try {%s;} "
+                            + "catch(Exception e) "
+                            + "{def %s; if(e instanceof ArithmeticException){%s=1;} else{%s=0;} return %s;} "
```
dai-chen: maybe a little better?

```
try { } 
catch (ArithmeticException e) { return 1; }
return 0;
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: ok will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
chloe-zh: There is a post-processing of the functionStr right after the function method switch: 
`String returnCommand = ";return " + generatedFieldName + ";";`
That's why I made it that complicated.

chloe-zh: But I will make it catch arithmetic exception rather than generic exception

dai-chen: I see. Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
dai-chen: Thanks.

Issue #232, opened 2019-10-18
Add support for JOINs without aliases


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights & Ecommerce sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT AvgTicketPrice AS AvgTicketPrice,   Cancelled AS Cancelled,   Carrier AS Carrier,   DestAirportID AS DestAirportID,   DestCityName AS DestCityName,   DestCountry AS DestCountry,   DestLocation AS DestLocation FROM kibana_sample_data_ecommerce   LEFT JOIN kibana_sample_data_flights ON (kibana_sample_data_ecommerce.day_of_week_i = dayOfWeek) LIMIT 50

**Results:**
No data
```
{
  "schema": [],
  "total": 10000,
  "datarows": [
    [],
    [],
    [],
    [],
...
    []
  ],
  "size": 50,
  "status": 200
}
```

**Notes:**
This is a known limitation of JOINs with Elasticsearch, but Tableau by default will be generating JOINs without aliases.

**Expectation:**
Support for JOINs without alias

dai-chen:
Currently alias is required for `JOIN`.

PR #274, opened 2019-11-07
Support JOIN without table alias

abbashus:
*Issue #232 *

*Description of changes:*
This pull request adds the ability to have JOIN queries without table alias given all the fields belong to either table but not common to both.

*Testing:*
- Passing Gradle build and existing unit and integration test
- Added new integration tests


src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+        MySqlSelectQueryBlock query = (MySqlSelectQueryBlock) sqlSelectQuery;
+        return query.getFrom() instanceof SQLJoinTableSource
+            && ((SQLJoinTableSource) query.getFrom()).getJoinType() != SQLJoinTableSource.JoinType.COMMA;
+    }
+
+    @Override
+    public void rewrite(SQLQueryExpr root) {
+
+        final Multimap<String, Table> tableByFieldName = ArrayListMultimap.create();
+        final Set<String> tableNameAndAlias = new HashSet<>();
```
davidcui-amzn: How is tableNameAndAlias used? I don't see what it's used for other than just storing the table name and alias which doesn't seem like it's necessary to create the Set for

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+    private static final String DOT = ".";
+    private int aliasSuffix = 0;
+    private final LocalClusterState clusterState;
+
+    public JoinRewriteRule(LocalClusterState clusterState) {
+        this.clusterState = clusterState;
+    }
+
+    @Override
+    public boolean match(SQLQueryExpr root) {
```
penghuo: Does this function been called?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
```diff
@@ -94,6 +98,7 @@ public static QueryAction create(Client client, String sql) throws SqlParseExcep
                             .parseMultiSelect((SQLUnionQuery) sqlExpr.getSubQuery().getQuery());
                     return new MultiQueryAction(client, multiSelect);
                 } else if (isJoin(sqlExpr, sql)) {
+                    new JoinRewriteRule(LocalClusterState.state()).rewrite(sqlExpr);
```
penghuo: Have you explor the possiblity to combine this with SubqueryAliasRewriter?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+
+    private boolean isJoin(SQLQueryExpr sqlExpr) {
+        SQLSelectQuery sqlSelectQuery = sqlExpr.getSubQuery().getQuery();
+
+        if (!(sqlSelectQuery instanceof MySqlSelectQueryBlock)) {
+            return false;
+        }
+
+        MySqlSelectQueryBlock query = (MySqlSelectQueryBlock) sqlSelectQuery;
+        return query.getFrom() instanceof SQLJoinTableSource
```
penghuo: Suggestion. Is it possible to make the solution more generic JOIN case? Even the solution only support two table join, but the alias rewrite solution could support more table cases.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+                idExpr.setName(String.join(DOT, table.getAlias(), columnName));
+            }
+        });
+    }
+
+    private void visitTable(SQLQueryExpr root,
+                            Consumer<SQLExprTableSource> visit) {
+        root.accept(new MySqlASTVisitorAdapter() {
+            @Override
+            public void endVisit(SQLExprTableSource tableExpr) {
```
davidcui-amzn: What's the benefit of defining the function in this way?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+        final Multimap<String, Table> tableByFieldName = ArrayListMultimap.create();
+        final Set<String> tableNameAndAlias = new HashSet<>();
+        final Map<String, String> tableNameToAlias = new HashMap<>();
+
+        visitTable(root, tableExpr -> {
+            // Copied from SubqueryAliasRewriter ; Removes index type name if any
+            String tableName = tableExpr.getExpr().toString().replaceAll(" ", "").split("/")[0];
+
+            if (tableExpr.getAlias() == null) {
+                // Could we not directly use table name as alias?
```
penghuo: If the table join with itself, the alias will be duplicate. not sure wheter the syntax or semantic check this before reach here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
abbashus: Right, good catch. I used it previously for a special case, and removed that after refactoring, Will remove it.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/ESActionFactory.java
abbashus: No, I haven't explored that possibility.  I wanted separation of concerns and did not wanted to convolute the code. Will check  if we can combine this cleanly with `SubqueryAliasRewriter`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
abbashus: Yes, this has not been used. I added here in case we want to use as part of `RewriteRuleExecutor` chain. I avoided adding to the `RewriteRuleExecutor` chain, to  not effect the latency in case of non-join queries.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+
+/**
+ * Rewrite rule for query involved JOIN.
+ */
+public class JoinRewriteRule implements RewriteRule<SQLQueryExpr> {
```
penghuo: Is it possible to summarize the alias rewriter alg in the doc?

abbashus: Sure, will add some documentation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
```diff
@@ -0,0 +1,191 @@
...
+                tableExpr.setAlias(createAlias(tableName));
+            }
+
+            Table table = new Table(tableName, tableExpr.getAlias());
+
+            tableNameAndAlias.add(table.getName());
+            tableNameAndAlias.add(table.getAlias());
+            tableNameToAlias.put(table.getName(), table.getAlias());
+
+            FieldMappings fieldMappings = clusterState.getFieldMappings(
```
penghuo: Is it possible to share the fieldMapping between different componnet. If I am right,  SemanticAnalyzer, TermFieldRewriter also use the mapping. As my understand, has three copy of index mapping in the memory could be expensive.

abbashus: They all already share the same [instance](https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/sql/esdomain/LocalClusterState.java#L96). So we are good here.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
abbashus: Ideally we would have needed a single Visitor like `TermFieldRewriter` and `NestedFieldRewriter` .  But since we are doing different things (conflicting) while visiting `SQLExprTableSource` in `visitTable` and `visitColumnName`, we need to have two Visitors and also they should be able to share the state `tableByFieldName` and `tableNameToAlias`. This pattern cleanly solves this problem and also makes it more readable by using intention revealing function names.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
abbashus: For self join, the alias will not be duplicate as we generate different alias for each table without alias. I have added more checks to handle joins with same tables on either side. Added more integration tests as well.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/join/JoinRewriteRule.java
abbashus: Added algorithm summary and comments where needed.

Issue #272, opened 2019-11-06
IP field types are currently not supported.


HernanMora:
### Description

It is not possible to make queries where an IP type field is selected

**Elasticsearch Version:** 6.8.3
**SQL Plugin version:** 0.10.0.0

### Request
```javascript
POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT host_ip, hostname FROM devices* WHERE host_ip = TERM('10.11.23.0/24') LIMIT 5"
}

```
### Response 
```javascript
{
  "error": {
    "reason": "IP field types are currently not supported.",
    "details": "java.lang.IllegalArgumentException: IP field types are currently not supported.",
    "type": "IllegalArgumentException"
  },
  "status": 500
}
```
Is there any way to make queries where these types of fields are selected?
Regards!

abbashus:
Did a small test to check the missing functionality.
```
PUT network

PUT network/_mapping?pretty
{
  "properties": {
    "host_ip": {
      "type": "ip"
    },
    "hostname": {
      "type": "text",
      "fields": {
        "keyword": {
          "type": "keyword"
        }
      }
    }
  }
}

PUT /network/_doc/1?pretty
{
  "host_ip": "192.168.1.1",
  "hostname" : "www.opendistro.com"
}

```

Explain seems fine:

```
POST _opendistro/_sql/_explain?
{
  "query": """
  SELECT host_ip, hostname 
  FROM network
  WHERE host_ip = TERM("192.168.0.0/16") LIMIT 5
  """
}

{
  "from" : 0,
  "size" : 5,
  "query" : {
    "bool" : {
      "filter" : [
        {
          "bool" : {
            "must" : [
              {
                "term" : {
                  "host_ip" : {
                    "value" : "192.168.0.0/16",
                    "boost" : 1.0
                  }
                }
              }
            ],
            "adjust_pure_negative" : true,
            "boost" : 1.0
          }
        }
      ],
      "adjust_pure_negative" : true,
      "boost" : 1.0
    }
  },
  "_source" : {
    "includes" : [
      "host_ip",
      "hostname"
    ],
    "excludes" : [ ]
  }
}

```

Query without `jdbc` format
```
POST _opendistro/_sql/?
{
  "query": """
  SELECT host_ip, hostname 
  FROM network
  WHERE host_ip = TERM("192.168.0.0/16") LIMIT 5
  """
}

{
  "took" : 4,
  "timed_out" : false,
  "_shards" : {
[...]

PR #275, opened 2019-11-07
Fix issue that IP type cannot pass JDBC formatter

chloe-zh:
*Issue #, if available:*

* [**Issue #272 IP field types are currently not supported.**](https://github.com/opendistro-for-elasticsearch/sql/issues/272)

*Description of changes:*

* Fixed the JDBC formatter for the IP type.
* Appended dataset with field of IP type.
* Added and passed the integration test, manually sanity tested on Kibana.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Schema.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Schema.java
@@ -99,13 +99,12 @@ public class Schema implements Iterable<Schema.Column> {
 
     // Only core ES datatypes currently supported
     public enum Type {
-        TEXT, KEYWORD, // String types
+        TEXT, KEYWORD, IP, // String types
         LONG, INTEGER, SHORT, BYTE, DOUBLE, FLOAT, HALF_FLOAT, SCALED_FLOAT, // Numeric types
         DATE, // Date types
         BOOLEAN, // Boolean types
         BINARY, // Binary types
-        INTEGER_RANGE, FLOAT_RANGE, LONG_RANGE, DOUBLE_RANGE, DATE_RANGE, // Range types
-        IP;
+        INTEGER_RANGE, FLOAT_RANGE, LONG_RANGE, DOUBLE_RANGE, DATE_RANGE; // Range types
 
         public String nameLowerCase() {
             return name().toLowerCase();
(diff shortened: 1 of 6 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Schema.java
```diff
@@ -104,7 +104,8 @@ public void remove() {
         DATE, // Date types
         BOOLEAN, // Boolean types
         BINARY, // Binary types
-        INTEGER_RANGE, FLOAT_RANGE, LONG_RANGE, DOUBLE_RANGE, DATE_RANGE; // Range types
+        INTEGER_RANGE, FLOAT_RANGE, LONG_RANGE, DOUBLE_RANGE, DATE_RANGE, // Range types
+        IP;
```
dai-chen: Do you know if any correspondent type in JDBC specification? Because all types here will be mapped to JDBC types in driver.

chloe-zh: It's gonna be varchar

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
```diff
@@ -126,4 +128,12 @@ public void dateFunctionNameCaseInsensitiveTest() {
                         "ORDER BY date_format(insert_time, 'yyyy-MM-dd', 'UTC')", "jdbc")
         );
     }
+
+    @Test
```
davidcui-amzn: Could you add ORDER BY and GROUP BY tests for JDBC format?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Schema.java
dai-chen: Cool, if so I think we need to check if there is already mapping from IP to Varchar in JDBC driver or not.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
chloe-zh: Yes sure, I will add group by aggregation here. Thanks! @davidcui-amzn

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Schema.java
chloe-zh: But the jdbc did't have the ip type mapped to varchar (or any other jdbc type). Need to add this mapping in JDBC. @dai-chen

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/JdbcTestIT.java
chloe-zh: I've added the IP type as VARCHAR of JDBC type. And it passed the sanity test in DbVisualizer. Thanks!

Issue #230, opened 2019-10-18
Add support for subqueries


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT t0.TEMP FROM (SELECT SUM(dayOfWeek) as TEMP FROM kibana_sample_data_flights) t0

**Result:**
```
Raw response received: {
  "error": {
    "reason": "Invalid SQL query",
    "details": "Different mappings are not allowed for the same field[config]: found [{type:text,fields:{keyword:{type:keyword,ignore_above:256}}}] and [{dynamic:true,properties:{buildNum:{type:keyword}}}] ",
    "type": "VerificationException"
  },
  "status": 400
}
```

**Note:**
This is a known limitation of Elasticsearch, but Tableau by default will be generating subqueries for a number of its queries. However, support for subqueries is something that can be customized using a Tableau Datasource Customization (TDC) file, though the extent of these customizations cannot be tested until the major Tableau blockers (#212, #213, #215) are resolved. 

**Expectation:**
Subquery support

dai-chen:
We don't have support for this kind of subquery. And this seems broke term field rewriter too. May need major changes to add the support.

jordanw-bq:
We've re-tested subqueries with the new changes (#278), but we are still finding errors. Even a simple case (adapted from the tests added as part of the aforementioned PR) does not seem to be working as expected.

Query: `SELECT t.TEMP as count FROM (SELECT COUNT(*) as TEMP FROM kibana_sample_data_flights) t`
Server logs:
```
[2019-12-12T01:16:05,452][INFO ][c.a.o.s.p.RestSqlAction  ] [odfe-node1] [b2007666-6ed1-4c7b-a6e8-2ce83325993a] Incoming request /_opendistro/_sql?format=jdbc: SELECT t.TEMP as count FROM (SELECT COUNT(*) as TEMP FROM kibana_sample_data_flights) t
[2019-12-12T01:16:05,457][ERROR][c.a.o.s.p.RestSqlAction  ] [odfe-node1] b2007666-6ed1-4c7b-a6e8-2ce83325993a Client side error during query execution
[...]

PR #278, opened 2019-11-08
Support subquery in from with parent only has select

penghuo:
*Issue #, if available:* #230 

*Description of changes:*
Support subquery in from with parent only has select. 

#### To reviewer
* The current solution bypass the failed semantic check failed when subquery in from. There is no good solution found to decide whether the context frame should be poped or keep. For example, the if the subquery in where clause, the frame context frame shoud be poped. But when subquery in from clause, the frame context should be keeped to make the expression in select could be resolved in the frame context.
*  The current solution push the select in the outer query to subquery in from clause and rewrite the field in the subquery.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
@@ -19,6 +19,7 @@ import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParse
 import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParser.InnerJoinContext;
 import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParser.QuerySpecificationContext;
 import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParser.SelectColumnElementContext;
+import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParser.SubqueryTableItemContext;
 import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParser.TableNamePatternContext;
 import com.amazon.opendistroforelasticsearch.sql.antlr.parser.OpenDistroSqlParserBaseVisitor;
 import org.antlr.v4.runtime.ParserRuleContext;
@@ -172,6 +173,11 @@ public class AntlrSqlParseTreeVisitor<T extends Reducible> extends OpenDistroSql
 
         visitor.endVisitQuery();
         return result;
+    }
+
+    @Override
+    public T visitSubqueryTableItem(SubqueryTableItemContext ctx) {
+        throw new EarlyExitAnalysisException("Exit when meeting subquery in from");
     }
 
     /** Visit here instead of tableName because we need alias */
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
@@ -1,0 +1,114 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.parser;
+
+import com.alibaba.druid.sql.ast.SQLExpr;
+import com.alibaba.druid.sql.ast.statement.SQLSelectItem;
+import com.alibaba.druid.sql.ast.statement.SQLSubqueryTableSource;
+import com.alibaba.druid.sql.dialect.mysql.ast.statement.MySqlSelectQueryBlock;
+import com.amazon.opendistroforelasticsearch.sql.domain.Condition;
+import com.amazon.opendistroforelasticsearch.sql.domain.Field;
+import com.amazon.opendistroforelasticsearch.sql.domain.Order;
+import com.amazon.opendistroforelasticsearch.sql.domain.Select;
+import com.amazon.opendistroforelasticsearch.sql.domain.Where;
+import com.amazon.opendistroforelasticsearch.sql.exception.SqlParseException;
+import com.google.common.base.Strings;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Definition of SubQuery Parser
+ */
+public class SubQueryParser {
+    private final SqlParser sqlParser;
+
+    public SubQueryParser(SqlParser sqlParser) {
+        this.sqlParser = sqlParser;
+    }
+
+    public boolean containSubqueryInFrom(MySqlSelectQueryBlock query) {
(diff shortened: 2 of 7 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
```diff
@@ -267,4 +274,101 @@ public void nonCorrelatedNotExistsParentWhere() throws IOException {
...
+
+        assertThat(countList, isInOrder(true));
+    }
+
+    @Test
+    public void selectFromSubqueryWithCountAndGroupByAndHavingShouldPass() throws Exception {
+        JSONObject result = executeQuery(
+                String.format("SELECT t.T1 as g, t.T2 as c " +
+                              "FROM (SELECT gender as T1, COUNT(*) as T2 " +
+                              "      FROM %s/account " +
```
davidcui-amzn: Is there a reason that there's so much whitespace before the keywords on some of the lines but not others?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
```diff
@@ -267,4 +274,101 @@ public void nonCorrelatedNotExistsParentWhere() throws IOException {
...
+        assertThat(gender.query(maleBucketPrefix + "/key"), equalTo("m"));
+        assertThat(gender.query(maleBucketPrefix + "/count/value"), equalTo(507));
+        assertThat(gender.query(femaleBucketPrefix + "/key"), equalTo("f"));
+        assertThat(gender.query(femaleBucketPrefix + "/count/value"), equalTo(493));
+    }
+
+    @Test
+    public void selectFromSubqueryWithCountAndGroupByAndOrderByShouldPass() throws IOException {
+        JSONObject result = executeQuery(
+                String.format(
```
davidcui-amzn: Should we use StringUtils instead of String here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
```diff
@@ -174,6 +175,11 @@ public T visitQuerySpecification(QuerySpecificationContext ctx) {
         return result;
     }
 
+    @Override
+    public T visitSubqueryTableItem(SubqueryTableItemContext ctx) {
+        throw new EarlyExitAnalysisException("Exit when meeting subquery in from");
```
dai-chen: Please help open an enhancement issue so we can try to figure out later. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
```diff
@@ -0,0 +1,114 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Definition of SubQuery Parser
+ */
+public class SubQueryParser {
```
dai-chen: Any chance to add UT for this new class? I'm thinking the test case, like what you did in `SubqueryAliasRewriterTest`, may help understand what the query looks like after rewrite.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
galkk: it's to differentiate nested query from outer query, I assume

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
```diff
@@ -267,4 +274,101 @@ public void nonCorrelatedNotExistsParentWhere() throws IOException {
...
+                        "SELECT t.TEMP1 as count, t.TEMP2 as balance " +
+                        "FROM (SELECT COUNT(*) as TEMP1, SUM(balance) as TEMP2 " +
+                        "      FROM %s/account) t",
+                        TEST_INDEX_ACCOUNT));
+
+        assertThat(result.query("/aggregations/count/value"), equalTo(1000));
+        assertThat(result.query("/aggregations/balance/value"), equalTo(25714837.0));
+    }
+
+    private Matcher<? super List<? extends Comparable>> isInOrder(boolean asc) {
```
galkk: That's useful. Could you move it to MatcherUtils?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+        return subquerySelect;
+    }
+
+    private Map<String, Function<String, String>> prepareFieldAliasRewriter(List<SQLSelectItem> selectItems,
+                                                                          String owner) {
+        HashMap<String, Function<String, String>> selectMap = new HashMap<>();
+        for (SQLSelectItem item : selectItems) {
+            if (Strings.isNullOrEmpty(item.getAlias())) {
+                selectMap.put(getFieldName(item.getExpr(), owner), s -> s);
```
galkk: replace 
    s->s 

with 

Function.identity()

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
galkk: I doubt the name of this class. What does it parse?

galkk: Why we can't rewrite query based purely on the druid's expression and need to parse it to our domain model?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
penghuo: Yes, it is to differentiate nested query.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
penghuo: Done. https://github.com/opendistro-for-elasticsearch/sql/issues/281

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
penghuo: Done. https://github.com/opendistro-for-elasticsearch/sql/issues/281

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
penghuo: Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
penghuo: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
penghuo: Done.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SubqueryIT.java
penghuo: I perfer to use `assertTrue(Ordering.natural().isOrdered(countList));` instead of add another matcher.
I think maybe this is the reason why this request has been in pending status for long time. https://code.google.com/archive/p/hamcrest/issues/174

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/SubQueryParser.java
penghuo: The main reason is the SqlParser could be modeled as LogicalPlan generation layer for the current architecture. In this, the subquery rewrite logic should in SqlParser model.

penghuo: Done.

Issue #234, opened 2019-10-18
Arithmetic operations not supported in WHERE clause


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT FlightNum FROM kibana_sample_data_flights where (AvgTicketPrice + 100) <= 1000

**Result:**
```
Raw response received: {
  "error": {
    "reason": "Invalid SQL query",
    "details": "No enum constant com.amazon.opendistroforelasticsearch.sql.domain.Where.CONN.<=",
    "type": "IllegalArgumentException"
  },
  "status": 400
}
```

**Expected Behaviour:**
No error, and the operation should be applied to the field before the comparison is made.
Example results:
```
| FlightNum | 
| 9HY9SWR | 
| X98CCZO | 
| UFK2WIZ | 
| EAYQW69 | 
| 58U013N | 
| XEJ78I2 | 
| EVARI8I | 
...
```

dai-chen:
Similar as this one: https://github.com/opendistro-for-elasticsearch/sql/issues/194. Currently we don't support arithmetic expression anywhere in query.

dai-chen:
This is already fixed in new query engine. However, the new engine is disabled by default for now. You can enable it by following the instruction here: https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst#opendistro-sql-engine-new-enabled

Issue #194, opened 2019-09-24
Aggregations with basic algebraic operations are not supported.


chloe-zh:
Aggregations combined with basic algebraic operations failed, including cases like:
`MAX(a + const)` ; `MAX(a + b)`; `MAX(a) + const`; `MAX(a) + MAX(b)`; etc.

Below is an example of a SQL query against AES instance v7.1.1.
**Example indexed document**
```
{
    "_index": "production.asset-metrics.history",
    "_type": "_doc",
    "_id": "EF602B6FF7E2956B0E627EF5901FFC08A9277556",
    "_score": 1.0,
    "_source": {
        "location": {
            "lat": 52.88041,
            "lon": -1.704096
        },
        "date": "2019-07-29T21:21:23Z",
        "organisation_id": "5c76863de13a541c18c3d963",
        "equipment_id": "5d3ebd4e2cf0ae28ea420604",
        "group_id": "5d3ebd4e2cf0ae28ea420627",
        "product_group_code": "500",
        "oem": "JCB",
        "model": "531-70",
        "latitude": 52.88041,
        "longitude": -1.704096,
        "altitude_in_metres": 81.875,
        "time_operating_in_hours": 110.0,
        "time_idle_in_hours": 33.1552777777778,
        "fuel_used_in_litres": 514,
        "fuel_remaining_percent": 98.0
    }
}
```

**Example query**
```
POST <host>/_opendistro/_sql
{
    "query": "SELECT equipment_id, (MAX(time_operating_in_hours) - MIN(time_operating_in_hours)) FROM development-asset-metrics-history GROUP BY equipment_id"
}
```

**Example Response**
```
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "",
    "type": "NullPointerException"
  },
  "status": 400
}
```

chloe-zh:
Currently have obtained the expected DSL for this case.

penghuo:
closed with commit.https://github.com/opendistro-for-elasticsearch/sql/commit/3730ac6e9b3a363c51432b34d2104567eb9babb7

PR #286, opened 2019-11-15
Better error reporting for unsupported case like max(x)-min(y)

galkk:
*Issue #, if available:* #194, #234 

*Description of changes:*
Better error reporting for such queries

```
POST /_opendistro/_sql/_explain
{
  "query":"""SELECT DestCountry, dayOfWeek, max(FlightDelayMin) - min(FlightDelayMin) + avg(FlightDelayMin)
    FROM kibana_sample_data_flights
    GROUP BY DestCountry, dayOfWeek
  """
```

returns now 

```
{
  "error": {
    "details": "The complex aggregate expressions are not implemented yet: MAX(FlightDelayMin) - MIN(FlightDelayMin)",
    "type": "SqlFeatureNotImplementedException"
  },
  "status": 503
}
```

instead of non-sensical:

```
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "",
    "type": "NullPointerException"
  },
  "status": 400
}
```

and another example:
````
                        "SELECT DestCountry, dayOfWeek, log(max(FlightDelayMin))" +
                        "   FROM kibana_sample_data_flights\n" +
                        "   GROUP BY DestCountry, dayOfWeek\n"));
```

witll throw

                  "Function calls of form 'log(MAX(...))' are not implemented yet");


### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
@@ -34,6 +34,7 @@ import com.amazon.opendistroforelasticsearch.sql.domain.KVValue;
 import com.amazon.opendistroforelasticsearch.sql.domain.MethodField;
 import com.amazon.opendistroforelasticsearch.sql.domain.ScriptMethodField;
 import com.amazon.opendistroforelasticsearch.sql.domain.Where;
+import com.amazon.opendistroforelasticsearch.sql.exception.SqlFeatureNotImplementedException;
 import com.amazon.opendistroforelasticsearch.sql.exception.SqlParseException;
 import com.amazon.opendistroforelasticsearch.sql.utils.SQLFunctions;
 import com.amazon.opendistroforelasticsearch.sql.utils.Util;
@@ -230,6 +231,7 @@ public class FieldMaker {
         SQLMethodInvokeExpr methodInvokeExpr = new SQLMethodInvokeExpr(operator, null);
         methodInvokeExpr.addParameter(expr.getLeft());
         methodInvokeExpr.addParameter(expr.getRight());
+        methodInvokeExpr.putAttribute("source", expr);
         return methodInvokeExpr;
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -30,18 +30,20 @@
...
 import com.amazon.opendistroforelasticsearch.sql.query.maker.QueryMaker;
 import com.amazon.opendistroforelasticsearch.sql.query.multi.MultiQuerySelect;
 import com.amazon.opendistroforelasticsearch.sql.util.CheckScriptContents;
+import org.elasticsearch.client.Client;
 import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.search.builder.SearchSourceBuilder.ScriptField;
-import org.junit.Assert;
-import org.junit.BeforeClass;
-import org.junit.Test;
+import org.junit.*;
```
dai-chen: I assume this violates our Checkstyle and should fail the GitHub Action check :)

dai-chen: It seems passed though?

galkk: yeah, seems like it is warning :)

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -317,6 +319,10 @@ public MethodField makeMethodField(String name, List<SQLExpr> arguments, SQLAggr
             } else if (object instanceof SQLCastExpr) {
                 String scriptCode = new CastParser((SQLCastExpr) object, alias, tableAlias).parse(false);
                 paramers.add(new KVValue("script", new SQLCharExpr(scriptCode)));
+            } else if (object instanceof SQLAggregateExpr) {
```
dai-chen: Does this also cover other cases such as `MAX(age + 1)`, `MAX(LOG(age))`? If not, any chance to take care of them too?

chloe-zh:
I think the code changes for now covers more than cases like `max(x) - min(y)`. It seems the `object` is always the left one of a binary operation, so it can cover the cases as long as it looks like `max(x) + y`. Could you implement the cases of `x + min(y)`-like in this PR as well?

galkk:
@chloe-zh 

The case that you've mentioned already works

```
POST /_opendistro/_sql/_explain
{
  "query":"""SELECT 1+max(FlightDelayMin)
    FROM kibana_sample_data_flights
    GROUP BY DestCountry, dayOfWeek
  """
}


{
  "error": {
    "details": "The complex aggregate expressions are not implemented yet: 1 + MAX(FlightDelayMin)",
    "type": "SqlFeatureNotImplementedException"
  },
  "status": 503
}
```

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
galkk: > MAX(age + 1),

This is quite different case, out of scope of this PR

> MAX(LOG(age))
surprisingly, this is **almost** working, I think it should be just fixed

Issue #294, opened 2019-11-19
Feature request: support SELECT DISTINCT


seraphjiang:
# Use case
We want to be able to use SELECT DISTINCT to find out list of unique item of specific field without duplicate data

Example SQL
```
SELECT DISTINCT(DestAirportID) FROM kibana_sample_data_flights order by DestAirportID
``` 

OD-SQL
```
GET _opendistro/_sql?format=csv
{
  "query": "select distinct(DestAirportID) from  kibana_sample_data_flights order by DestAirportID"
}

```

actual result(Partial)
***you could see, there is duplicate data in result***
```
DestAirportID
ABQ
ABQ
ADL
AICM
AICM
AICM
AICM
AICM
AICM
AICM
...
```

Expected result
***no duplicate data***
```
DestAirportID
ABQ
ADL
AICM
AMS
ARN
```

chloe-zh:
@seraphjiang Thanks for reporting. We've reproduced this issue:
```
POST _opendistro/_sql?format=csv
{
  "query": "select distinct(DestAirportID) from  kibana_sample_data_flights order by DestAirportID"
}
```
Output:
```
DestAirportID
ABQ
ABQ
ADL
AICM
AICM
AICM
AICM
...
```

We will investigate this problem and explore the solution to support the `DISTINCT` feature soon. Thanks!

dai-chen:
For now we only support `COUNT(DISTINCT)` because we rely on Elasticsearch aggregation to de-duplicate. Basically `SELECT DISTINCT` is equivalent to `GROUP BY` without aggregate function, so one option is to translate distinct to aggregation. In other database, `DISTINCT` and `GROUP BY` generates different physical plan, so it makes more sense to do this translation when generating DSL rather than by rewriter.

PR #300, opened 2019-11-21
Support DISTINCT feature in SELECT clause

chloe-zh:
*Issue #, if available:*

* [**Issue 294 Feature request: support SELECT DISTINCT**](https://github.com/opendistro-for-elasticsearch/sql/issues/294)

*Description of changes:*

* Supported DISTINCT in select clause with one and more fields
* Added tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
@@ -16,6 +16,7 @@
 package com.amazon.opendistroforelasticsearch.sql.domain;
 
 import com.alibaba.druid.sql.ast.SQLExpr;
+import com.alibaba.druid.sql.ast.expr.SQLAggregateOption;
 import com.amazon.opendistroforelasticsearch.sql.parser.ChildrenType;
 import com.amazon.opendistroforelasticsearch.sql.parser.NestedType;
 
@@ -38,12 +39,14 @@ public class Field implements Cloneable {
     private NestedType nested;
     private ChildrenType children;
     private SQLExpr expression;
+    private SQLAggregateOption option;
 
     public Field(String name, String alias) {
         this.name = name;
         this.alias = alias;
         this.nested = null;
         this.children = null;
+        this.option = null;
     }
 
     public Field(String name, String alias, NestedType nested, ChildrenType children) {
@@ -102,6 +105,14 @@ public class Field implements Cloneable {
             return null;
         }
         return this.children.childType;
+    }
+
+    public void setOption(SQLAggregateOption option) {
+        this.option = option;
+    }
+
+    public SQLAggregateOption getOption() {
+        return option;
     }
 
     @Override
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
@@ -35,11 +35,11 @@ public class Field implements Cloneable {
     public static final Field STAR = new Field("*", "");
 
     protected String name;
+    protected SQLAggregateOption option;
     private String alias;
     private NestedType nested;
     private ChildrenType children;
     private SQLExpr expression;
-    private SQLAggregateOption option;
 
     public Field(String name, String alias) {
         this.name = name;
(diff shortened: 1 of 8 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/AggregationIT.java
```diff
@@ -1184,6 +1184,23 @@ public void groupByScriptedHistogram() throws Exception {
         Assert.assertThat(result, containsString("\"script\":{\"source\""));
     }
 
+    @Test
+    public void distinctWithOneField() {
```
davidcui-amzn: what is the behavior if I want to run an aliased query (e.g SELECT DISTINCT name.lastname AS alias FROM ...)

chloe-zh: It works exactly the same with the proper GROUP BY, for example in the JDBC format there’s a row of “alias”:  “alias”

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
```diff
@@ -104,6 +107,14 @@ public String getChildType() {
         return this.children.childType;
     }
 
+    public void setOption(SQLAggregateOption option) {
```
galkk: consider renaming to setAggregateOption, because setOption doesn't tell much about content

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/AggregationOptionTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Unit test class for feature of aggregation options: DISTINCT, ALL, UNIQUE, DEDUPLICATION
+ */
+public class AggregationOptionTest {
+
+    @Test
+    public void distinctWithOneField() {
+        Assert.assertEquals(
+                explain("SELECT DISTINCT lastname FROM accounts"),
```
galkk: what will happen in case of user error?

SELECT DISTINCT(lastname) from accounts GROUP BY lastname?

chloe-zh: The DSL will look like `SELECT lastname, lastname from accounts GROUP BY lastname, lastname`

src/main/java/com/amazon/opendistroforelasticsearch/sql/domain/Field.java
chloe-zh: ok will do

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -57,6 +61,22 @@
     public Field makeField(SQLExpr expr, String alias, String tableAlias) throws SqlParseException {
         Field field = makeFieldImpl(expr, alias, tableAlias);
 
+        if (expr.getParent() != null && expr.getParent() instanceof SQLSelectItem
```
abbashus: Can we add a comment here on what exactly is the workaround is.

chloe-zh: Sure, will do!

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/AggregationOptionTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import static com.amazon.opendistroforelasticsearch.sql.util.SqlExplainUtils.explain;
+
+/**
+ * Unit test class for feature of aggregation options: DISTINCT, ALL, UNIQUE, DEDUPLICATION
+ */
+public class AggregationOptionTest {
```
dai-chen: I think test cases in this class is actually IT and should be moved to `AggregationIT`. To do UT, we can create AST by `SqlParserUtils.parse`, pass it to `makeField` and assert return value `Field`.

Issue #293, opened 2019-11-18
Automated Documentation - Basics


dai-chen:
This is the first sub issue for https://github.com/opendistro-for-elasticsearch/sql/issues/243 in which docs for interfaces and administration as well as foundation of Scaffold code will be added.

Issue #243, opened 2019-10-22
Documentation improvement


dai-chen:
Currently the SQL plugin documentation is only a README with reference to technical docs https://opendistro.github.io/for-elasticsearch-docs/docs/sql/. However, both are missing important details and not sufficient to be served as a good reference for user when they have trouble. Particularly, here are things missing and could be improved:

 1. **Description**: name and briefly describe its functionality.
 2. **Syntax**: rigor syntactical form in grammar.
 3. **Semantics**: semantical requirement, ex. specification for function.
 4. **Use Cases**: typical use case and example with response to help understand.
 5. **Explain**: ES DSL generated by our plugin for each sample query.
 6. **Constraints**: any case we don't support for now for some reason.

Apart from these items, we can also improve our documentation in other aspect, for example enable Wiki on GitHub to host our development docs in https://github.com/opendistro-for-elasticsearch/sql/tree/master/docs/dev. This is more focused on internal implementation as plugin developer and contributor facing design docs instead of user facing guidance. Because of its different intention, it is not supposed to have overlap with existing technical documentation.

dai-chen:
After scoping this work can be divided into 2 parts: docs for developer and for user. For the former, we plan to add more design docs and instructions for contributor manually so they can get involved easily.

As for user facing docs, we evaluated different options such as executable documentation and generation from code etc. At this stage our goal is to generate docs with small efforts and in sync with existing codebase as our architecture is tend to change. And we don't plan to use these docs as integration test for now.

Based on our requirements, we decide to add a set of "Scaffold" to automated docs generation from existing codebase. The generated docs still have the flexibility to be executable for testing and can be polished further by tech writer.

dai-chen:
Here is an incomplete list of what we want to cover in docs for user. 4 sub issues will be created soon.

* Interfaces
    * Endpoint
        * Query via GET/POST
        * Explain
    * JDBC protocol
        * Request & response format
        * Error response format
        * Prepared statement
    * Other formats
        * Default DSL
        * CSV
        * Raw
* Administration
    * Configuration
        * Plugin settings
    * Monitoring
        * Stats
* DQL (Data Query Language)
    * Basics
        * SELECT
        * FROM
        * WHERE
        * GROUP BY
        * HAVING
        * ORDER BY
        * LIMIT
    * Functions
        * SQL functions
        * ES special functions
    * Full text
    * Complex
        * PartiQL support
        * Subquery
        * Join
        * Multi-query
[...]

PR #302, opened 2019-11-23
Documentation for basic usage of plugin

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/293, https://github.com/opendistro-for-elasticsearch/sql/issues/243

*Description of changes:* This is the first PR for the documentation improvement work. In this PR, basic usage of our plugin is documented including endpoint, interfaces, plugin monitoring and settings. The generated documentation could be processed and polished further if needed.

*Documentation:* For what the documentation generated looks like, please refer to https://github.com/dai-chen/sql/blob/improve-documentation-basics/docs/user/index.rst


docs/user/interfaces/endpoint.rst
```diff
@@ -0,0 +1,108 @@
...
+	  },
+	  "_source" : {
+	    "includes" : [
+	      "firstname",
+	      "lastname"
+	    ],
+	    "excludes" : [ ]
+	  }
+	}
+
```
chloe-zh: Do we need to add the instruction of `PUT` endpoint?

docs/user/index.rst
```diff
@@ -0,0 +1,27 @@
+
```
chloe-zh: Will this manual entrance be put onto the README file? I think that would be an effective way for user to see refer to the manual.

docs/user/admin/monitoring.rst
```diff
@@ -0,0 +1,59 @@
...
+
+.. contents::
+   :local:
+   :depth: 1
+
+
+Introduction
+============
+
+By a stats endpoint, you are able to collect metrics for the plugin within the interval. Note that only node level statistics collecting is implemented for now. In other words, you only get the metrics for the node you're accessing. Cluster level statistics is yet to implement.
```
davidcui-amzn: Not sure how strict we want to be on correct grammar and semantics- but in the interest of insisting on the highest standards, the last sentence should be worded something like "Cluster level statistics have yet to be implemented."

docs/user/interfaces/endpoint.rst
davidcui-amzn: +1, I don't see a TODO for adding `PUT` or `DELETE` requests which I think should be included

docs/user/index.rst
dai-chen: Sure, I forgot to do that. :) Will update README. Thanks!

docs/user/interfaces/endpoint.rst
dai-chen: We only support `PUT` and `POST`: https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java#L83. User needs to `PUT` against Elasticsearch endpoint instead.

docs/user/admin/monitoring.rst
dai-chen: Sure, will do. Thanks!

dai-chen:
@aetter it'd be awesome if you can help take a quick look at the 4 documents in this PR, especially the description in each Section and Example. Because we're planning to push these generated docs to `for-elasticsearch-docs` repo later, it would save the trouble to come back to change to code again. Here is the [index page](https://github.com/dai-chen/sql/blob/improve-documentation-basics/docs/user/index.rst). Thanks in advance!

Btw, do you see any problem with these ReStructure Text docs (rather than Markdown) to be pushed to `for-elasticsearch-docs` later?

aetter:
Hi @dai-chen, yeah, I love reStructuredText, but its big appeal is for use with [Sphinx](http://www.sphinx-doc.org/) in large, harder-to-manage sets of documentation. If you're just rendering content on GitHub, it loses a lot of its appeal. Many directives don't work, no [cross-references](http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations), no [transclusion](http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment), no [conditional processing](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#including-content-based-on-tags), etc.

Our current documentation site uses Jekyll, which doesn't support RST out of the box. Even if we got RST working through a plugin, we'd have to add in the metadata that Jekyll uses to categorize/order/index the content at build time. You can see an example [here](https://raw.githubusercontent.com/opendistro/for-elasticsearch-docs/master/docs/sql/index.md).

So I suspect it would be a lot of effort, and in the end, we'd be left with a documentation site that's 95% Markdown and 5% RST. If the goal is to put these into the documentation site, I'd highly recommend switching to Markdown. [Pandoc](https://pandoc.org/) is super useful for moving between document formats. The syntax is something like:

```bash
pandoc myfile.rst -f rst -t markdown_github -o myfile.md
```

If the reason for choosing RST was that Markdown has some feature gaps, we use the [kramdown](https://kramdown.gettalong.org/syntax.html) parser, which closes a few of them (with some questionable syntax choices). You can get an autogenerated table of contents, for example, like so:

```
#### Table of contents
1. TOC
{:toc}
```

I don't see anything else in your content that uses RST-specific features, so I suspect the conversion to Markdown should be fairly clean if you decide to go that route.

dai-chen:
> Hi @dai-chen, yeah, I love reStructuredText, but its big appeal is for use with [Sphinx](http://www.sphinx-doc.org/) in large, harder-to-manage sets of documentation. If you're just rendering content on GitHub, it loses a lot of its appeal. Many directives don't work, no [cross-references](http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations), no [transclusion](http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment), no [conditional processing](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#including-content-based-on-tags), etc.
> 
> Our current documentation site uses Jekyll, which doesn't support RST out of the box. Even if we got RST working through a plugin, we'd have to add in the metadata that Jekyll uses to categorize/order/index the content at build time. You can see an example [here](https://raw.githubusercontent.com/opendistro/for-elasticsearch-docs/master/docs/sql/index.md).
> 
> So I suspect it would be a lot of effort, and in the end, we'd be left with a documentation site that's 95% Markdown and 5% RST. If the goal is to put these into the documentation site, I'd highly recommend switching to Markdown. [Pandoc](https://pandoc.org/) is super useful for moving between document formats. The syntax is something like:
> 
> ```shell
> pandoc myfile.rst -f rst -t markdown_github -o myfile.md
> ```
> 
> If the reason for choosing RST was that Markdown has some feature gaps, we use the [kramdown](https://kramdown.gettalong.org/syntax.html) parser, which closes a few of them (with some questionable syntax choices). You can get an autogenerated table of contents, for example, like so:
> 
> ```
> #### Table of contents
> 1. TOC
> {:toc}
> ```
> 
> I don't see anything else in your content that uses RST-specific features, so I suspect the conversion to Markdown should be fairly clean if you decide to go that route.

Hi @aetter, thanks for the information! The main reason (apart from its nicer looking) I use ReST is to keep the flexibility to integrate with other tool like Sphinx as you mentioned and Python DocTest etc. And yes, I found GitHub itself has very limited support so I avoid some directives for now :). Will convert and push Markdown version of the docs to `for-elasticsearch-docs` later. Thanks!

dai-chen:
@aetter Pandoc works for me. I'm sending PR https://github.com/opendistro/for-elasticsearch-docs/pull/146 for your review. Will merge this PR afterwards.

Issue #299, opened 2019-11-21
Column aliases not working for GROUP BY 


chloe-zh:
Example

```
POST _opendistro/_sql?format=jdbc
{
  "query": "select DestAirportID as dest from  kibana_sample_data_flights group by dest limit 5"
}
```

Output
```
{
  "schema": [],
  "total": 5,
  "datarows": [
    [],
    [],
    [],
    [],
    []
  ],
  "size": 5,
  "status": 200
}
```

Expected output
```
{
  "schema": [{
    "name": "DestAirportID",
    "alias": "dest",
    "type": "keyword"
  }],
  "total": 5,
  "datarows": [
    ["ZRH"],
    ["XIY"],
    ["YWG"],
    ["WAW"],
    ["VR10"]
  ],
  "size": 5,
  "status": 200
}
```

PR #307, opened 2019-11-27
Fix the issue of column alias not working for GROUP BY

chloe-zh:
*Issue #, if available:*

* [**Issue #299 Column aliases not working for GROUP BY**](https://github.com/opendistro-for-elasticsearch/sql/issues/299)

*Description of changes:*

* Fixed the alias of group by issue
* Fixed the shards size issue 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
@@ -86,6 +86,8 @@ public class AggregationQueryAction extends QueryAction {
                     // raw javascript dsl, so I'd like to scope the changes as of now to one particular fix for
                     // scripted functions
 
+                    // the condition `field.getName().equals("script")` is to include the CAST cases, since the cast
+                    // method is instance of MethodField with script. => corrects the shard size of CASTs
                     if (!(field instanceof MethodField) || field instanceof ScriptMethodField
                             || field.getName().equals("script")) {
                         //if limit size is too small, increasing shard  size is required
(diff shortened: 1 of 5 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/ExplainIT.java
```diff
@@ -61,6 +62,8 @@ public void searchSanity() throws IOException {
         Assert.assertThat(result.replaceAll("\\s+",""), equalTo(expectedOutput.replaceAll("\\s+","")));
     }
 
+    // This test was ignored because group by case function is not supported
+    @Ignore
```
abbashus: If we don't support it , this test case should be failing previously - right?

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
```diff
@@ -86,7 +86,8 @@ public SqlElasticSearchRequestBuilder explain() throws SqlParseException {
                     // raw javascript dsl, so I'd like to scope the changes as of now to one particular fix for
                     // scripted functions
 
-                    if (!(field instanceof MethodField) || field instanceof ScriptMethodField) {
+                    if (!(field instanceof MethodField) || field instanceof ScriptMethodField
+                            || field.getName().equals("script")) {
```
dai-chen: Need more context. So what case this change tries to handle?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/ExplainIT.java
chloe-zh: Yes, but I don't think the case should pass, there might be something wrong and need investigating.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/AggregationQueryAction.java
chloe-zh: Done, thanks!

PR #310, opened 2019-12-04
adding DATETIME cast support

davidcui-amzn:
*Issue #, if available:*
#268 
*Description of changes:*
Adding support for casting fields of different type (e.g Keyword, String) to Datetime type. 
Also refactored CAST() to be implemented as a function.


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -314,6 +317,47 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+        String[] response_str = new String[2];
+
+        for (int i = 0; i < response.getJSONArray("datarows").length(); ++i) {
+            response_str[i] = response.getJSONArray("datarows").getJSONArray(i).getString(0);
+        }
+        Arrays.sort(response_str);
+        Arrays.sort(expectedOutput);
+        for (int i = 0; i < response_str.length; ++i) {
+            Assert.assertThat(response_str[i], equalTo(expectedOutput[i]));
+        }
```
chloe-zh: I would suggest to exempt the sorts and put these two loops together instead. That looks more concise and readable.

davidcui-amzn: The sort is there because the order of the JSONArray outputs are not consistent on each run.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -314,6 +317,47 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+        String date_type_cast = "{\"name\":\"cast_date_keyword\",\"type\":\"date\"}";
+        assertEquals(response.getJSONArray("schema").get(0).toString(), date_type_cast);
+        String[] expectedOutput = new String[] {"Tue Aug 19 07:09:13", "Wed Sep 25 02:04:13"};
+        String[] response_str = new String[2];
+
+        for (int i = 0; i < response.getJSONArray("datarows").length(); ++i) {
+            response_str[i] = response.getJSONArray("datarows").getJSONArray(i).getString(0);
+        }
+        Arrays.sort(response_str);
+        Arrays.sort(expectedOutput);
```
galkk: why do you sort expected output?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -314,6 +317,47 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+        for (int i = 0; i < response_str.length; ++i) {
+            Assert.assertTrue(response_str[i].contains(expectedOutput[i]));
```
galkk: If you control the data, why not put expected output into the set, then it's going to be easy to check that the set will contain every value? (here and everywhere)

davidcui-amzn: Good tip, thanks- will look into it

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -314,6 +317,47 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+        }
+    }
+
+    @Test
+    public void castKeywordFieldToDatetimeWithAliasJdbcFormatTest() {
```
dai-chen: Could you add another test case for `CAST` in WHERE clause? Do we support it?

davidcui-amzn: Yes- will do

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -314,6 +319,66 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+        for (int i = 0; i < response.getJSONArray("datarows").length(); ++i) {
+            Assert.assertThat(response.getJSONArray("datarows").getJSONArray(i).getString(0)
+                    .indexOf(expectedOutput[i]), not(-1));
+        }
+    }
+
+    @Test
+    public void castFieldToDatetimeWithWhereClauseJdbcFormatTest() {
+        JSONObject response = executeJdbcRequest("SELECT CAST(date_keyword AS DATETIME) FROM "
+                + TestsConstants.TEST_INDEX_DATE + " WHERE date_keyword IS NOT NULL ORDER BY date_keyword");
```
dai-chen: Could you use `CAST` in WHERE, for example `WHERE CAST(date_keyword AS DATETIME) = '2019-12-09 ...'`?

davidcui-amzn: Ah yes, I misunderstood your original comment. Will add.

davidcui-amzn: This is currently not supported, will have to add this functionality.

dai-chen: Okay. Do you know the reason? Because normally functions that we support should be able to work in SELECT and WHERE.

davidcui-amzn: Not yet- I'm currently debugging.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
```diff
@@ -405,4 +405,14 @@ public MethodField makeMethodField(String name, List<SQLExpr> arguments, SQLAggr
             return new MethodField(name, paramers, option, alias);
         }
     }
+
+    public ScriptMethodField makeCastMethodField(SQLCastExpr castExpr) throws SqlParseException {
```
dai-chen: I wonder what's the intention of this method because I noticed it seems already be covered by `makeField`: https://github.com/opendistro-for-elasticsearch/sql/blob/d3e6477b6e9bea8acf498ae8ca23bdfa69f3c257/src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java#L121

The only difference I notice is that boolean passed to `CastParser.parse()` method. Is all the code change in `WhereParser` due to this new method? Any chance to avoid those changes and reuse the logic in `makeField` above?

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CastParser.java
```diff
@@ -63,8 +63,8 @@ public String parse(boolean isReturn) throws SqlParseException {
             } else if (DataType.valueOf(dataType) == DataType.STRING) {
                 result.add(String.format("def %s = %s.toString()", name, fileName));
             } else if (DataType.valueOf(dataType) == DataType.DATETIME) {
-                result.add(String.format("def %s = new Date(Double.parseDouble(%s.toString()).longValue())",
-                        name, fileName));
+                result.add(String.format("def %s = new SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")"
+                        + ".parse(%s.toString())", name, fileName));
```
dai-chen: So if the format of field value is slightly different from `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'`, ex. `2020-01-10T10:00:00Z`, do we support this case? It would be great if we can support several popular formats because this `CAST` function is not like `STR_TO_DATE` function which can specify format.

Here is what MySQL can support:

```
>create table test (name varchar(255));
>insert into test values ('2020-01-10T10:00:00.000Z');
>insert into test values ('2020-01-10T11:00:00Z');
>insert into test values ('2020-01-11');
>insert into test values ('2020-01-10 12:00:00');
>
>select cast(name as datetime) from test;
cast(name as datetime)
2020-01-10 10:00:00
2020-01-10 11:00:00
2020-01-11 00:00:00
2020-01-10 12:00:00
```

davidcui-amzn: Currently we only support `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` format. I can create an issue to support different formats but I feel that should not be a part of this PR

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
davidcui-amzn: Yes, that is all the code change in `WhereParser`. I think it is possible but it would require adding a new parameter to `makeField` and `makeFieldImpl` to account for the changed boolean passed. Since those methods are used in many different places and this case is only needed for this, I chose to make a new method, but can refactor if it's deemed the better option.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/CastParser.java
dai-chen: Sounds good! Please also mention `STR_TO_DATE` function in the new issue. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/FieldMaker.java
dai-chen: I may need to look into the code but I'm wondering how other functions handle this case of being called in SELECT and WHERE. I can see a boolean parameter in `SQLFunctions.function()` and called by `makeField` with different value.

davidcui-amzn: I will step through the code with some other functions to see if there's a way for me to eliminate the need for that method

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/WhereParser.java
```diff
@@ -522,11 +523,26 @@ private MethodField parseSQLMethodInvokeExprWithFunctionInWhere(SQLMethodInvokeE
...
+    }
+
     private SQLMethodInvokeExpr parseSQLBinaryOpExprWhoIsConditionInWhere(SQLBinaryOpExpr soExpr)
             throws SqlParseException {
 
         if (!(soExpr.getLeft() instanceof SQLMethodInvokeExpr
-                || soExpr.getRight() instanceof SQLMethodInvokeExpr)) {
+                || soExpr.getRight() instanceof SQLMethodInvokeExpr)
+                && !(soExpr.getLeft() instanceof SQLCastExpr
+                    || soExpr.getRight() instanceof SQLCastExpr)) {
```
dai-chen: Could you refactor this complex condition expression? ex. `bothSidesAreNotFunction() && bothSizeAreNotCast()`

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -929,15 +944,39 @@ private String upper(String property, String culture) {
         return new Tuple<>(name, def(name, resultStr));
     }
 
+    private String getCastScriptStatement(String name, String castType, List<KVValue> paramers) {
+        String fileName = String.format("doc['%s'].value", paramers.get(0).toString());
```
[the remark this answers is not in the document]
davidcui-amzn: `paramers.get(0).toString()` is the `fieldName` in this case - should I make a separate `fieldName` variable instead?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -1209,6 +1209,7 @@ public void caseWhenSwitchTest()  {
         );
     }
 
+    @Ignore
```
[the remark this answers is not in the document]
davidcui-amzn: Stepping through after my code changes, looks like these UTs do test against `SQLFunctions.function` and `SQLFunctions.cast` correctly

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -1313,13 +1316,15 @@ public void castToDateTimeTest() throws Exception {
...
-        Assert.assertTrue(scriptCode.contains("new Date(Double.parseDouble(doc['age'].value.toString()).longValue())"));
+        Assert.assertTrue(scriptCode.contains("new SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")"
+                + ".parse(doc['age'].value.toString())"));
     }
 
     @Test
     public void castToDoubleThenDivideTest() throws Exception {
         String query = "select cast(age as double)/2 from "+ TestsConstants.TEST_INDEX_ACCOUNT + "/account limit 10";
         SQLExpr sqlExpr = queryToExpr(query);
+        System.out.println(sqlExpr);
```
penghuo: Remove it.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/parser/SqlParserTest.java
```diff
@@ -1328,6 +1333,7 @@ public void castToDoubleThenDivideTest() throws Exception {
         Assert.assertEquals("script",castField.getName());
 
         String scriptCode = (String) methodField.getParams().get(1).value;
+        System.out.println(scriptCode);
```
penghuo: Ditto.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/SQLFunctionsTest.java
```diff
@@ -27,7 +28,7 @@
 public class SQLFunctionsTest {
 
     @Test
-    public void testAssign() {
+    public void testAssign() throws SqlParseException {
```
penghuo: why add the exception?

davidcui-amzn: It is part of `SQLFunctions.function`

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -316,6 +318,149 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+    @Test
+    public void castKeywordFieldToDatetimeWithoutAliasJdbcFormatTest() {
+        JSONObject response = executeJdbcRequest("SELECT CAST(date_keyword AS DATETIME) FROM "
+                + TestsConstants.TEST_INDEX_DATE + " ORDER BY date_keyword");
+
+        String date_type_cast = "{\"name\":\"cast_date_keyword\",\"type\":\"date\"}";
+        assertEquals(response.getJSONArray("schema").get(0).toString(), date_type_cast);
+        String[] expectedOutput = new String[] {"Tue Aug 19 07:09:13", "Wed Sep 25 02:04:13"};
+
+        for (int i = 0; i < response.getJSONArray("datarows").length(); ++i) {
```
penghuo: The assert in for-loop may not executed when the array is empty. Refer the implementation of https://github.com/opendistro-for-elasticsearch/sql/blob/3730ac6e9b3a363c51432b34d2104567eb9babb7/src/test/java/com/amazon/opendistroforelasticsearch/sql/util/MatcherUtils.java#L144

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -84,7 +88,7 @@
             "if", "ifnull", "isnull"
     );
 
-    private static final Set<String> utilityFunctions = Sets.newHashSet("field", "assign");
+    private static final Set<String> utilityFunctions = Sets.newHashSet("field", "assign", "cast");
```
penghuo: minor. using ImmutableSet if possible.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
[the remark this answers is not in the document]
dai-chen: I meant is `fileName` a typo?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -929,15 +945,41 @@ private String upper(String property, String culture) {
...
+                return String.format("def %s = Double.parseDouble(%s.toString()).floatValue()", name, fileName);
+            case "DOUBLE":
+                return String.format("def %s = Double.parseDouble(%s.toString()).doubleValue()", name, fileName);
+            case "STRING":
+                return String.format("def %s = %s.toString()", name, fileName);
+            case "DATETIME":
+                return String.format("def %s = new SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")"
+                    + ".parse(%s.toString())", name, fileName);
+            default:
+                throw new SqlParseException("Unsupported cast type");
```
dai-chen: np: add `castType` to error message.

davidcui-amzn: Good idea- will add

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -929,15 +945,42 @@ private String upper(String property, String culture) {
...
+                return String.format("def %s = Double.parseDouble(%s.toString()).longValue()", name, castFieldName);
+            case "FLOAT":
+                return String.format("def %s = Double.parseDouble(%s.toString()).floatValue()", name, castFieldName);
+            case "DOUBLE":
+                return String.format("def %s = Double.parseDouble(%s.toString()).doubleValue()", name, castFieldName);
+            case "STRING":
+                return String.format("def %s = %s.toString()", name, castFieldName);
+            case "DATETIME":
+//                return String.format("def %s = new SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")"
+//                    + ".parse(%s.toString())", name, castFieldName);
```
penghuo: remove it

davidcui-amzn: done

penghuo:
LGTM, Thanks.

Issue #131, opened 2019-07-24
Improve DELETE functionality.


abbashus:
Currently, the SQL DELETE query does not give proper results to the user. 

Data:

```
PUT bank/_doc/1
{
  "name" : "Abbas",
  "state" : "WA"
}

PUT bank/_doc/2
{
  "name" : "John",
  "state" : "CA" 
}
```
Vanilla DELETE query:

```
POST _opendistro/_sql/_explain?
{
    "query" : "DELETE FROM bank" 
}

==== Explain: ====
{"size":1000,"query":{"match_all":{"boost":1.0}},"_source":false}

==== ES raw output: ====
{
  "took" : 84,
  "timed_out" : false,
  "total" : 2,
  "updated" : 0,
  "created" : 0,
  "deleted" : 2, 
  "batches" : 1,
  "version_conflicts" : 0,
  "noops" : 0,
  "retries" : {
    "bulk" : 0,
    "search" : 0
  },
  "throttled_millis" : 0,
  "requests_per_second" : -1.0,
  "throttled_until_millis" : 0,
  "failures" : [ ]
}

==== _opendistro/_sql?_format=jdbc ====
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "class com.amazon.opendistroforelasticsearch.sql.domain.Delete cannot be cast to class com.amazon.opendistroforelasticsearch.sql.domain.Select (com.amazon.opendistroforelasticsearch.sql.domain.Delete and com.amazon.opendistroforelasticsearch.sql.domain.Select are in unnamed module of loader java.net.FactoryURLClassLoader @46963479)",
    "type": "ClassCastException"
  },
  "status": 500
}

==== _opendistro/_sql?format=csv ====
null 

```

DELETE query with condition(s):

```
POST _opendistro/_sql/_explain?
{
    "query" : "DELETE FROM bank WHERE name = 'Abbas' " 
}

{
  "size": 1000,
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "name": {
              "value": "Abbas",
              "boost": 1
            }
          }
        }
      ],
      "adjust_pure_negative": true,
      "boost": 1
    }
  },
  "_source": false
}
```

**Observation:**

- Even though the actual data is deleted , output of  `format=jdbc` emits exception and for `format=csv` it just outputs  `null` which is not very informative.
[...]

PR #337, opened 2020-01-08
Fix JDBC response for delete query

abbashus:
*Issue*: #131 

*Description of changes:*
Fixes the JDBC output format of delete query

Previously
```
==== _opendistro/_sql?_format=jdbc ====
{
  "error": {
    "details": "class com.amazon.opendistroforelasticsearch.sql.domain.Delete cannot be cast to class com.amazon.opendistroforelasticsearch.sql.domain.Select (com.amazon.opendistroforelasticsearch.sql.domain.Delete and com.amazon.opendistroforelasticsearch.sql.domain.Select are in unnamed module of loader java.net.FactoryURLClassLoader @46963479)",
    "type": "ClassCastException"
  },
  "status": 500
}
```

Now
```
{
  "schema": [{
    "name": "deleted_rows",
    "type": "long"
  }],
  "total": 1,
  "datarows": [[1]],
  "size": 1,
  "status": 200
}

```


src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/DeleteResultSet.java
```diff
@@ -0,0 +1,51 @@
...
+    public DeleteResultSet(Client client, Delete query, Object queryResult) {
+        this.client = client;
+        this.query = query;
+        this.queryResult = queryResult;
+        this.schema = new Schema(loadColumns());
+        this.dataRows = new DataRows(loadRows());
+    }
+
+    private List<Schema.Column> loadColumns() {
+        return Collections.singletonList(new Schema.Column(DELETED, null, Schema.Type.LONG));
```
chloe-zh: What if the data type is not `LONG`? Do we need to retrieve the data type from the schema rather than hard coding?

abbashus: It will always be being [long](https://github.com/elastic/elasticsearch/blob/7a622f024f555566ab06176f85b830baf47e61c0/server/src/main/java/org/elasticsearch/index/reindex/BulkByScrollResponse.java#L1310). We can change this if there is a breaking change for future ODFE release.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DeleteIT.java
```diff
@@ -77,12 +68,76 @@ public void deleteWithConditionTest() throws IOException, InterruptedException {
...
+        String deleteQuery = StringUtils.format("DELETE FROM %s/account", TestsConstants.TEST_INDEX_ACCOUNT);
+
+        response = new JSONObject(executeQuery(deleteQuery, "jdbc"));
+        System.out.println(response);
+        assertThat(response.query("/schema/0/name"), equalTo("deleted_rows"));
+        assertThat(response.query("/schema/0/type"), equalTo("long"));
+        assertThat(response.query("/datarows/0/0"), equalTo(totalHits));
+        assertThat(response.query("/total"), equalTo(1));
+        assertThat(response.query("/status"), equalTo(200));
+        assertThat(response.query("/size"), equalTo(1));
```
dai-chen: Should assertion for `total` and `size` be `equalTo(totalHits)`? And any reason not to use `total` or `size` as deleted rows directly?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/DeleteResultSet.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.amazon.opendistroforelasticsearch.sql.domain.Delete;
+import org.elasticsearch.client.Client;
+import org.elasticsearch.index.reindex.BulkByScrollResponse;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class DeleteResultSet extends ResultSet {
```
dai-chen: Probably not necessary for this data class. But it would be better to add a UT for test coverage.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DeleteIT.java
abbashus: Yes, we could do it. I made my changes based on the solution proposed #131 .  Based on your recommendation should the response look like the following:

```
{
  "schema": [],
  "total": <number of deleted rows> ,
  "datarows": [],
  "size": <number of deleted rows> >,
  "status": 200
}
```

Should we get rid of `schema` and `datarows` altogether?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/DeleteResultSet.java
abbashus: Sure, will add unit tests.

dai-chen: Thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DeleteIT.java
dai-chen: Yeah, maybe. Or just leave them empty if any issue.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DeleteIT.java
```diff
@@ -77,12 +68,76 @@ public void deleteWithConditionTest() throws IOException, InterruptedException {
...
+
+        response = executeRequest(makeRequest(selectQuery));
+        assertThat(getTotalHits(response), equalTo(0));
+
+        // Multiple invocation of delete query should return deleted == 0
+        response = new JSONObject(executeQuery(deleteQuery, "jdbc"));
+        assertThat(response.query("/datarows/0/0"), equalTo(0));
+    }
+
+    @Test
```
davidcui-amzn: Perhaps not necessary- but I think it could be beneficial to add 1-2 more variety of queries to test with

abbashus: Can you provide examples of what you kind of variety you are expecting.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DeleteIT.java
abbashus: I realized , we need the `schema` with` deleted_rows` column, because when converting to `csv` we don't emit the `total` and `size` values. Since we plan to derive the `csv` output directly from the common representation (currently used for `jdbc` format), in absence of `deleted_rows` column , the `csv` output would be empty.

dai-chen: Sure, let's follow your current format. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/DeleteResultSet.java
abbashus: Added unit tests.

PR #340, opened 2020-01-10
Fix the issue in semantic analyzer.

penghuo:
1. Change the return Type of the LOG function.
2. Add the support of CAST function.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/visitor/TypeChecker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/visitor/TypeChecker.java
@@ -174,6 +174,11 @@ public class TypeChecker implements GenericSqlParseTreeVisitor<Type> {
     }
 
     @Override
+    public Type visitConvertedType(String text) {
+        return ESDataType.typeOf(text);
+    }
+
+    @Override
     public Type defaultValue() {
         return NULL_TYPE;
     }
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/GenericTypeTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/GenericTypeTest.java
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.sql.antlr.semantic.types;
 
 import org.junit.Test;
 
+import static com.amazon.opendistroforelasticsearch.sql.antlr.semantic.types.base.ESDataType.DOUBLE;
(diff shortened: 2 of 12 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/visitor/TypeChecker.java
```diff
@@ -173,6 +173,11 @@ public Type visitNull() {
         return UNKNOWN;
     }
 
+    @Override
+    public Type visitConvertedType(String text) {
+        return ESDataType.typeOf(text);
```
dai-chen: David's PR reminds me that some type in `convertedDataType` doesn't have corresponding type in `ESDateType`, such as `DATETIME` and `INT`. Could you verify if it's an issue or not? I guess `UNKNOWN` is returned in this case. It is fine for type checking but it may be not what you expect.

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/GenericTypeTest.java
```diff
@@ -34,17 +35,17 @@
...
     public void passIntegerArgToLogShouldReturnInteger() {
-        assertEquals(INTEGER, LOG.construct(singletonList(INTEGER)));
+        assertEquals(DOUBLE, LOG.construct(singletonList(INTEGER)));
     }
 
     @Test
     public void passLongArgToLogShouldReturnLong() {
-        assertEquals(LONG, LOG.construct(singletonList(LONG)));
+        assertEquals(DOUBLE, LOG.construct(singletonList(LONG)));
     }
```
davidcui-amzn: is there a need for test cases going the other way around (`e.g assertEquals(LONG, LOG.construct(singletonList(DOUBLE))`) ?

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/visitor/TypeChecker.java
penghuo: Sure, I will add the default type for UNKNOWN use cases when applying the return type.

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/GenericTypeTest.java
penghuo: Done. Change the test case name to passLongArgToLogShouldReturnDouble.

Issue #341, opened 2020-01-10
Functions in WHERE Clause do not work with IS/IS NOT


davidcui-amzn:
Functions cannot be called in the WHERE clause with IS/IS NOT.

Sample query:
```
GET _opendistro/_sql
{
  "query": """
  SELECT dayOfWeek FROM kibana_sample_data_flights WHERE (ABS(DistanceKilometers) IS NOT NULL)
  """
}
```
Returns the error message:
```
{
  "error": {
    "root_cause": [
      {
        "type": "script_exception",
        "reason": "compile error",
        "script_stack": [
          "... ilometers'].value);abs_1 IS NOT java.lang.Object@2 ...",
          "                             ^---- HERE"
        ],
        "script": "def abs_1 = Math.abs(doc['DistanceKilometers'].value);abs_1 IS NOT java.lang.Object@243d59fd",
        "lang": "painless"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "kibana_sample_data_flights",
        "node": "3ZdEGtaiRt2uUzeNPRmNZQ",
        "reason": {
          "type": "query_shard_exception",
          "reason": "failed to create query: {\n  \"bool\" : {\n    \"filter\" : [\n      {\n        \"bool\" : {\n          \"must\" : [\n            {\n              \"script\" : {\n                \"script\" : {\n                  \"source\" : \"def abs_1 = Math.abs(doc['DistanceKilometers'].value);abs_1 IS NOT java.lang.Object@243d59fd\",\n                  \"lang\" : \"painless\"\n                },\n                \"boost\" : 1.0\n              }\n            }\n          ],\n          \"adjust_pure_negative\" : true,\n          \"boost\" : 1.0\n        }\n      }\n    ],\n    \"adjust_pure_negative\" : true,\n    \"boost\" : 1.0\n  }\n}",
          "index_uuid": "zPAMxE3-QzaMFke6NhCxSg",
          "index": "kibana_sample_data_flights",
          "caused_by": {
            "type": "script_exception",
            "reason": "compile error",
            "script_stack": [
              "... ilometers'].value);abs_1 IS NOT java.lang.Object@2 ...",
              "                             ^---- HERE"
            ],
[...]

Issue #308, opened 2019-11-28
"all shards fail" error when functions/operators are applied to null-value field


chloe-zh:
Example index:
```
{"index":{"_index": "people", "_id":"1"}}
{"account_number":1,"balance":39225,"firstname":"Amber","lastname":"Duke","age":32,"gender":"M"}
{"index":{"_index": "people", "_id":"2"}}
{"account_number":6,"balance":5686,"firstname":"Hattie","lastname":"Bond","age":36,"gender":"M"}
{"index":{"_index": "people", "_id":"3"}}
{"account_number":13,"balance":32838,"firstname":"Nanette","age":28,"gender":"F"}
{"index":{"_index": "people", "_id":"4"}}
{"account_number":18,"firstname":"Dale","lastname":"Adams","age":33,"gender":"M"}
```

Example query 1:
```
SELECT abs(balance) FROM people
```

Expected result:
```
abs(balance)
39225
5686
32838
NULL
```

Actual output:
```
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "all shards failed",
    "type": "SearchPhaseExecutionException"
  },
  "status": 500
}
```

Example query 2:
```
SELECT substring(lastname, 1, 3) FROM people
```

Expected result:
```
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "all shards failed",
    "type": "SearchPhaseExecutionException"
  },
  "status": 500
}
```

Actual result:
```
substring(lastname, 1, 3)
Duk
Bon
Null
Ada
```

chloe-zh:
Related: #320

dai-chen:
This has been taken care of in new query engine: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/dql/expressions.rst#null-handling

PR #344, opened 2020-01-13
Function support for IS/IS NOT in WHERE clause

davidcui-amzn:
*Issue #, if available:*
#341 
*Description of changes:*
Adding support for usage of functions with the IS/IS NOT keywords in the WHERE clause.

Example query: 
` SELECT dayOfWeek FROM kibana_sample_data_flights WHERE ABS(balance) IS NOT NULL`


src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/WhereParser.java
```diff
@@ -576,8 +576,10 @@ private SQLMethodInvokeExpr parseSQLBinaryOpExprWhoIsConditionInWhere(SQLBinaryO
 
         String operator = soExpr.getOperator().getName();
 
-        if (operator.equals("=")) {
+        if (operator.equals("=") || operator.equals("IS")) {
             operator = "==";
+        } else if (operator.equals("IS NOT")) {
```
anirudha: do we maintain enums for these string literals ? 
eg. IS / IS NOT
@davidcui-amzn @dai-chen @penghuo

dai-chen: > do we maintain enums for these string literals ?
> eg. IS / IS NOT
> @davidcui-amzn @dai-chen @penghuo

There is enum we added in type system in new ANTLR parser: https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/types/operator/ComparisonOperator.java#L37. But I guess nothing in the old codebase.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -310,6 +310,66 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
     }
 
+    @Test
+    public void functionAbsInWhereClauseWithIsNullTest() {
+        JSONObject response = executeJdbcRequest(
+                "SELECT age FROM " + TEST_INDEX_ACCOUNT + " WHERE (ABS(balance) IS NULL)"
+        );
+
+        String schema_result = "{\"name\":\"age\",\"type\":\"long\"}";
+        assertEquals(response.getJSONArray("schema").get(0).toString(), schema_result);
```
abbashus: np: I found that we can make it more concise and readable like [this](https://github.com/opendistro-for-elasticsearch/sql/blob/179429ad06d241d676a9a6a1eb93038d5ad8be0a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/DeleteIT.java#L123) using `JSONObject.query()` here and elsewhere.

src/main/java/com/amazon/opendistroforelasticsearch/sql/parser/WhereParser.java
dai-chen: @davidcui-amzn please check if case is ignored or not, ex `is` and `is not`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
davidcui-amzn: Good suggestion! Will make that change

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -310,6 +310,66 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
         }
     }
 
+    @Test
+    public void functionAbsInWhereClauseWithIsNullTest() {
+        JSONObject response = executeJdbcRequest(
+                "SELECT age FROM " + TEST_INDEX_ACCOUNT + " WHERE (ABS(balance) IS NULL)"
```
dai-chen: So all your IT use `TEST_INDEX_ACCOUNT ` index for testing, I wonder is there any document in this index missing `balance` or `firstname` field? Because I recall the painless scripts we generate are unable to handle missing field (https://github.com/opendistro-for-elasticsearch/sql/issues/308), I assume your IT actually doesn't cover the typical use of `IS NULL` or `IS NOT NULL`.

davidcui-amzn: Good point- I'll look for/create an index with some missing field values

penghuo: One suggestion, instead of create new index, could we consider using the existing index. I think there are already 10+ different indices in place.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLFunctionsIT.java
```diff
@@ -316,6 +318,71 @@ public void castIntFieldToDoubleWithAliasJdbcFormatGroupByTest() {
...
+        assertThat(response.query("/schema/0/type"), equalTo("text"));
+
+//        Integer[] expectedOutput = new Integer[] {40, 39, 38, 37, 36};
+        System.out.println(response.getJSONArray("datarows").length());
+        for (int i = 0; i < response.getJSONArray("datarows").length(); ++i) {
+//            Assert.assertThat(
+//                    response.getJSONArray("datarows")
+//                            .getJSONArray(i).get(0),
+//                    equalTo(expectedOutput[i]));
+            System.out.println(response.getJSONArray("datarows").getJSONArray(i).getString(0));
```
penghuo: Please remove it.

Issue #320, opened 2019-12-11
Exception handling


anirudha:
tbd

dai-chen:
Related: https://github.com/opendistro-for-elasticsearch/sql/issues/184

dai-chen:
Probably related: https://github.com/opendistro-for-elasticsearch/sql/issues/193

PR #345, opened 2020-01-13
Syntax and semantic exceptions handling for unsupported features

chloe-zh:
*Issue #, if available:*

* #320 

*Description of changes:*

Added an extra filter in parser to stop the queries with unsupported features going down to the ES engine, the behaviors are throwing `SqlFeatureNotImplementedException` and sent proper error messages to the users.

* Sample data: kibana_sample_data_flights

* Nested function example:
```
POST _opendistro/_sql
{
  "query": "select abs(max(AvgTicketPrice)) from kibana_sample_data_flights"
}

/** Response: */
{
  "error": {
    "details": "Nested function calls with aggregation argument like [abs(max(AvgTicketPrice))] are not supported yet",
    "type": "SqlFeatureNotImplementedException"
  },
  "status": 503
}
```

* Nested aggregate examples: 
```
/** Function with aggregation argument example: */
POST _opendistro/_sql
{
  "query": "select abs(max(age)) from accounts"
}

/** Response: */
{
  "error": {
    "details": "Nested function calls with aggregation argument like [abs(max(age))] are not supported yet",
    "type": "SqlFeatureNotImplementedException"
  },
  "status": 503
}

/** Aggregate with function aggregator example: */
POST _opendistro/_sql
{
  "query": "select max(abs(AvgTicketPrice)) from kibana_sample_data_flights"
}

/** Response */
{
  "error": {
    "details": "Aggregation calls with function aggregator like [max(abs(AvgTicketPrice))] are not supported yet",
    "type": "SqlFeatureNotImplementedException"
  },
  "status": 503
}
```

* Unsupported feature example:
```
POST _opendistro/_sql
{
  "query": "select AvgTicketPrice DIV DistanceMiles from kibana_sample_data_flights"
}

/** Response: */
{
  "error": {
    "details": "Operator [DIV] is not supported yet",
    "type": "SqlFeatureNotImplementedException"
  },
  "status": 503
}
```


src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerScalarFunctionTest.java
```diff
@@ -95,6 +96,7 @@ public void logFunctionCallWithUnknownFieldShouldPass() {
         validate("SELECT LOG(new_field) FROM semantics");
     }
 
+    @Ignore /** nested functions are blocked by throwing SqlFeatureNotImplementedException yet before implemented */
```
davidcui-amzn: I'm not sure I completely understand the explanation from the comment- are you saying that these tests throw the `SqlFeatureNotImplementedException` too quickly currently, which is why so many of them are set to `@Ignore`?

chloe-zh: @davidcui-amzn Yes, how shall I put it to make it clear?

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerOperatorTest.java
```diff
@@ -55,9 +55,9 @@ public void compareSubstringFunctionCallEqualsToNumberShouldFail() {
     }
 
     @Test
-    public void compareLogAndAbsFunctionCallWithIntegerSmallerThanStringShouldFail() {
+    public void compareLogFunctionCallWithIntegerSmallerThanStringShouldFail() {
```
penghuo: Could you add the exception expectation?

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerScalarFunctionTest.java
```diff
@@ -104,21 +106,25 @@ public void substringWithLogFunctionCallWithUnknownFieldShouldPass() {
                 );
     }
 
+    @Ignore /** nested functions are blocked by throwing SqlFeatureNotImplementedException yet before implemented */
```
penghuo: Why ignore the test cases? Should we add the exception assert?

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitorTest.java
```diff
@@ -74,6 +74,42 @@ public void selectStarShouldReturnEmptyProductAsQueryVisitingResult() {
         Assert.assertTrue(result.isCompatible(new Product(emptyList())));
     }
 
+    @Test(expected = SqlFeatureNotImplementedException.class)
```
penghuo: Could you also add the exception message assert?

src/main/antlr/OpenDistroSqlParser.g4
```diff
@@ -324,20 +336,29 @@ aggregateWindowedFunction
     | COUNT '(' aggregator=DISTINCT functionArgs ')'
     ;
 
+functionAsAggregatorFunction
+    : (AVG | MAX | MIN | SUM)
```
penghuo: COUNT?

src/main/antlr/OpenDistroSqlParser.g4
```diff
@@ -324,20 +336,29 @@ aggregateWindowedFunction
...
+      '(' aggregator=(ALL | DISTINCT)? functionCall ')'
+    ;
+
 scalarFunctionName
     : functionNameBase
     ;
 
 functionArgs
-    : (constant | fullColumnName | functionCall | expression)
+    : (constant | fullColumnName | expression)
```
penghuo: why remove the functionCall from args?

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerOperatorTest.java
chloe-zh: Ok will do.

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerScalarFunctionTest.java
chloe-zh: That was because these cases are not valid for now since the plugin does not support these features so far, but they are remained here for future implementation. 
UTs and ITs are added in `AntlrSqlParseTreeVisitorTest.java` and `QueryAnalysis.java` instead.

src/main/antlr/OpenDistroSqlParser.g4
chloe-zh: `functionCall` is moved to `nestedFunctionArgs`.

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerScalarFunctionTest.java
davidcui-amzn: I think it is mostly clear- maybe just add something at the beginning such as `Test set to ignore due to nested functions blocked by...`

src/main/antlr/OpenDistroSqlParser.g4
chloe-zh: Added now. Thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitorTest.java
chloe-zh: Will do, thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerScalarFunctionTest.java
chloe-zh: Great, have changed them. Thanks!

chloe-zh: Great, have changed them. Thanks!

src/main/antlr/OpenDistroSqlParser.g4
```diff
@@ -324,20 +336,30 @@ aggregateWindowedFunction
...
+      '(' aggregator=(ALL | DISTINCT)? functionCall ')'
+    | COUNT '(' aggregator=(ALL | DISTINCT)? functionCall ')'
+    ;
+
 scalarFunctionName
     : functionNameBase
     ;
 
 functionArgs
-    : (constant | fullColumnName | functionCall | expression)
```
dai-chen: As my understanding, the changes are temporary and may be reverted later. Could you comment it out rather than remove it and put new rules you added together (probably better with comments explaining why)? Just in case someone else would make change on top of yours and hard to track only by commit history.

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
```diff
@@ -217,11 +231,37 @@ public T visitUdfFunctionCall(UdfFunctionCallContext ctx) {
...
+     * nested function calls and aggregations (b&c). Besides, the math constants included in the [mathConstants] set
+     * are regraded as scalar functions, but they are working well in the painless script.
+     *
+     * Thus, the types of functions to throw exceptions:
+     * (I)case (b) except that the arguments are from the [mathConstants] set;
+     * (II) case (b) except that the arguments are from the [supportedNestedFunctions] set;
+     * (III) case (c) except that the aggregators are from thet [supportedNestedFunctions] set.
+     */
     @Override
     public T visitScalarFunctionCall(ScalarFunctionCallContext ctx) {
```
dai-chen: Same suggestion for the new code you're adding here: could you move them to another class, ex. `UnsupportedFunction`?

Btw, could you explore the possibility of intercepting and checking what you're interested in an implementation class of `GenericSqlParseTreeVisitor` without touching grammar file? I think at least it's doable for unsupported function and operator which can help you minimize the changes.

src/main/antlr/OpenDistroSqlParser.g4
chloe-zh: Ok that makes sense, will do. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
chloe-zh: Ok will do!

chloe-zh: It's done now. Thanks!

src/test/java/com/amazon/opendistroforelasticsearch/sql/antlr/semantic/SemanticAnalyzerScalarFunctionTest.java
```diff
@@ -95,6 +96,7 @@ public void logFunctionCallWithUnknownFieldShouldPass() {
         validate("SELECT LOG(new_field) FROM semantics");
     }
 
+    @Ignore /** Test set to ignore due to nested functions not supported and blocked by throwing SqlFeatureNotImplementedException */
```
dai-chen: I think you can put the comment into `reason` field, ex.`@Ignore("Test set ...")`

src/main/java/com/amazon/opendistroforelasticsearch/sql/antlr/visitor/AntlrSqlParseTreeVisitor.java
dai-chen: Thanks for the changes! And we can make it better. Sorry I didn't make it clear at first.

```
class AntlrSqlParseTreeVisitor {
    public T visitScalarFunctionCall(ScalarFunctionCallContext ctx) {
        xxx.verify(ctx);
        T func = visit(ctx.scalarFunctionName());
        return reduce(func, ctx.functionArgs());
    }

    public T visitMathOperator(MathOperatorContext ctx) {
        xxx.verify(ctx);
        return super.visitMathOperator(ctx);
    }
    ...
}

class XXX { // Need a good name for this abstraction (semantics that we don't support for now).
    private static final Set<String> mathConstants, unsupportedFunctions ...

    public void verify(ScalarFunctionCallContext ctx) {
        ...
    }

    public void verify(MathOperatorContext ctx) {
        if (unsupportedOperators.contains(StringUtils.toLower(ctx.getText())))
            throw new SqlFeatureNotImplementedException(...);
    }

   ...
}
```

Issue #292, opened 2019-11-18
multiple function calls in aggregation queries


anirudha:
eg. enabling the following query, in exploratory/design stage

SELECT DestCountry, dayOfWeek, (max(FlightDelayMin) - min(FlightDelayMin)) as diff FROM flights
    GROUP BY DestCountry, dayOfWeek

#placeholder description

penghuo:
closed with commit.https://github.com/opendistro-for-elasticsearch/sql/commit/3730ac6e9b3a363c51432b34d2104567eb9babb7

Issue #270, opened 2019-11-05
Aggregation result should not contain unselected field


penghuo:
### Description
The unselected field `f1.DestCityName` should not be included in the result.
### SQL
```
GET /_opendistro/_sql?format=jdbc
{
  "query": """
      SELECT count(*) as tCount 
      FROM kibana_sample_data_flights f1      
      GROUP BY f1.DestCityName
      HAVING count(*) > 500
  """
}
```
### Actual Result
```
{
  "schema": [
    {
      "name": "DestCityName",
      "type": "keyword"
    },
    {
      "name": "tCount",
      "type": "long"
    }
  ],
  "total": 2,
  "datarows": [
    [
      "Zurich",
      "691.0"
    ],
    [
      "Xi'an",
      "526.0"
    ]
  ],
  "size": 2,
  "status": 200
}
```

### Expected Result
```
{
  "schema": [
    {
      "name": "tCount",
      "type": "long"
    }
  ],
  "total": 2,
  "datarows": [
    [
      "691.0"
    ],
    [
      "526.0"
    ]
  ],
  "size": 2,
  "status": 200
}
```

penghuo:
closed with commit.https://github.com/opendistro-for-elasticsearch/sql/commit/3730ac6e9b3a363c51432b34d2104567eb9babb7

Issue #229, opened 2019-10-18
Unable to process nested aggregate functions


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT TAN(MAX(AvgTicketPrice)) FROM kibana_sample_data_flights

**Results:**
Raw response received: {
  "error": {
    "reason": "Invalid SQL query",
    "details": "",
    "type": "NullPointerException"
  },
  "status": 400
}

**Expectation:**
No error, and outer function should be correctly applied to the aggregation result.
Example: (run with MySQL using subset of Flights data, so calculation result is not exact)
```
| TAN(MAX(AvgTicketPrice)) | 
| 0.3662764123396319 |
```

dai-chen:
Just checked `SELECTMAX(AvgTicketPrice) ...` can work individual for query without `GROUP BY`. But function composition breaks both execution and explain and ends up throwing NPE.

penghuo:
closed with commit.https://github.com/opendistro-for-elasticsearch/sql/commit/3730ac6e9b3a363c51432b34d2104567eb9babb7

PR #352, opened 2020-01-23
Add the expression support post processing

penghuo:
*Issue #, if available:*
#229 
#194 
#270
#292 

*Description of changes:*
### 1.Data Models
The following paragraph describe the BNF grammar for Expression Values.
```
value             -> collection_value
                     tuple_value
                     scalar_value
                     missing_value
missing_value     -> MISSING
tuple_value       -> {}
                   | {string_value: value (, string_value: value)*}
collection_value  -> [(value)+]
scalar_value      -> string_value
                   | float_value
                   | double_value
                   | integer_value
                   | long_value         
```
The type of the values are: collection_value, tuple_value, missing_value, scalar_value.

* collection_value is an array could contain duplicated values. It is similar to the JSON array.
* tuple_value is a set of <name, value> pairs.
* missing_value denote the value is missing in the context.
* scalar_value is the core data types defined in SQL 

### 2.BindingTuple
BindingTuple is a set of <bindingName, value>. e.g. Taken the following sample JSON object as example, There are two <bindingName, value> pairs in the BindingTuple. (1) <“integerValue”, integer_value(1)>, (2) <“stringValue”, “string”>. 
```
{
    "integerValue": 1,
    "stringValue":  "string"
}
```
### 3.Expression
The following paragraph describe the behavior of the Expression.
```
// constructors
reference exp    : ExprValue -> Exp
literal exp      : ExprValue -> Exp
unary exp        : Exp       -> Exp
binary exp       : Exp x Exp -> Exp

// observer
valueOf          : Exp x BindingTuple -> ExprValue
```
Let’s take an example to explain how to construct the expression and evaluate the expression in the BindingTuple context.
```
// BindingTuple
{
    "a": 1,
    "b": 2
}
// Construct Expression
addExpression = expression(ref("a"), ref("b"))

// valueOf Expression with BindingTuple.
// The evaluation is executed in application order.
addExpression.valueOf(bindingTuple)
-> (+ ref("a").valueOf(bindingTuple) ref("b").valueOf(bindingTuple))
-> (+ 1 2)
-> 3
```


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/expression/core/UnaryExpressionTest.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Test
+    public void log10ShouldPass() {
+        assertEquals(0.3010299956639812,
+                     apply(ScalarOperation.LOG10, literal(doubleValue(2d))));
+    }
+
+    @Test
+    public void absWithStringShouldThrowException() {
```
davidcui-amzn: Any specific reason only ABS() and Tan2() were tested against with String values?

penghuo: ideally all the type should been tested. the purpose of the test is to verify the error message is as expected.

Issue #316, opened 2019-12-11
Type information is lost in JDBC format


dai-chen:
In JDBC formatter, we need to return a type for each field in SELECT which is required by JDBC driver. However, this type is hardcoding for now. For example, `SUM` always return `double` even if the field given is an `integer`.

```
GET kibana_sample_data_flights/_mapping
{
  "kibana_sample_data_flights" : {
    "mappings" : {
      "properties" : {
        ...
        "FlightDelayMin" : {
          "type" : "integer"
        },
       ...
      }
    }
  }
}

POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT SUM(FlightDelayMin) FROM kibana_sample_data_flights"
}
{
  "schema": [{
    "name": "SUM(FlightDelayMin)",
    "type": "double"
  }],
  "total": 1,
  "datarows": [["618150.0"]],
  "size": 1,
  "status": 200
}
```

PR #365, opened 2020-02-07
Return Correct Type Information for Fields 

davidcui-amzn:
*Issue #, if available:*
#316 
*Description of changes:*
Type Information for fields in JDBC format used to be hard-coded in, and would return consistently regardless of the field type itself for scripted fields (e.g `ABS(int_type)` would always return type `double`)

Now, JDBC type information returns correctly based on the field. Script fields in the `ORDER BY` clause are returned hard-coded, as we currently only return either `STRING` or `NUMBER`. 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
@@ -25,7 +25,6 @@ import com.amazon.opendistroforelasticsearch.sql.executor.adapter.QueryPlanReque
 import com.amazon.opendistroforelasticsearch.sql.executor.format.DataRows.Row;
 import com.amazon.opendistroforelasticsearch.sql.executor.format.Schema.Column;
 import com.amazon.opendistroforelasticsearch.sql.expression.domain.BindingTuple;
-import com.amazon.opendistroforelasticsearch.sql.plugin.RestSqlAction;
 import com.amazon.opendistroforelasticsearch.sql.query.DefaultQueryAction;
 import com.amazon.opendistroforelasticsearch.sql.query.QueryAction;
 import com.amazon.opendistroforelasticsearch.sql.query.planner.core.ColumnNode;
@@ -58,7 +57,7 @@ public class Protocol {
             this.columnNodeList =
                     ((QueryPlanRequestBuilder) (((QueryPlanQueryAction) queryAction).explain())).outputColumns();
         } else if (queryAction instanceof DefaultQueryAction) {
-            scriptColumnType = RestSqlAction.performAnalysis(queryAction.getSqlRequest().getSql());
+            scriptColumnType = queryAction.getScriptColumnType();
         }
         this.formatType = formatType;
         QueryStatement query = queryAction.getQueryStatement();
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
@@ -15,17 +15,20 @@
 
 package com.amazon.opendistroforelasticsearch.sql.executor.format;
 
+import com.amazon.opendistroforelasticsearch.sql.domain.ColumnTypeProvider;
 import com.amazon.opendistroforelasticsearch.sql.domain.Delete;
 import com.amazon.opendistroforelasticsearch.sql.domain.IndexStatement;
 import com.amazon.opendistroforelasticsearch.sql.domain.Query;
 import com.amazon.opendistroforelasticsearch.sql.domain.QueryStatement;
-import com.amazon.opendistroforelasticsearch.sql.executor.format.DataRows.Row;
-import com.amazon.opendistroforelasticsearch.sql.executor.format.Schema.Column;
 import com.amazon.opendistroforelasticsearch.sql.executor.adapter.QueryPlanQueryAction;
 import com.amazon.opendistroforelasticsearch.sql.executor.adapter.QueryPlanRequestBuilder;
+import com.amazon.opendistroforelasticsearch.sql.executor.format.DataRows.Row;
+import com.amazon.opendistroforelasticsearch.sql.executor.format.Schema.Column;
 import com.amazon.opendistroforelasticsearch.sql.expression.domain.BindingTuple;
-import com.amazon.opendistroforelasticsearch.sql.query.planner.core.ColumnNode;
+import com.amazon.opendistroforelasticsearch.sql.plugin.RestSqlAction;
+import com.amazon.opendistroforelasticsearch.sql.query.DefaultQueryAction;
 import com.amazon.opendistroforelasticsearch.sql.query.QueryAction;
+import com.amazon.opendistroforelasticsearch.sql.query.planner.core.ColumnNode;
 import org.elasticsearch.client.Client;
 import org.json.JSONArray;
 import org.json.JSONObject;
@@ -48,11 +51,14 @@ public class Protocol {
     private ResultSet resultSet;
     private ErrorMessage error;
     private List<ColumnNode> columnNodeList;
+    private ColumnTypeProvider scriptColumnType = new ColumnTypeProvider();
 
     public Protocol(Client client, QueryAction queryAction, Object queryResult, String formatType) {
         if (queryAction instanceof QueryPlanQueryAction) {
             this.columnNodeList =
                     ((QueryPlanRequestBuilder) (((QueryPlanQueryAction) queryAction).explain())).outputColumns();
+        } else if (queryAction instanceof DefaultQueryAction) {
+            scriptColumnType = RestSqlAction.performAnalysis(queryAction.getSqlRequest().getSql());
         }
         this.formatType = formatType;
         QueryStatement query = queryAction.getQueryStatement();
@@ -75,7 +81,7 @@ public class Protocol {
         if (queryStatement instanceof Delete) {
             return new DeleteResultSet(client, (Delete) queryStatement, queryResult);
         } else if (queryStatement instanceof Query) {
-            return new SelectResultSet(client, (Query) queryStatement, queryResult);
+            return new SelectResultSet(client, (Query) queryStatement, queryResult, scriptColumnType);
         } else if (queryStatement instanceof IndexStatement) {
             IndexStatement statement = (IndexStatement) queryStatement;
             StatementType statementType = statement.getStatementType();
--- a/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
@@ -16,6 +16,7 @@
 package com.amazon.opendistroforelasticsearch.sql.executor.format;
 
 import com.alibaba.druid.sql.ast.expr.SQLCaseExpr;
+import com.amazon.opendistroforelasticsearch.sql.domain.ColumnTypeProvider;
 import com.amazon.opendistroforelasticsearch.sql.domain.Field;
 import com.amazon.opendistroforelasticsearch.sql.domain.JoinSelect;
 import com.amazon.opendistroforelasticsearch.sql.domain.MethodField;
@@ -65,17 +66,19 @@ public class SelectResultSet extends ResultSet {
     private String indexName;
     private String typeName;
     private List<Schema.Column> columns = new ArrayList<>();
+    private ColumnTypeProvider scriptColumnType;
 
     private List<String> head;
     private long size;
     private long totalHits;
     private List<DataRows.Row> rows;
 
-    public SelectResultSet(Client client, Query query, Object queryResult) {
+    public SelectResultSet(Client client, Query query, Object queryResult, ColumnTypeProvider scriptColumnType) {
         this.client = client;
         this.query = query;
         this.queryResult = queryResult;
         this.selectAll = false;
+        this.scriptColumnType = scriptColumnType;
 
         if (isJoinQuery()) {
             JoinSelect joinQuery = (JoinSelect) query;
@@ -325,7 +328,7 @@ public class SelectResultSet extends ResultSet {
                 if (field.getExpression() instanceof SQLCaseExpr) {
                     return Schema.Type.TEXT;
                 }
-                return SQLFunctions.getScriptFunctionReturnType(field);
+                 return SQLFunctions.getScriptFunctionReturnType(field, scriptColumnType);
             }
             default:
                 throw new UnsupportedOperationException(
@@ -374,6 +377,7 @@ public class SelectResultSet extends ResultSet {
              * name instead.
              */
             if (fieldMap.get(fieldName) instanceof MethodField) {
+
                 Field methodField = fieldMap.get(fieldName);
                 columns.add(
                         new Schema.Column(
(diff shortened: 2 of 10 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -1023,4 +1015,31 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
...
+    public static Schema.Type getOrderByFieldType(Field field) {
+        String functionName = ((ScriptMethodField) field).getFunctionName().toLowerCase();
+        if (functionName.equals("cast")) {
+            String castType = ((SQLCastExpr) field.getExpression()).getDataType().getName();
+            return getCastFunctionReturnType(castType);
+        }
+
+        if (numberOperators.contains(functionName) || mathConstants.contains(functionName)
+                || trigFunctions.contains(functionName) || binaryOperators.contains(functionName)) {
+            return Schema.Type.DOUBLE;
```
chloe-zh: A reminder: the number operators do not always return `DOUBLE`, for example, `abs(integer)` should return `INTEGER`, and `sign(...)` should always return an integer type.

davidcui-amzn: This is just for the `ORDER BY` clause- the `scriptSortType` will only return either `STRING` or `NUMBER`, so returning `DOUBLE` for every numeric field in the `ORDER BY` will not affect the field information of the fields in `SELECT`

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
```diff
@@ -48,11 +51,14 @@
     private ResultSet resultSet;
     private ErrorMessage error;
     private List<ColumnNode> columnNodeList;
+    private ColumnTypeProvider scriptColumnType = new ColumnTypeProvider();
 
     public Protocol(Client client, QueryAction queryAction, Object queryResult, String formatType) {
         if (queryAction instanceof QueryPlanQueryAction) {
             this.columnNodeList =
                     ((QueryPlanRequestBuilder) (((QueryPlanQueryAction) queryAction).explain())).outputColumns();
+        } else if (queryAction instanceof DefaultQueryAction) {
```
penghuo: Is it possible to avoid running performAnalysis two times by build another QueryPlanner around DefaultQueryAction?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -325,7 +328,7 @@ private String getFieldName(Field field) {
                 if (field.getExpression() instanceof SQLCaseExpr) {
                     return Schema.Type.TEXT;
                 }
-                return SQLFunctions.getScriptFunctionReturnType(field);
+                 return SQLFunctions.getScriptFunctionReturnType(fieldIndex, field, scriptColumnType);
```
dai-chen: np: remove first space

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -973,28 +974,19 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
      * approach will return type of result column as DOUBLE, although there is enough information to understand that
      * it might be safely treated as INTEGER.
      */
-    public static Schema.Type getScriptFunctionReturnType(Field field) {
+    public static Schema.Type getScriptFunctionReturnType(
+            int fieldIndex, Field field, ColumnTypeProvider scriptColumnType) {
+        Schema.Type returnType = null;
+        if (scriptColumnType != null && fieldIndex != -1) {
```
dai-chen: In which case `fieldIndex` is -1?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -308,7 +311,7 @@ private String getFieldName(Field field) {
         }
     }
 
-    private Schema.Type fetchMethodReturnType(Field field) {
+    private Schema.Type fetchMethodReturnType(int fieldIndex, Field field) {
```
dai-chen: The 2nd argument should be `MethodField`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -973,28 +974,19 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
      * approach will return type of result column as DOUBLE, although there is enough information to understand that
      * it might be safely treated as INTEGER.
      */
-    public static Schema.Type getScriptFunctionReturnType(Field field) {
+    public static Schema.Type getScriptFunctionReturnType(
+            int fieldIndex, Field field, ColumnTypeProvider scriptColumnType) {
```
dai-chen: 2nd argument should be `MethodField`. Then I think you should reorganize this method like this:

```
returnType
if (methodField is cast)
    returnType = ...
else
    returnType = ...

if (returnType == null)
    throw exception ...

return returnType
```

After this, could you add UT for this function because of its importance? Maybe 3 test cases for the `if`, `else` and `throw exception` in `SQLFunctionsTest`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -1023,4 +1015,31 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
                 );
         }
     }
+
+    public static Schema.Type getOrderByFieldType(Field field) {
```
dai-chen: Could you add JavaDoc here to clarify that why we stick with hardcoding return type here? Basically I think it's because there are only 2 order by type TEXT and NUMBER in ES so what is returned here is essentially the category of function rather than the actual return type.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TypeInformationIT.java
```diff
@@ -0,0 +1,200 @@
...
+        JSONObject response = executeJdbcRequest("SELECT ABS(age) FROM " + TestsConstants.TEST_INDEX_ACCOUNT +
+                " ORDER BY age LIMIT 5");
+
+        verifySchema(response, schema("ABS(age)", null, "long"));
+        verifyDataRows(response,
+                rows(20),
+                rows(20),
+                rows(20),
+                rows(20),
+                rows(20));
```
dai-chen: I think you're safe to remove all these `verifyDataRows()` and focus on what this IT want to test (the type in schema). The correctness of data rows should be covered by other IT.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TypeInformationIT.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /*
+    trigFunctions
+     */
+    @Test
+    public void testSinWithLongFieldReturnsLong() {
+        JSONObject response = executeJdbcRequest("SELECT sin(balance) FROM " +
+                TestsConstants.TEST_INDEX_ACCOUNT + " ORDER BY firstname LIMIT 2");
+
+        verifySchema(response, schema("sin(balance)", null, "long"));
```
dai-chen: I think I made this mistake. `sin` should always return double like `log` right? Could you help confirm and fix this as well as `TAN`, `ATAN`, `ATAN2`, `COS` and `COSH` in `ScalarFunction` type system?

For example, change `SIN(func(T(NUMBER)).to(T)),` to `SIN(func(T(NUMBER)).to(DOUBLE)),`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
davidcui-amzn: if `fieldName` is not found in the `fieldNameList`

dai-chen: > if `fieldName` is not found in the `fieldNameList`

Is this possible?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
davidcui-amzn: good idea- will do

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TypeInformationIT.java
davidcui-amzn: will do

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TypeInformationIT.java
davidcui-amzn: Sure- will look into it

dai-chen: Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
davidcui-amzn: Looks like it's not- think its safe to remove based on `for (String fieldName : fieldNameList)` in `SelectResultSet`

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -209,7 +210,7 @@ private boolean isSQLFeatureEnabled() {
         return allowExplicitIndex && isSqlEnabled;
     }
 
-    private static ColumnTypeProvider performAnalysis(String sql) {
+    public static ColumnTypeProvider performAnalysis(String sql) {
```
penghuo: why change to public?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -65,17 +66,19 @@
...
     private long totalHits;
     private List<DataRows.Row> rows;
 
-    public SelectResultSet(Client client, Query query, Object queryResult) {
+    public SelectResultSet(Client client, Query query, Object queryResult, ColumnTypeProvider scriptColumnType) {
         this.client = client;
         this.query = query;
         this.queryResult = queryResult;
         this.selectAll = false;
+        this.scriptColumnType = scriptColumnType;
```
penghuo: why named scriptColumnType? In my understanding, it not limit to script, right?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -374,12 +377,14 @@ private String getFieldName(Field field) {
              * name instead.
              */
             if (fieldMap.get(fieldName) instanceof MethodField) {
-                Field methodField = fieldMap.get(fieldName);
+
```
penghuo: remove the empty line

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -973,34 +974,27 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
...
+    public static Schema.Type getScriptFunctionReturnType(
+            int fieldIndex, MethodField field, ColumnTypeProvider scriptColumnType) {
+        Schema.Type returnType;
         String functionName = ((ScriptMethodField) field).getFunctionName().toLowerCase();
+        if (!numberOperators.contains(functionName) && !mathConstants.contains(functionName)
+                && !trigFunctions.contains(functionName) && !stringOperators.contains(functionName)
+                && !stringFunctions.contains(functionName) && !binaryOperators.contains(functionName)
+                && !dateFunctions.contains(functionName) && !conditionalFunctions.contains(functionName)
+                && !utilityFunctions.contains(functionName)) {
+            throw new UnsupportedOperationException(
```
penghuo: Is it possible to handle the unsupported case in SemanticAnalyzer?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -973,34 +974,27 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
      * approach will return type of result column as DOUBLE, although there is enough information to understand that
      * it might be safely treated as INTEGER.
      */
-    public static Schema.Type getScriptFunctionReturnType(Field field) {
+    public static Schema.Type getScriptFunctionReturnType(
+            int fieldIndex, MethodField field, ColumnTypeProvider scriptColumnType) {
```
penghuo: The input argument could be reduced as (MethodFiled field, Schema.Type resolvedType)

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
davidcui-amzn: outdated change from when it was called twice- will revert

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
davidcui-amzn: good point - will change to `outputColumnType`

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
davidcui-amzn: good catch - will change

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
davidcui-amzn: good catch - looks like SemanticAnalyzer checks for unsupported functions

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -974,25 +973,14 @@ public String getCastScriptStatement(String name, String castType, List<KVValue>
...
-                            functionName));
-        }
         if (functionName.equals("cast")) {
             String castType = ((SQLCastExpr) field.getExpression()).getDataType().getName();
             return getCastFunctionReturnType(castType);
         } else {
-            returnType = scriptColumnType.get(fieldIndex);
+            returnType = resolvedType;
         }
         return returnType;
```
penghuo: could it be simplified as `return resolvedType`

penghuo:
LGTM, Thanks for the change.

davidcui-amzn:
> LGTM, Thanks for the change.

Looks like you may have forgotten to approve the PR- just posted the last change.

Issue #353, opened 2020-01-23
Add capability to test against external Elasticsearch cluster.


dai-chen:
@rishabh6788 created branch `sql-integ-test-1.3` for testing against external ES cluster. And he found many SQL integration test failed due to node unavailable issue.

PR #374, opened 2020-03-05
Integration test with external ES cluster

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/353

*Description of changes:* To automate the sanity test for new release, we need the capability to run all our integration tests with external cluster setup by release candidate artifacts, such as docker, rpm, zip etc. To make this happen, this PR covered the following code changes:

 1. **Migrating to `ESRestTestCase`**: we used `ESIntegTestCase` which has problem with external cluster.
 2. **Removing index type**: we used deprecated index type which is going to be removed in ES 8 and will be rejected by external cluster.
 3. **Fixing broken tests**: some ES API in use is actually deprecated and previous internal cluster didn't complain.

*For reviewers:* Since this is quite a large PR (all test index, data and queries using type need to be changed), **please focus on `SQLIntegTestCase` only for your convenience**.

*Testing*:

 1. All tests: `./gradlew build`
 2. Test with internal cluster: `./gradlew integTestRunner`
 3. Test with external cluster: `./gradlew integTestRunner -Dtests.cluster= -Dtests.rest.cluster=127.0.0.1:9200` (not sure why tests.cluster is required though it is not in use by new ESIntegTestCase)
 4. Comparison test: `./gradlew integTestRunner -DtestType=comparison`
 5. Doc test: `./gradlew integTestRunner -DtestType=doctest`


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MetaDataQueriesIT.java
```diff
@@ -256,7 +256,7 @@ public void showWildcardIndex() throws IOException {
 
         String pattern = String.format("%s.*", TestsConstants.TEST_INDEX);
         JSONArray dataRows = getDataRows(response);
-        assertThat(dataRows.length(), equalTo(3));
+        assertThat(dataRows.length(), greaterThan(0));
```
abbashus: What is the reason to change this? Aren't we loosing the assertion.

dai-chen: Good catch! I didn't figure out how to do index clean up after each set of class with the new base class. The cleanup works now so I'm reverting this line of change. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/maker/AggMaker.java
```diff
@@ -555,6 +555,9 @@ private DateHistogramAggregationBuilder dateHistogram(MethodField field) throws
                     case "interval":
                         dateHistogram.dateHistogramInterval(new DateHistogramInterval(kv.value.toString()));
                         break;
+                    case "fixed_interval":
+                        dateHistogram.fixedInterval(new DateHistogramInterval(kv.value.toString()));
+                        break;
```
abbashus: What is this change about?

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
```diff
@@ -530,9 +536,14 @@ private static String quoteParams(String... params) {
         }
     }
 
+    /** Explicitly pass in name used to generate variable ID because methodName is not always valid */
+    private Tuple<String, String> dateFunctionTemplate(String name, String methodName, SQLExpr field) {
+        String id = nextId(name);
+        return new Tuple<>(id, def(id, doc(field) + ".value." + methodName));
+    }
+
```
abbashus: Sorry, not very clear about this change. Can you give an example how this affects things.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/maker/AggMaker.java
dai-chen: Changes regarding ES and painless API are due to the old ones we used are deprecated. The previous testing with internal cluster didn't complain about this but it became problem now with external cluster.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/SQLFunctions.java
dai-chen: Sorry for confusion. Basically this is caused by the painless API changes above, for example:

```
functionStr = dateFunctionTemplate("weekOfWeekyear",
                                   "get(WeekFields.ISO.weekOfWeekBasedYear())",
                                   (SQLExpr) paramers.get(0).value);
```

The old `dateFunctionTemplate(methodName, field)` passes `get(WeekFields.ISO.weekOfWeekBasedYear())` to `nextId()` which generates an invalid variable name in painless script. Will add a little more elaboration in JavaDoc.

abbashus: Thanks, clear now.

dai-chen: Cool. Thanks!

Issue #339, opened 2020-01-09
Comparison test framework improvement


dai-chen:
Creating this issue for followup work on test framework. Will add more to-do items if needed.

1.In our JDBC implementation, `getColumnName()` returns column original name in SELECT and `getColumnLabel` returns its alias. But other reference database seems return alias directly from `getColumnName()`.

2.Just found that org.json seems not respect column order in `LinkedHashMap`. Needs to convert it to list manually.

3.After fixing bugs in our JDBC driver, we need to add dependency for latest artifact in Gradle. One option is to use local path: `testCompile files('libs/opendistro-sql-jdbc-1.4.0.0-SNAPSHOT.jar')` in build.gradle. In this case, our GitHub Action should check out and build JDBC master branch whenever it runs comparison test.

4.The current test framework passes Java `LinkedHashMap` and `HashSet` for schema and data rows in `DBResult` class. `HashMap` and `ArrayList` needs to be used instead for the case of schema in `SELECT *` and data rows in `ORDER BY`.

dai-chen:
The first 2 issue fixes have been merged. Need to figure out how to handle the 3rd issue especially.

dai-chen:
5.Need minor changes to support MySQL database.

The follow issue may be caused by length of `VARCHAR` is required in MySQL.

```
  2> java.lang.IllegalStateException: Failed to create table [kibana_sample_data_flights]
        at __randomizedtesting.SeedInfo.seed([10BD31FCB60C0A4A:F613C15757C860A7]:0)
        at com.amazon.opendistroforelasticsearch.sql.correctness.runner.connection.JDBCConnection.create(JDBCConnection.java:76)
        at com.amazon.opendistroforelasticsearch.sql.correctness.runner.ComparisonTest.loadData(ComparisonTest.java:74)
        at com.amazon.opendistroforelasticsearch.sql.correctness.CorrectnessIT.performComparisonTest(CorrectnessIT.java:60)

        Caused by:
        java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ',Origin VARCHAR,FlightDelay boolean,DistanceMiles float,FlightTimeMin float,Orig' at line 1
            at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
            at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
[...]

PR #377, opened 2020-03-10
Bug fix, return object type for field which has implicit object datatype when describe the table

penghuo:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql-jdbc/issues/57

*Description of changes:*
Return the object as type when field use implicit object datatype.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MetaDataQueriesIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MetaDataQueriesIT.java
@@ -16,14 +16,19 @@
 package com.amazon.opendistroforelasticsearch.sql.esintgtest;
 
 import org.elasticsearch.client.Request;
+import org.hamcrest.Description;
+import org.hamcrest.TypeSafeMatcher;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.Test;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
+import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_GAME_OF_THRONES;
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/MetaDataQueriesIT.java
```diff
@@ -290,6 +290,34 @@ public void describeSingleIndex() throws IOException {
...
+         * Assumed indices of fields in dataRows based on "schema" output for DESCRIBE given above:
+         * "TABLE_NAME"  : 2
+         * "COLUMN_NAME" : 3
+         * "TYPE_NAME"   : 5
+         */
+        for (int i = 0; i < dataRows.length(); i++) {
+            JSONArray row = dataRows.getJSONArray(i);
+            assertThat(row.get(2), equalTo(TestsConstants.TEST_INDEX_GAME_OF_THRONES));
+            assertThat(row.get(3), not(equalTo(JSONObject.NULL)));
+            assertThat(row.get(5), not(equalTo(JSONObject.NULL)));
```
abbashus: We should be checking `TYPE_NAME` to be `object`, instead of checking its `not null`.

Also can you test how this would impact JDBC driver.

dai-chen: I'm thinking of adding IT with JDBC driver involved later because our JDBC driver is already added as test dependency in Gradle. However, the problem is we don't have nightly build so it's difficult to test with latest changes in JDBC driver. The comparison test has same issue: https://github.com/opendistro-for-elasticsearch/sql/issues/339#issue-547727767

penghuo: Done. Verify the each row.
For testing with JDBC driver, I agree with chen's idea to do integration test based on nightly build pipeline.

Issue #16, opened 2019-03-27
Add Simple Cursor Support 


dai-chen:
Add cursor to support pagination of query on large index, including SELECT, GROUP BY and JOIN.

PR #390, opened 2020-03-19
Simple Query Cursor support

abbashus:
*Issue* #16 

*Description of changes:* This is the 1st phase of Cursor support  work. This PR implements the pagination support for simple SQL queries of the form `SELECT ... FROM ... WHERE ... ORDER BY`

*Reviewer:* To make CR easier, please have a look at the following sections of the [design doc](https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/dev/Pagination.md).
Protocol, JDBC, Salient points, Settings.
Also the integration tests in `CursorIT.java`


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
```diff
@@ -0,0 +1,317 @@
...
+        String rawResult = executeFetchQuery(query, 100, "raw");
+        rows = rawResult.split("\n");
+        // all the 1000 records (NO headers) are retrieved instead of fetch_size number of records
+        assertThat(rows.length, equalTo(1000));
+    }
+
+
+    public void verifyWithAndWithoutPaginationResponse(String sqlQuery, String cursorQuery, int fetch_size) throws IOException {
+        // we are only checking here for schema and aatarows
+        JSONObject withoutCursorResponse = new JSONObject(executeFetchQuery(sqlQuery, 0, JDBC));
```
penghuo: Is fetch_size the mandatory field for query?, if not, we don't need include it.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
```diff
@@ -0,0 +1,317 @@
...
+
+        JSONObject withCursorResponse = new JSONObject("{\"schema\":[],\"datarows\":[]}");
+        JSONArray schema = withCursorResponse.getJSONArray("schema");
+        JSONArray dataRows = withCursorResponse.getJSONArray("datarows");
+
+        JSONObject tempResponse = new JSONObject(executeFetchQuery(cursorQuery, fetch_size, JDBC));
```
penghuo: suggestion: tempResponse -> firstResponse

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
abbashus: No, its not mandatory. There is another version `executeFetchLessQuery` without passing the `fetch_size`.  We passed `0` as `fetch_size` here for default non-paginated behavior.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
```diff
@@ -57,6 +68,16 @@ public String getSql() {
         return this.sql;
     }
 
+    public String cursor() {
+        return this.cursor;
+//        return (jsonContent == null) ? null : jsonContent.getString(SqlRequestFactory.SQL_CURSOR_FIELD_NAME);
```
penghuo: delete it

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
```diff
@@ -59,16 +61,36 @@ private static SqlRequest parseSqlRequestFromPayload(RestRequest restRequest) {
...
             List<PreparedStatementRequest.PreparedStatementParameter> parameters = parseParameters(paramArray);
             return new PreparedStatementRequest(sql, jsonContent, parameters);
         }
-        return new SqlRequest(sql, jsonContent);
+        return new SqlRequest(sql, validateAndGetFetchSize(jsonContent), jsonContent);
+    }
+
+
+    private static Integer validateAndGetFetchSize(JSONObject jsonContent) {
+        Integer fetchSize = null;
```
penghuo: null or 0?

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
```diff
@@ -59,16 +61,36 @@ private static SqlRequest parseSqlRequestFromPayload(RestRequest restRequest) {
...
+        try {
+            if (jsonContent.has(SQL_FETCH_FIELD_NAME)) {
+                fetchSize = jsonContent.getInt(SQL_FETCH_FIELD_NAME);
+                if (fetchSize < 0) {
+                    throw new IllegalArgumentException("Fetch_size must be greater or equal to 0");
+                }
+            }
+        } catch (JSONException e) {
+            throw new IllegalArgumentException("Failed to parse field [" + SQL_FETCH_FIELD_NAME +"]", e);
```
penghuo: In which case the exception will been throw? Does it been tested?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlSettings.java
```diff
@@ -68,6 +73,13 @@ public SqlSettings() {
         settings.put(METRICS_ROLLING_INTERVAL, Setting.longSetting(METRICS_ROLLING_INTERVAL, 60L, 1L,
                 NodeScope, Dynamic));
 
+        // Settings for cursor
+        settings.put(CURSOR_ENABLED, Setting.boolSetting(CURSOR_ENABLED, true, NodeScope, Dynamic));
```
penghuo: Have you updated the doc in PluginSettingIT?

abbashus: Will do.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorActionRequestRestExecutorFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package com.amazon.opendistroforelasticsearch.sql.executor.cursor;
+
+import com.amazon.opendistroforelasticsearch.sql.executor.Format;
+import org.elasticsearch.rest.RestRequest;
+
+public class CursorActionRequestRestExecutorFactory {
+    //TODO: add javadocs, see RestExecutor
+
+    public static CursorRestExecutor createExecutor(RestRequest request, String cursor, Format format) {
+        if (isCursorCloseRequest(request)) {
```
penghuo: what happen if it is an explain request

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorRestExecutor.java
```diff
@@ -0,0 +1,30 @@
...
+
+
+package com.amazon.opendistroforelasticsearch.sql.executor.cursor;
+
+import org.elasticsearch.client.Client;
+import org.elasticsearch.rest.RestChannel;
+
+import java.util.Map;
+
+public interface CursorRestExecutor {
```
penghuo: Doc

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+        this.isBlocking = isBlocking;
+    }
+
+
+    public void execute(Client client, Map<String, String> params, RestChannel channel) throws Exception {
+        LOG.info("executing something inside CursorAsyncRestExecutor execute ");
+        async(client, params, channel);
+    }
+
+    public String execute(Client client, Map<String, String> params) throws Exception {
```
penghuo: What is the function of this?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+import org.elasticsearch.threadpool.ThreadPool;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Map;
+import java.util.function.Predicate;
+
+import static com.amazon.opendistroforelasticsearch.sql.plugin.SqlSettings.QUERY_SLOWLOG;
+
+public class CursorAsyncRestExecutor implements CursorRestExecutor {
```
penghuo: It almost as same as AsyncRestExecutor. Any improvement?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorActionRequestRestExecutorFactory.java
abbashus: As per plugin usage ,  `_explain` does make sense with `cursor`. Will handle this case.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
```diff
@@ -0,0 +1,110 @@
...
+        this.cursorId = cursorId;
+    }
+
+    public void execute(Client client, Map<String, String> params, RestChannel channel) throws Exception {
+        try {
+            String formattedResponse = execute(client, params);
+            channel.sendResponse(new BytesRestResponse(OK, "application/json; charset=UTF-8", formattedResponse));
+        } catch (IllegalArgumentException | JSONException e) {
+            Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_CUS).increment();
+            e.printStackTrace();
```
penghuo: Log.error?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
```diff
@@ -0,0 +1,110 @@
...
+            } else if (status > 499) {
+                Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+            }
+            e.printStackTrace();
+            channel.sendResponse(new BytesRestResponse(channel, e));
+        }
+    }
+
+    public String execute(Client client, Map<String, String> params) throws Exception {
+        String decodedCursorContext = new String(Base64.getDecoder().decode(cursorId));
```
penghuo: Add separate class to encapsulate is much better.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
```diff
@@ -0,0 +1,110 @@
...
+        JSONObject cursorJson = new JSONObject(decodedCursorContext);
```
penghuo: Instead of using JSONObject, I suggest to abstract cursor as JavaBean.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    private String handleAggregationCursorCloseRequest(Client client, JSONObject cursorContext) {
+        return SUCCEEDED_TRUE;
+    }
+
+    private String handleJoinCursorCloseRequest(Client client, JSONObject cursorContext) {
+        return SUCCEEDED_FALSE;
+    }
```
penghuo: Why the response is different?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
abbashus: The signature of `execute` methods ini `CursorRestExecutor` does not take `QueryAction` as parameter. Will try to combine them in one.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Will add.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorRestExecutor.java
abbashus: Will add.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+    public CursorResultExecutor(String cursorId, Format format) {
+        this.cursorId = cursorId;
+        this.format = format;
+    }
+
+    public void execute(Client client, Map<String, String> params, RestChannel channel) throws Exception {
+        try {
+            String formattedResponse = execute(client, params);
+            channel.sendResponse(new BytesRestResponse(OK, "application/json; charset=UTF-8", formattedResponse));
+        } catch (IllegalArgumentException | JSONException e) {
```
penghuo: Should we also catch SQLFeatureNotSupportedException in here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Bad copy, will make it consitent.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
```diff
@@ -153,6 +198,36 @@ private String outputInTableFormat() {
         return null;
     }
 
+
+    public String cursorFormat() {
+        if (status == OK_STATUS && cursorContext!=null) {
+            switch (formatType) {
+                case "jdbc":
+                    return cursorOutputInJDBCFormat();
+                case "table":
```
penghuo: could be removed.

abbashus: Which part- `case table`?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/ResultSet.java
```diff
@@ -48,4 +50,8 @@ protected boolean matchesPattern(String string, String pattern) {
         Matcher matcher = p.matcher(string);
         return matcher.find();
     }
+
+    public String getCursor() {
```
penghuo: return null or cursor?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
```diff
@@ -203,4 +278,79 @@ private JSONArray dataEntry(Row dataRow, Schema schema) {
...
+
+                } else {
+                    // explicitly setting this to null to avoid any ambiguity
+                    cursor = null;
+                }
+                options.remove("scrollId");
+                options.remove("fetch_size");
+                break;
+            case AGGREGATION:
+                throw new ElasticsearchException("Cursor not yet supported for GROUP BY queries");
```
penghuo: FeatureNotSupportedExecption?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
```diff
@@ -203,4 +278,79 @@ private JSONArray dataEntry(Row dataRow, Schema schema) {
...
+                    cursorJson.put("left", pagesLeft());
+                    setIndexNameInCursor(cursorJson);
+                    setFieldAliasMapInCursor(cursorJson);
+                    cursor = encodeCursorContext(cursorJson);
+
+                } else {
+                    // explicitly setting this to null to avoid any ambiguity
+                    cursor = null;
+                }
+                options.remove("scrollId");
```
penghuo: why remove?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
```diff
@@ -203,4 +278,79 @@ private JSONArray dataEntry(Row dataRow, Schema schema) {
         }
         return entry;
     }
-}
+
+    public void generateCursorId() {
```
penghuo: suggest to separate cursor related code from Protocol and add UT for that.

penghuo:
Thanks for this important change!

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
```diff
@@ -59,16 +61,36 @@ private static SqlRequest parseSqlRequestFromPayload(RestRequest restRequest) {
...
+        return new SqlRequest(sql, validateAndGetFetchSize(jsonContent), jsonContent);
+    }
+
+
+    private static Integer validateAndGetFetchSize(JSONObject jsonContent) {
+        Integer fetchSize = null;
+        try {
+            if (jsonContent.has(SQL_FETCH_FIELD_NAME)) {
+                fetchSize = jsonContent.getInt(SQL_FETCH_FIELD_NAME);
+                if (fetchSize < 0) {
```
davidcui-amzn: if fetchSize is initialized to null, will this throw a NullPointerException?

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
davidcui-amzn: +1

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
```diff
@@ -0,0 +1,317 @@
...
+
+    @Override
+    protected void init() throws Exception {
+        loadIndex(Index.ACCOUNT);
+    }
+
+    @Test
+    public void invalidFetchSize() throws IOException {
+        // invalid fetch_size --> negative(-2), non-numeric("hello")
+        // acceptable fetch_size --> positive numbers, even in string form "532.4"
```
davidcui-amzn: should these style of comments go outside of the actual test case as a JavaDoc comment?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/QueryIT.java
```diff
@@ -1297,6 +1297,7 @@ public void isNotNullTest() throws IOException {
     }
 
     @Test
+    @Ignore("Use of scroll hint is deprecated in lieu of cursor support")
```
davidcui-amzn: if the scrollhint is deprecated should we delete these tests?

abbashus: Even I had the same concern, will delete these tests.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
abbashus: `null` to imply that `fetch_size` was not passed in query. In `DefaultQueryAction`, if `fetch_size == null`, we will use the `fetch_size` value from cluster settings.

`fetch_size == 0` will imply falling back to non-paginated behavior.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
abbashus: This exception will be thrown when passing non-numeric value to `fetch_size` , for example `"fetch_size" : "hello world"`. An `IllegalArgumentException` is thrown to be caught as client error (400) in `RestSqlAction.java`. Will add a test for this.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
abbashus: Not necessary for tests, does not affect readability or clarity. Its a best practice for code exposed as public libraries.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
abbashus: This will never throw `NullPointerException`, as we are either assigning integer value to `fetch_size` or throwing `JSONException`

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
abbashus: Keeping it simply `response` as the variable is being re-used to hold subsequent page results.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
```diff
@@ -36,11 +36,22 @@
 
     String sql;
     JSONObject jsonContent;
+    String cursor;
```
dai-chen, line 39: Since we're adding new fields to the protocol and version information is missing, could you add a `version` field too? For example, `version` missing or 1.0 means "old" client code and `version` = 2.0 means "new".

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/CursorIT.java
```diff
@@ -0,0 +1,373 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestUtils.getResponseBody;
+import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_ACCOUNT;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.containsString;
+
+public class CursorIT extends SQLIntegTestCase {
```
dai-chen, line 33: I assume cursor can work with PreparedStatement. If so, could you check if our cursor implementation can work with prepared statement request and add test case here?

abbashus: Sure, will confirm and add a test case as well.

dai-chen: Thanks!

abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
dai-chen: I took a quick look into version field of other protocol. Take Internet Protocol (IP) for example, the first field in the header is version which indicates IPv4 or IPv6. What follows this field in the header differs for different version so application is aware of how to parse the entire header.

In our case, because we are only adding new fields, probably we can delay adding version field. We need to reconsider this and may need to refactor our protocol carefully (ex. divide into header and body) when adding such as binary protocol.

Reference: https://en.wikipedia.org/wiki/IPv4#Version and https://en.wikipedia.org/wiki/IPv6_packet#Fixed_header

abbashus: Yes, we can add this later if needed. This will not break existing clients now as we adding new field and not modifying/deleting existing one.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+    public String execute(Client client, Map<String, String> params) throws Exception {
+
+        String decodedCursorContext = new String(Base64.getDecoder().decode(cursorId));
+        JSONObject cursorJson = new JSONObject(decodedCursorContext);
+
+        String type = cursorJson.optString("type", null); // see if it is a good case to use Optionals
+        CursorType cursorType = null;
+
+        if (type != null) {
+            cursorType = CursorType.valueOf(type);
```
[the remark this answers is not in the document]
abbashus: This logic has been removed after refactoring.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
```diff
@@ -0,0 +1,110 @@
...
+                    return handleDefaultCursorCloseRequest(client, cursorJson);
+                case AGGREGATION:
+                    return handleAggregationCursorCloseRequest(client, cursorJson);
+                case JOIN:
+                    return handleJoinCursorCloseRequest(client, cursorJson);
+                default: throw new VerificationException("Unsupported cursor");
+            }
+        }
+
+        throw new VerificationException("Invalid cursor");
```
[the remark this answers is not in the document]
abbashus: Added logging for invalid cursor type. Not adding type as part of exception message to hide implementation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
abbashus: Refactored cursor related code from `Protocol`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
abbashus: Removed this logic as part of refactoring.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
abbashus: Removed this logic as part of refactoring.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/ResultSet.java
abbashus: Removed this method as part of refactoring.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Abstracted all cursor encode/decode logic via `Cursor` interface and individual implementation.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorActionRequestRestExecutorFactory.java
abbashus: Handled this case. Added integration test for same.

docs/user/admin/settings.rst
```diff
@@ -384,3 +384,129 @@ Result set::
...
+opendistro.sql.cursor.enabled
+=============================
+
+Description
+-----------
+
+User can enable/disable pagination for all queries that are supported.
+
+1. The default value is false.
+2. This setting is node scope.
```
penghuo, line 396: what is the meaning of "node scope"? does it means the setting only apply to specific node?

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
penghuo: I suggest to use Optional instead of null value in this case.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
```diff
@@ -78,51 +84,61 @@ public void initialize(SearchRequestBuilder request) {
...
-        }
         updateRequestWithIndexAndRoutingOptions(select, request);
         updateRequestWithHighlight(select, request);
         updateRequestWithCollapse(select, request);
         updateRequestWithPostFilter(select, request);
         updateRequestWithInnerHits(select, request);
+    }
 
-        return new SqlElasticSearchRequestBuilder(request);
+    private void checkAndSetScroll() {
```
penghuo, line 106: Do we have UT to test the logic in here. IMO, the  setting, fetchSize, rowCount has different combination which determine the scroll logic which should be covered in the UT.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
penghuo: Yes, do we support table?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
```diff
@@ -0,0 +1,103 @@
...
+                Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_CUS).increment();
+            } else if (status > 499) {
+                Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+            }
+            LOG.error("Error completing cursor request", e);
+            channel.sendResponse(new BytesRestResponse(channel, e));
+        }
+    }
+
+    public String execute(Client client, Map<String, String> params) throws Exception {
```
penghuo, line 69: Two concerns.
1. The execute method is almost as same as CursorResultExecutor.execute, only difference is the LOG.error. Is it possible to avoid the duplication.
2. In case we already define the Curser, could we construct CursorCloseExecutor by using cursor and encapsulate the Cursor construct operation in CursorFactory?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+            /** Clear the scroll context on last page */
+            ClearScrollResponse clearScrollResponse = client.prepareClearScroll().addScrollId(newScrollId).get();
+            if (!clearScrollResponse.isSucceeded()) {
+                Metrics.getInstance().getNumericalMetric(MetricName.FAILED_REQ_COUNT_SYS).increment();
+                LOG.info("Error closing the cursor context {} ", newScrollId);
+            }
+        }
+
+        cursor.setRowsLeft(rowsLeft);
+        cursor.setScrollId(newScrollId);
```
penghuo, line 142: In case of rowsLeft = 0, we still include the newScrollId in response?

src/main/java/com/amazon/opendistroforelasticsearch/sql/cursor/DefaultCursor.java
```diff
@@ -0,0 +1,217 @@
...
+         cursor.setIndexPattern(json.getString(INDEX_PATTERN));
+         cursor.setScrollId(json.getString(SCROLL_ID));
+         cursor.setColumns(getColumnsFromSchema(json.getJSONArray(SCHEMA_COLUMNS)));
+         cursor.setFieldAliasMap(fieldAliasMap(json.getJSONObject(FIELD_ALIAS_MAP)));
+
+         return cursor;
+    }
+
+
+    public String getIndexPattern() {
```
penghuo, line 113: minor: consider to use Lombok to auto generate get/set method.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
abbashus: No, as part of cursor generation if `rowsleft <=0` the `cursor` is set to null, and thus omitted from response. See `DefaultCursor.generateId()`.  The code was structured like this to avoid `if...else` condition and the final logic to generate cursor id delegated to implementing class.

docs/user/admin/settings.rst
abbashus: > what is the meaning of "node scope"? 

This is a cluster wide setting, change on one node will be reflected on all other nodes.
Not sure, why it is named like this. This is not different from all other settings we already have.

> does it means the setting only apply to specific node?

No.

src/main/java/com/amazon/opendistroforelasticsearch/sql/cursor/DefaultCursor.java
abbashus: Sure, will consider using Lombok.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Sure, will give a shot to refactor as described.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
abbashus: Will add unit test for this.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlSettings.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
abbashus: Removed this method.

docs/user/admin/settings.rst
dai-chen: I may have a little context here. I recall the name was from ES `Setting` in which there is a NodeScope and IndexScope.

docs/user/admin/settings.rst
```diff
@@ -384,3 +384,129 @@ Result set::
...
+	  }
+	}
+
+opendistro.sql.cursor.keep_alive
+================================
+
+Description
+-----------
+
+User can set this value to indicate how long the cursor context should be kept open.Cursor contexts are resource heavy, and a lower value should be used if possible.
```
dai-chen, line 477: np: space between `open.Cursor`

docs/user/interfaces/endpoint.rst
```diff
@@ -91,3 +91,62 @@ Explain::
 	  }
 	}
 
+Cursor
+======
+
+Description
+-----------
+
+To get paginated response for a query, user needs to provide `fetch_size` parameter as part of normal query. The value of `fetch_size` should be greater than `0`. In absence of `fetch_size`, default value of 1000 is used. A value of `0` will fallback to non-paginated response. This feature is only available over `jdbc` format for now.
```
dai-chen, line 100: I think this section can be moved to new example 3 in Request Format in protocol doc: https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/interfaces/protocol.rst#request-format.

As for endpoint, I think the new cursor close API can be documented here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/cursor/CursorType.java
```diff
@@ -0,0 +1,48 @@
...
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.cursor;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum CursorType {
```
dai-chen, line 21: Could you add some JavaDoc here?

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
```diff
@@ -0,0 +1,119 @@
...
+import org.elasticsearch.rest.RestStatus;
+import org.elasticsearch.threadpool.ThreadPool;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Map;
+
+import static com.amazon.opendistroforelasticsearch.sql.plugin.SqlSettings.QUERY_SLOWLOG;
+
+public class CursorAsyncRestExecutor {
```
dai-chen, line 38: Any chance to reuse `AsyncRestExecutor`? I don't see any special logic here and you can make `AsyncRestExecutor` async all cursor request.

```
    AsyncRestExecutor(RestExecutor executor) {
        this(executor, ALL_ACTION_IS_BLOCKING);
    }
```

docs/user/interfaces/endpoint.rst
abbashus: I did trying adding it as example 3 in ProtocolIT ,  but I was not getting correct cursor response, also the explain output was showing non-scroll explain with `from` field. So omitted as it would have required extra plumbing.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
abbashus: Sure, will give a shot.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
abbashus: Removed.

docs/user/admin/settings.rst
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/query/DefaultQueryAction.java
abbashus: Added.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequestFactory.java
abbashus: Done.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
abbashus: Not needed after refactoring.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorAsyncRestExecutor.java
abbashus: Not straightforward to use AsyncRestExecutor here. AsyncRestExecutor requires using QueryAction, which we don't have for Cursor request.  Keeping it as it is for now.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorCloseExecutor.java
abbashus: Keeping it as it now, little tricky to refactor.

penghuo: In my opinion, the cursor format ":" which should be encapsulated to Cursor implementation and should not been exposed to caller.
We can leave it for now for time consideration, Please consider add UT to cover the logic in CursorCloseExecutor and CursorResultExecutor. If possible, add the issue to tracking the refactoring.

abbashus: https://github.com/opendistro-for-elasticsearch/sql/issues/426.

Adding UTs.

abbashus:
> No major concern for the functionality.
> I think we already have pretty good IT. But It seems we miss UT for new Class CursorXXXExecutor and changed class Protocol and SelectResultSet.
> For the time consideration, we could postpone refactoring the Cursor code structure to the next stage, but adding enough UT is the precondition to make the refactor easily.
> What do you think?

Agreed. Adding UTs for CursorXXXExecutor where possible.

src/main/java/com/amazon/opendistroforelasticsearch/sql/cursor/DefaultCursor.java
```diff
@@ -0,0 +1,170 @@
...
+    private Map<String, String> fieldAliasMap;
+
+    /** To get next batch of result */
+    private String scrollId;
+
+    /** To reduce the number of rows left by fetchSize */
+    @NonNull
+    private Integer fetchSize;
+
+    private Integer limit;
```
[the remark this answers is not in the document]
abbashus: Used to pass `limit` from `PrettyFormatExecutor`-> `Protocol` -> `SelectResultSet`

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+    public String execute(Client client, Map<String, String> params) throws Exception {
+        /**
+         * All cursor's are of the form <cursorType>:<base64 encoded cursor>
+         * The serialized form before encoding is upto Cursor implementation
+         */
+        String[] splittedCursor = cursorId.split(":");
+
+        if (splittedCursor.length!=2) {
+            throw new VerificationException("Not able to parse invalid cursor");
+        }
```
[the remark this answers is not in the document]
abbashus: > Is it possible that more colons present in Base64 encoded string after first colon?

No possible as whole string after first `:` is Base64 encoded.


> If so could we only split the whole string by first colon found?

Sure, will change to  `cursorId.split(":", 2)`

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/SelectResultSet.java
```diff
@@ -525,20 +579,66 @@ private void extractData() {
...
+
+    private void populateDefaultCursor(DefaultCursor cursor) {
+        /**
+         * Assumption: scrollId, fetchSize, limit already being set in
+         * @see PrettyFormatRestExecutor.buildProtocolForDefaultQuery()
+         */
+
+        Integer limit = cursor.getLimit();
+        long rowsLeft = rowsLeft(cursor.getFetchSize(), cursor.getLimit());
+        if (rowsLeft <= 0) {
```
dai-chen, line 613: I saw this cursor populate logic and scroll close logic in both here and `CursorResultExecutor`. Is it due to different code path?

abbashus: Yes, this logic is for first page request.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+             * This condition implies we are on the last page, and we might need to truncate the result from SearchHit[]
+             * Avoid truncating in following two scenarios
+             * 1. number of rows to be sent equals fetchSize
+             * 2. size of SearchHit[] is already less that rows that needs to be sent
+             *
+             * Else truncate to desired number of rows
+             */
+            SearchHit[] newSearchHits = Arrays.copyOf(searchHitArray, rowsLeft);
+            searchHits = new SearchHits(newSearchHits, searchHits.getTotalHits(), searchHits.getMaxScore());
+        }
```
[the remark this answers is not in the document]
abbashus: Good suggestion, will do, wasn't apparent initially.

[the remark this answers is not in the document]
abbashus: I realized that though we can skip unwanted `DataRows.Row` using `limit`, but this will apply to final flattened row, and will impact correctness result for nested queries.

For example:
If the final page has `rowLeft` as 2, implying two un-nested doc, and assuming each doc has atleast 3 nested docs. If use above logic there will be 2 `Row`, instead of 8 that we expect.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/format/Protocol.java
```diff
@@ -153,6 +182,30 @@ private String outputInTableFormat() {
         return null;
     }
 
+    public String cursorFormat() {
```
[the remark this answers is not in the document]
abbashus: True, ideally we should have single interface method in Protocol, but not doing it here for two reasons:

1. We don't support non-JDBC format for cursor.
2. Also protocol is different for cursor result  either `datarows` or `datarows + cursor`.

We can do a check to know if first page or subsequent pages, but the code does not turn out to be clean.

src/main/java/com/amazon/opendistroforelasticsearch/sql/executor/cursor/CursorResultExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+        String[] splittedCursor = cursorId.split(":");
+
+        if (splittedCursor.length!=2) {
+            throw new VerificationException("Not able to parse invalid cursor");
+        }
+
+        String type = splittedCursor[0];
+        CursorType cursorType = CursorType.getById(type);
+
+        if (cursorType!=CursorType.NULL) {
```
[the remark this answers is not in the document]
abbashus: Done.

PR #391, opened 2020-03-21
fix Jacoco integration issue by updating internal testing infrastructure

zhongnansu:
*Issue #, if available:*
#376 
*Description of changes:*
- fix Jacoco integration issue by updating internal testing infrastructure
- fix debug jvm property in gradle.build

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #393, opened 2020-03-24
Fix failure in IT test with external cluster

zhongnansu:
*Issue #, if available:*
#391 causes github workflow failure
*Description of changes:*
1. Fix failure in IT test with external cluster, basically the previous PR #391 which caused a github workflow failure
2. Add try with resource to avoid JMX connector won't close when an exception is thrown. 

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistroforelasticsearch.sql.esintgtest;
 
+import com.google.common.base.Strings;
 import org.elasticsearch.client.Request;
 import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.Response;
@@ -110,7 +111,7 @@ public abstract class SQLIntegTestCase extends ESRestTestCase {
         // jacoco.dir is set in sqlplugin-coverage.gradle, if it doesn't exist we don't
         // want to collect coverage so we can return early
         String jacocoBuildPath = System.getProperty("jacoco.dir");
-        if (jacocoBuildPath == null || jacocoBuildPath.isEmpty()) {
+        if (Strings.isNullOrEmpty(jacocoBuildPath)) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
```diff
@@ -107,23 +106,25 @@ protected boolean preserveClusterUponCompletion() {
...
-        connector.close();
+        try(JMXConnector connector = JMXConnectorFactory.connect(new JMXServiceURL(serverUrl))) {
+            IProxy proxy = MBeanServerInvocationHandler.newProxyInstance(
+                    connector.getMBeanServerConnection(), new ObjectName("org.jacoco:type=Runtime"), IProxy.class,
+                    false);
+
+            Path path = Paths.get(jacocoBuildPath + "/integTest.exec");
+            Files.write(path, proxy.getExecutionData(false));
+        } catch (Exception ex) {
+            throw new RuntimeException("Failed to dump coverage: " + ex);
```
dai-chen: Would be better if we keep the full error stack by `throw new RuntimeException("Failed to dump coverage", ex);`

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
```diff
@@ -107,23 +106,25 @@ protected boolean preserveClusterUponCompletion() {
     }
 
     @AfterClass
-    public static void dumpCoverage() throws IOException, MalformedObjectNameException {
+    public static void dumpCoverage() {
         // jacoco.dir is set in sqlplugin-coverage.gradle, if it doesn't exist we don't
         // want to collect coverage so we can return early
         String jacocoBuildPath = System.getProperty("jacoco.dir");
-        if (jacocoBuildPath.isEmpty()) {
+        if (jacocoBuildPath == null || jacocoBuildPath.isEmpty()) {
```
dai-chen: np: `isNullOrEmpty` in Strings or StringUtils is cleaner.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
zhongnansu: Agreed.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/SQLIntegTestCase.java
zhongnansu: good catch, use Strings.isNullOrEmpty(Guava) instead.

PR #400, opened 2020-03-30
New SQL cluster settings endpoint

abbashus:
**Description of changes:**
This change exposes a new endpoint to manage OpenDistro SQL only cluster settings.
All other settings are ignored. In terms of behavior, it is identical to `_cluster/settings` API. 

**Why is this change required?**
Though we can manage SQL settings via `POST _cluster/settings` API, some distributions blacklist access to ` _cluster/_settings` url. This change provides mechanism to modify SQL plugin only settings, thus providing better control on plugin behavior without affecting other settings.

**Testing**
New integration tests added. All existing unit and integration tests passing.


src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
```diff
@@ -0,0 +1,122 @@
...
+            );
+            clusterUpdateSettingsRequest.persistentSettings(
+                    filterSettings(clusterUpdateSettingsRequest.persistentSettings())
+            );
+
+            return channel -> client.admin().cluster().updateSettings(
+                    clusterUpdateSettingsRequest, new RestToXContentListener<>(channel));
+        } catch (Exception e) {
+            LOG.error("Error changing OpenDistro SQL plugin cluster settings", e);
+            e.printStackTrace();
```
penghuo: log.error already done this.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            if (source.containsKey(TRANSIENT)) {
+                clusterUpdateSettingsRequest.transientSettings((Map) source.get(TRANSIENT));
+            }
+            if (source.containsKey(PERSISTENT)) {
+                clusterUpdateSettingsRequest.persistentSettings((Map) source.get(PERSISTENT));
+            }
+
+            // filter out all non-sql settings
+            clusterUpdateSettingsRequest.transientSettings(
```
penghuo: why call transientSettings two times? ref: line 86.

penghuo:
please change the https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst accordingly.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
abbashus: Right. Missed that. Will remove.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlPlug.java
```diff
@@ -78,7 +78,8 @@ public String description() {
         Metrics.getInstance().registerDefaultMetrics();
```
davidcui-amzn: np: should this file be named SqlPlugin vs SqlPlug?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
abbashus: This is to take advantage of [`ClusterUpdateSettingsRequest.transientSettings(Map<String, ?> source)`](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterUpdateSettingsRequest.java#L117) method, instead of redoing again. The second `transientSettings()` call just assigns new Settings after filtering non `opendistro.sql` settings.


We can bring the same logic from [`ClusterUpdateSettingsRequest.transientSettings(Map<String, ?> source)`](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterUpdateSettingsRequest.java#L117), if there is a performance impact.

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/PluginIT.java
```diff
@@ -52,22 +55,399 @@ public void sqlEnableSettingsTest() throws IOException {
...
-        Assert.assertThat(error.getString("details"), equalTo("Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false"));
-        Assert.assertThat(error.getString("type"), equalTo("SQLFeatureDisabledException"));
-        resetClusterSettings(PERSISTENT, "opendistro.sql.enabled");
+        assertThat(error.getString("reason"), equalTo("Invalid SQL query"));
+        assertThat(error.getString("details"), equalTo("Either opendistro.sql.enabled or rest.action.multi.allow_explicit_index setting is false"));
+        assertThat(error.getString("type"), equalTo("SQLFeatureDisabledException"));
+        wipeAllClusterSettings();
+    }
+
+    @Test
```
davidcui-amzn: not a big deal, just curious- is there a way we can export some of these JSON strings to a separate file?

abbashus: I initially thought of putting it in some file under `tests/resources` directory, but reason of keeping here in the test suite is to avoid the breaking the workflow for the reviewer to go back and forth between tests and fixtures file or any developer looking at it without IDE.

abbashus:
> please change the https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst accordingly.

Should we change [Endpoint](https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/interfaces/endpoint.rst) doc instead , since we are introducing a new endpoint?

I think we should not change [Settings](https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst) doc, since we are not introducing any new setting in this PR. 

Please confirm.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/SqlPlug.java
abbashus: Good suggestion. But we can do it as part of another maintenance PR, because I am not sure about the implication of changing this name, as it is consumed by `plugin-description.properties` file, which in turn used by Elasticsearch for loading plugins.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
penghuo: No performance impact, I have a little bit concern that the current implementation coupled with ClusterUpdateSettingRequest internal implementation. It could be pain for long term maintenance.
As you said, maybe we can use copy the logic and only use the public interface if possible.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
```diff
@@ -0,0 +1,122 @@
...
+                    filterSettings(clusterUpdateSettingsRequest.transientSettings())
+            );
+            clusterUpdateSettingsRequest.persistentSettings(
+                    filterSettings(clusterUpdateSettingsRequest.persistentSettings())
+            );
+
+            return channel -> client.admin().cluster().updateSettings(
```
penghuo: Does this block operation on transport thread?

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
dai-chen: I'm thinking should we add an abstraction for this rather than exposing our internal implementation to customer directly. For example,

```
POST _opendistro/_sql/settings
{
    "defaultReponseFormat": "JSON"
}
```

abbashus: That works fine as well. Will change as suggested.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
abbashus: Let me confirm.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
dai-chen: Just my thought and open for discussion. Not sure if any potential problem and security concern with the current implementation. Just feel it would be safer and better to hide our implementation.

abbashus: My thoughts:

1. Yes we can abstract, but I don't see any major advantage of doing this. In terms of safety, we are already filtering based on `opendistro.sql.` prefix. Since the plugin is  open source, hiding implementation would be just security by obscurity.

2. On top of that, there would be major work in changing the response coming from backend, to make request/response consistent. These most likely will affect existing clients which are parsing the old response from `_cluster/settings` API.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
abbashus: Verified that this is not blocking operation on Transport thread, in two ways:

1. Our integration test cluster already has assertions enabled (as confirmed with @dai-chen ) and all settings related integration tests are passing.

2. Manually tested by adding `-ea` flag in `config/jvm.options`, and changed settings via Kibana Devtools console.

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlSettingsAction.java
abbashus: > No performance impact, I have a little bit concern that the current implementation coupled with ClusterUpdateSettingRequest internal implementation. It could be pain for long term maintenance.
> As you said, maybe we can use copy the logic and only use the public interface if possible.

Done.

dai-chen: @abbashus Got it. Do you know if this would work well with security plugin? Understood that the impact would be limited to SQL plugin only, but wonder if a customer without enough permission, let's say disable SQL plugin, can do it and impact others.

abbashus: > @abbashus Got it. Do you know if this would work well with security plugin? Understood that the impact would be limited to SQL plugin only, but wonder if a customer without enough permission, let's say disable SQL plugin, can do it and impact others.

Let me deep dive on this and confirm how this will work in tandem with security plugin, and if there are any security implications or not.

abbashus: @dai-chen I verified the behavior with security plugin. A user which does not have `cluster:admin/settings/update` permission cannot modify SQL settings by using new endpoint as underneath both requires same permissions.

dai-chen: Thanks a lot for verifying!

abbashus: Adding here for record , how the user will be denied changing SQL settings in absence of `cluster:admin/settings/update` permissions.


```
PUT /_opendistro/_sql/settings?pretty 
{
  "transient": {
    "opendistro.sql.query.slowlog": "20"
  }
}

{
  "error" : {
    "root_cause" : [
      {
        "type" : "security_exception",
        "reason" : "no permissions for [cluster:admin/settings/update] and User [name=abbas, backend_roles=[], requestedTenant=__user__]"
      }
    ],
    "type" : "security_exception",
    "reason" : "no permissions for [cluster:admin/settings/update] and User [name=abbas, backend_roles=[], requestedTenant=__user__]"
  },
  "status" : 403
}

```

Issue #413, opened 2020-04-07
Invalidate HTTP GET method


chloe-zh:
Currently SQL plugin is able to apply HTTP GET method to send requests with the queries/parameters laying in the url without encryption.
We are going to deprecate GET method out of the data security.

PR #414, opened 2020-04-07
Invalidate HTTP GET method

chloe-zh:
*Issue #, if available:*

* #413 

*Description of changes:*

* removed http get method
* added integtest

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,6 @@ The package uses the [Gradle](https://docs.gradle.org/4.10.2/userguide/userguide
 To use the feature, send requests to the `_opendistro/_sql` URI. You can use a request parameter or the request body (recommended).
 
 * Simple query
-```
-GET https://<host>:<port>/_opendistro/_sql?sql=select * from my-index limit 50
-```
 
 ```
 POST https://<host>:<port>/_opendistro/_sql
--- a/docs/user/interfaces/endpoint.rst
+++ b/docs/user/interfaces/endpoint.rst
@@ -16,21 +16,6 @@ Introduction
 
 To send query request to SQL plugin, you can either use a request parameter in HTTP GET or request body by HTTP POST request. POST request is recommended because it doesn't have length limitation and allows for other parameters passed to plugin for other functionality such as prepared statement. And also the explain endpoint is used very often for query translation and troubleshooting.
 
-GET
-===
-
-Description
------------
-
-You can send HTTP GET request with your query embedded in URL parameter.
-
-Example
--------
-
-SQL query::
-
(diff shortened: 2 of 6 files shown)


penghuo:
Is there any doc need to be updated?

zhongnansu:
> Is there any doc need to be updated?

I guess at least README also need to update

abbashus:
Why is only POST request secure? 

Also the whole HTTP payload  (including HTTP verb , URL,  headers, body) is encrypted not just the body.

So in the absence of HTTPS, both GET and POST will reveal sensitive information, and in presence of HTTPS both GET and POST will equally hide the sensitive information, because the whole payload is encrypted.

Since this is a breaking change (client using GET requests will fail after this change), we should 
have a strong and valid case to make this change.

Please correct if I am wrong or missing something.

https://en.wikipedia.org/wiki/HTTPS

> Because HTTPS piggybacks HTTP entirely on top of TLS, the entirety of the underlying HTTP protocol can be encrypted. This includes the request URL (which particular web page was requested), query parameters, headers, and cookies (which often contain identity information about the user)

chloe-zh:
> Is there any doc need to be updated?

Hi @penghuo @zhongnansu , thank you for the reminder. README and the reference manual are updated now. Thanks!

chloe-zh:
> Why is only POST request secure?
> 
> Also the whole HTTP payload (including HTTP verb , URL, headers, body) is encrypted not just the body.
> 
> So in the absence of HTTPS, both GET and POST will reveal sensitive information, and in presence of HTTPS both GET and POST will equally hide the sensitive information, because the whole payload is encrypted.
> 
> Since this is a breaking change (client using GET requests will fail after this change), we should
> have a strong and valid case to make this change.
> 
> Please correct if I am wrong or missing something.
> 
> https://en.wikipedia.org/wiki/HTTPS
> 
> > Because HTTPS piggybacks HTTP entirely on top of TLS, the entirety of the underlying HTTP protocol can be encrypted. This includes the request URL (which particular web page was requested), query parameters, headers, and cookies (which often contain identity information about the user)

Thanks @abbashus  for raising your concerns. 
You're correct that both GET and POST on HTTPS are encrypted while on HTTP are revealing data. And we can encrypt the transmission using SSL/TSL. 
However, to my understanding, HTTP GET method might expose the info in the url to browser plugins or even like the browser history, and the server log etc. so that it gets the risk of leaking out data. But POST method passes the queries in request body, which is safe to use HTTP on SSL/TSL.

abbashus:
@chloe-zh From browser point of view, it makes sense. But w.r.t to server, it not guaranteed, completely upto server what it wants to do with the HTTP payload.

PR #415, opened 2020-04-08
[PPL] add basic parser support

penghuo:
### Spring
integrate with Spring IOC
### PPL
1. add the basic ANTLR
2. add the parser
### elasticsearch
1. add the new module for elasticsearch related operation.
### common
1. add the common module for shard utils.


plugin/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -44,17 +48,23 @@ public String getName() {
...
-            }
-
-            @Override
-            public void onFailure(Exception e) {
-                channel.sendResponse(new BytesRestResponse(OK, "application/json; charset=UTF-8", "error"));
-            }
-        });
+        AnnotationConfigApplicationContext context = SecurityAccess.doPrivileged(
+                () -> new AnnotationConfigApplicationContext(
+                        PPLServiceConfig.class));
```
dai-chen, line 53: Just curious, is this privilege call required for all classes annotated by Spring annotation?

penghuo: If we use Spring IoC, it should be fine. But for other features, I am not sure now.

dai-chen: I see. Thanks!

Issue #97, opened 2019-06-26
RestSqlAction logs raw SQL queries during normal execution flow


galkk:
The RestSqlAction logs raw SQL queries with [INFO severity.](https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java#L100) This could expose customer information.

The possible options that immediately come to mind

- Move raw query logging into Exception handling
- In normal execution path, the queries should be logged using Debug severity
- Obfuscate all raw parameters
- Move all raw parameters into some secured storage with controlled access and replace values with identifiers

Anythign else?

PR #419, opened 2020-04-09
Anonymize sensitive data in queries exposed to RestSqlAction logs

chloe-zh:
*Issue #, if available:*

* [**#97 RestSqlAction logs raw SQL queries during normal execution flow**](https://github.com/opendistro-for-elasticsearch/sql/issues/97)

*Description of changes:*

* Created a rewriting sensitive data rule
* Created a sensitive data anonymizer
* Changed the log message for the `incoming request` to a request with sensitive data gone
* Added UT


src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/QueryDataMask.java
```diff
@@ -0,0 +1,83 @@
...
+
+import com.alibaba.druid.sql.ast.expr.SQLQueryExpr;
+import com.amazon.opendistroforelasticsearch.sql.rewriter.identifier.RemoveSensitiveDataRule;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.amazon.opendistroforelasticsearch.sql.utils.Util.toSqlExpr;
+
+/**
+ * Utility class for mask sensitive info in incoming sql queries
```
abbashus, line 26: np:
`for` -> `to`
`info` -> `information`
`sql` -> `SQL`

Same elsewhere.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
```diff
@@ -105,4 +107,15 @@ public BoolQueryBuilder checkAndAddFilter(BoolQueryBuilder boolQuery) throws Sql
...
     }
+
+    public String removeSensitiveData() {
+        String query = "";
+        try {
+            query = QueryDataMask.maskData(this.sql);
+        } catch (Exception e) {
+            LogManager.getLogger().error("Caught an exception when removing sensitive data: "
+                    + e.getClass().getSimpleName());
+        }
```
abbashus, line 118: Consider using 

 ```
private static final Logger LOG = LogManager.getLogger(SqlRequest.class);

...

 LOG.error("Caught an exception when masking sensitive data ",  e)

```

Let the stack trace be logged for faster debugging, instead of Exception name.

File: src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/QueryDataMaskTest.java [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+        String expectedQuery = "SELECT *** FROM ***";
+        Assert.assertEquals(maskedQuery, expectedQuery);
+    }
+
+    @Test
+    public void selectQueriesWithQuotedIdentifiersWithSpacesShouldMaskSensitiveData() {
+        String maskedQuery = QueryDataMask.maskData("SELECT `a   g e` FROM accounts");
+        String expectedQuery = "SELECT *** FROM ***";
+        Assert.assertEquals(maskedQuery, expectedQuery);
+    }
```
abbashus, line 50: Does this only works for simple queries?
Can you add tests for nested, subquery, join and union/minus queries as well for better coverage.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/QueryDataMaskTest.java
```diff
@@ -0,0 +1,51 @@
...
+        String maskedQuery = QueryDataMask.maskData("SELECT age FROM accounts");
+        String expectedQuery = "SELECT *** FROM ***";
+        Assert.assertEquals(maskedQuery, expectedQuery);
+    }
+
+    @Test
+    public void selectQueriesShouldMaskAllSensitiveData() {
+        String query = "SELECT ABS(balance), lastname FROM accounts WHERE age > 30";
+        String maskedQuery = QueryDataMask.maskData(query);
+        Assert.assertEquals(maskedQuery, "SELECT ABS(***), *** FROM *** WHERE *** > 30");
```
penghuo, line 35: Question, what is definition sensitive data? age or 30?
IMO, the value is more sensitive than field. e.g. password = '123456', it more make sense to mask the value like password = 'xxx'

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/StringUtils.java
```diff
@@ -126,6 +126,11 @@ public static boolean isNumeric(String text) {
         return Doubles.tryParse(text) != null;
     }
 
+    public static String getFirstWord(String sql) {
```
penghuo, line 129: add UT

src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/RestSqlAction.java
```diff
@@ -112,7 +112,8 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
             }
 
             final SqlRequest sqlRequest = SqlRequestFactory.getSqlRequest(request);
-            LOG.info("[{}] Incoming request {}: {}", LogUtils.getRequestId(), request.uri(), sqlRequest.getSql());
+            LOG.info("[{}] Incoming request {}: {}", LogUtils.getRequestId(), request.uri(),
+                    sqlRequest.removeSensitiveData());
```
penghuo, line 116: why adding a method on SqlRequest? could we use QueryDataMask.mask(sqlRequest.sql) instead.

penghuo:
1. Do we need the setting to control this feature?
2. How this solution could be used in other place when logging is needed?

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/QueryDataMaskTest.java
dai-chen: Agree. I think we can just mask all values (rewrite literals, number, string etc.)

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/QueryDataMask.java
chloe-zh: Will do, thanks!

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
chloe-zh: Sounds good, will do. Thanks!

abbashus:
Open to discussion:

I assume more work, but instead of masking, why not anonymize the identifiers and literals. This way we will not loose the semantics.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/identifier/RemoveSensitiveDataRule.java
```diff
@@ -0,0 +1,48 @@
...
+    public boolean match(SQLQueryExpr expr) {
+        return true;
+    }
+
+    @Override
+    public void rewrite(SQLQueryExpr expr) {
+        expr.accept(this);
+    }
+
+    public Set<String> getIdentifierSet() {
```
abbashus, line 45: np: `getIdentifierSet ` -> `getDataSet `?

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
```diff
@@ -105,4 +109,14 @@ public BoolQueryBuilder checkAndAddFilter(BoolQueryBuilder boolQuery) throws Sql
         }
         return boolQuery;
     }
+
+    public String removeSensitiveData() {
+        String query = "";
+        try {
+            query = QueryDataMask.maskData(this.sql);
```
dai-chen, line 116: As Peng commented, I think you can either move all mask logic here or just call `maskData()` outside along with `getSql()`.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/identifier/RemoveSensitiveDataRule.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class RemoveSensitiveDataRule extends MySqlASTVisitorAdapter implements RewriteRule<SQLQueryExpr> {
+
+    private Set<String> dataSet = new HashSet<>();
+
+    @Override
+    public boolean visit(SQLIdentifierExpr identifierExpr) {
+        dataSet.add(identifierExpr.getName());
+        return true;
+    }
```
dai-chen, line 33: Because values (literals) in query is more sensitive, you may need to rewrite methods like `visit(SQLNumberExpr x)`, `visit(SQLCharExpr x)` etc. We can add a few UTs to cover common literals such as string, numerical, bool, datetime etc.

src/main/java/com/amazon/opendistroforelasticsearch/sql/utils/QueryDataMask.java
```diff
@@ -0,0 +1,83 @@
...
+import com.amazon.opendistroforelasticsearch.sql.rewriter.identifier.RemoveSensitiveDataRule;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.amazon.opendistroforelasticsearch.sql.utils.Util.toSqlExpr;
+
+/**
+ * Utility class to mask sensitive information in incoming SQL queries
+ */
+public class QueryDataMask {
```
dai-chen, line 28: I think you can move all logic in this class into `RemoveSensitiveDataRule` except rebuilding. (rename it to `MaskSensitiveDataRule`?) Then you just need a util method for rebuilding AST (after applied our masking rule) to a query string.

src/main/java/com/amazon/opendistroforelasticsearch/sql/request/SqlRequest.java
```diff
@@ -37,7 +37,6 @@
     String sql;
     JSONObject jsonContent;
 
-
```
penghuo, line 40: unnecessary change

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/QueryDataAnonymizerTest.java
```diff
@@ -0,0 +1,135 @@
...
+    @Test
+    public void queriesShouldHaveAnonymousFieldAndIndex() {
+        String query = "SELECT ABS(balance) FROM accounts WHERE age > 30 GROUP BY ABS(balance)";
+        String expectedQuery = "(\n" +
+                "\tSELECT ABS(identifier)\n" +
+                "\tFROM table\n" +
+                "\tWHERE identifier > number\n" +
+                "\tGROUP BY ABS(identifier)\n" +
+                "\t)";
+        Assert.assertEquals(expectedQuery, QueryDataAnonymizer.anonymizeData(query));
```
penghuo, line 33: Curious what will be the output in log4j, does it output multiple line for the same query?
If so, I would suggest to remove the newline to make it in same line. Because usually, we process the log data by using line based tools and group the data with request id. e.g., grep requestId=xxx which will return all the log corresponding to specific request.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/identifier/AnonymizeSensitiveDataRule.java
```diff
@@ -0,0 +1,73 @@
...
+import com.alibaba.druid.sql.ast.expr.SQLCharExpr;
+import com.alibaba.druid.sql.ast.expr.SQLIdentifierExpr;
+import com.alibaba.druid.sql.ast.expr.SQLIntegerExpr;
+import com.alibaba.druid.sql.ast.expr.SQLNumberExpr;
+import com.alibaba.druid.sql.ast.expr.SQLQueryExpr;
+import com.alibaba.druid.sql.ast.statement.SQLExprTableSource;
+import com.alibaba.druid.sql.dialect.mysql.visitor.MySqlASTVisitorAdapter;
+import com.amazon.opendistroforelasticsearch.sql.rewriter.RewriteRule;
+
+public class AnonymizeSensitiveDataRule extends MySqlASTVisitorAdapter implements RewriteRule<SQLQueryExpr> {
```
dai-chen, line 28: Please add some doc here.

src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/utils/QueryDataAnonymizerTest.java
chloe-zh: yes, it outputs multiple lines and tabs in the front of every line as well. ok that makes sense, i will remove the redundant returns and tabs.

src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/identifier/AnonymizeSensitiveDataRule.java
chloe-zh: Will do, thanks!

Issue #424, opened 2020-04-10
Subquery in FROM Clause fails


zhongnansu:
#### Issue description
This is an example query from the doc [Example 2: Subquery in FROM Clause](https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/dql/complex.rst#example-2-subquery-in-from-clause)
```
POST /_opendistro/_sql/
{
"query" : """
        SELECT a.f, a.l, a.a
        FROM (
          SELECT firstname AS f, lastname AS l, age AS a
          FROM account
          WHERE age > 30
        ) AS a
        
        """
}
```

#### Error message
```
{
  "error": {
    "reason": "Error occurred in Elasticsearch engine: Different mappings are not allowed for the same field[age]: found [{type:integer}] and [{type:long}] ",
    "details": "com.amazon.opendistroforelasticsearch.sql.rewriter.matchtoterm.VerificationException: Different mappings are not allowed for the same field[age]: found [{type:integer}] and [{type:long}] \nFor more details, please send request for Json format to see the raw response from elasticsearch engine.",
    "type": "VerificationException"
  },
  "status": 400
}
```
#### Test data
[accounts.json](https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/test/resources/doctest/testdata/accounts.json)

penghuo:
same issue with #355

Issue #355, opened 2020-01-23
Mapping Compatibility Check Issue


penghuo:
## Issue Description
The following query has subquery in from clause.
```
GET /_opendistro/_sql/
{
  "query": """
      SELECT t.dcn
      FROM (SELECT f1.AvgTicketPrice as dcn 
          FROM kibana_sample_data_flights f1) t
  """
}
```
But it throw mapping conflict exception
```
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "Different mappings are not allowed for the same field[gender]: found [{type:text,fields:{keyword:{type:keyword}}}] and [{type:text,fielddata:true}] ",
    "type": "VerificationException"
  },
  "status": 400
}
```
## Expectation
Alt1. The query should running successfully without error.
Alt2. The engine should throw meaningful exception if the query is not supported.

penghuo:
## RCA
The kibana_sample_data_flights index doesn't have gender field. But there are two other indices in the cluster which have gender field with different type. These gender mapping is exactly what the exception complained.
```
// gender mapping in index 1
        "gender" : {
          "type" : "text",
          "fielddata" : true
        }

// gender mapping in index 2
        "gender" : {
          "type" : "text",
          "fields" : {
            "keyword" : {
              "type" : "keyword"
            }
          }
```

The [getMappings](https://github.com/opendistro-for-elasticsearch/sql/blob/239908b37665b78709305bebc7ff996607dcfa64/src/main/java/com/amazon/opendistroforelasticsearch/sql/rewriter/matchtoterm/TermFieldRewriter.java#L306) method defined in TermFieldRewriter try to load all the indices mapping from cluster when allIndexes is empty.
```
    public IndexMappings getMappings(Map<String, String> indexToType) {
        String[] allIndexes = indexToType.keySet().stream().toArray(String[]::new);
        // GetFieldMappingsRequest takes care of wildcard index expansion
        return LocalClusterState.state().getFieldMappings(allIndexes);
    }
```
Then, the checkMappingCompatibility function will check the mapping compatibility cross the cluster which is an unexpected behavior.

PR #425, opened 2020-04-11
Bug fix, ignore the term query rewrite if there is no index found

penghuo:
*Issue #, if available:* #355, #424 

*Description of changes:*


src/test/java/com/amazon/opendistroforelasticsearch/sql/unittest/rewriter/term/TermFieldRewriterTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        public String ignoreCase(String toBeLower) {
+            return toBeLower.toLowerCase();
+        }
+
+        public static Matcher<String> equalToIgnoreCaseAndWhiteSpace(String expectedString) {
+            return new IsEqualIgnoreCaseAndWhiteSpace(expectedString);
+        }
+    }
+}
```
chloe-zh, line 124: Would it be better to move it to utils/stringutils?

penghuo: Make sense. Done.

PR #444, opened 2020-04-28
Feature/ppl logical plan

penghuo:
# Analyzer

The analyzer take the unresolved plan and unresolved expression as input and generate the resolved logical plan and resolved expression. while doing the analysis, the analyzer also take the responsibility to do the semantic check. The analyzer should:

1. Translate the unresolved plan to resolved logical plan.
2. Translate the unresolved expression to resolved expression.
3. Check the field is valid in the context.
4. Check the function/operator is valid in the context. 

In the following section, we will target the design of each component in detail.

## Analysis Context

The analyzer is composed by the environment analysis context and keep refreshing the analysis context during the analyzer procedure.
The analysis context include two environments

* Type Environment, include the expression and expression type mapping.
* Function Environment, include the built-in function mapping.

### Type Environment

The type environment define the mapping between field and field type. The specification of type environment is as below:

```
Constructors
empty-env: () -> Env
extend-env: Expression x ExprType x Env -> Env

Observer:
apply: Env x Expression -> ExprType
```

The analyzer require the type environment as the input. All the fields could been resolved in the type environment, if not, the `semantic analysis exception` will been thrown. The type environment could also been extended by new expression and type define during the analysis process.
Let’s take an example to explain the process, considering the sample query: `search source=access_log | rename action as a, status as s | where s == 200`.
The query will been translate to the following unresolved plan.
[Image: Screen Shot 2020-04-24 at 8.30.44 AM.png]
Then, the analyzer resolve the unresolved plan bottom to up. 
The analyzer is initialized with scheme which define the mapping of field and type.

* Firstly, analyze the relation, it will load the mapping of field and type from scheme and extend the empty type env.
* Secondly, analyze the rename, it will extend the type env by new definition of <a:STRING> and <s:INT>. 
* Thirdly, analyze the where, it will resolve the type of “s” from type env.


chloe-zh:
Could you remove core/out directory if they were unnecessary

PR #469, opened 2020-05-15
Implement get index mapping in ES storage engine

dai-chen:
*Issue #, if available:*

*Description of changes:* Add get index mapping logic to Elasticsearch storage engine. The method in ES storage is used by query engine to fetch index mapping and do analysis, such as semantic check.

*Implementation*:

 1. **ElasticsearchStorageEngine**: Storage engine implementation for Elasticsearch which simply returns *ElasticsearchIndex* for now.
 2. **ElasticsearchIndex**: Implement get index mapping (field types) method by `ElasticsearchClient` underlying.
 3. **ElasticsearchClient**: Abstraction for different Elasticsearch clients. In our plan, apart from a `ElasticsearchNodeClient`, another `ElasticsearchRestClient` will be added soon for entire engine running outside Elasticsearch container.

*Testing*: Added and passed all unit test and coverage required. Integration test will be added later when executor is in place.


elasticsearch/build.gradle
```diff
@@ -1,10 +1,65 @@
...
+            limit {
+                minimum = 1.0
+            }
+
+        }
+    }
+    afterEvaluate {
+        classDirectories.setFrom(files(classDirectories.files.collect {
+            fileTree(dir: it,
+                    exclude: ['**/ast/**'])
```
penghuo, line 60: remove this?

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/mapping/IndexMapping.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param transform     transform function to transform field type in string to another type
+     * @param <Type>        expected field type class
+     * @return              mapping from field name to field type
+     */
+    public <Type> Map<String, Type> getAllFieldTypes(Function<String, Type> transform) {
+        return fieldMappings.entrySet().
+                             stream().
+                             collect(Collectors.toMap(
+                                 Map.Entry::getKey,
+                                 e -> transform.apply(e.getValue())
```
penghuo, line 78: The null value should be filtered out.

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/mapping/IndexMapping.java
```diff
@@ -0,0 +1,124 @@
...
+    public int size() {
+        return fieldMappings.size();
+    }
+
+    /**
+     * Return field type by its name.
+     * @param fieldName     field name
+     * @return              field type in string
+     */
+    public String getFieldType(String fieldName) {
```
penghuo, line 63: Is it possible to return null value? or we accept null value as return value?

elasticsearch/build.gradle
dai-chen: Sure. Copy-pasted by mistake.

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/mapping/IndexMapping.java
dai-chen: Yeah, it is. For now this is used in UT. Will add an existential check method or refactor it to return `Optional` if in use in future.

elasticsearch/build.gradle
dai-chen: Added JavaDoc and UT. Thanks.

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/mapping/IndexMapping.java
dai-chen: Thanks. Double checked there is only null value in Elasticsearch so we should be good. https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-data-types.html

PR #476, opened 2020-05-20
[PPL] Add simple json formatter

dai-chen:
*Issue #, if available:*

*Description of changes:* Add a new protocol module for query request parsing and response formatting. Note that the reason of not use existing common module is due to dependency cycle between core and common module.


protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/QueryResponse.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Parse column name from results
+     * @return  mapping from column names to its expression type
+     */
+    public Map<String, String> columnNameTypes() {
+        if (exprValues.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
+        Map<String, ExprValue> tupleValue = getFirstTupleValue();
```
penghuo, line 57: it is possible the ExprValue has missing value, in this case, the type value may missing.
e.g. [{"name": "susan"}, {"name": "bob", "age":30}].
it may required the type info from the planner?

penghuo: let's add Todo here.

dai-chen: Sure, I'm adding comment and a disabled UT. Thanks!

Issue #360, opened 2020-01-29
Possible to config the SQL in a standalone server to connect to an ElasticSearch server?


YouXiang-Wang:
Actually, in many cases, we may use the SAAS service provided by AWS/Azure/GCP etc. We could not install the ElasticSearch plugins freely.  So I am wondering if it is possible to config the SQL in a standalone server to connect to an ElasticSearch server?

YouXiang-Wang:
@abbashus Do we have the plan to support this big feature?

abbashus:
It's right now in a Elasticsearch plugin from, and thus can only be bootstrapped alongside Elasticsearch. We may in the future break down the SQL plugin functionality in a standalone consumable form.

abbashus:
We will certainly evaluate the need for such a feature. Currently we don't have a ETA on this.

dai-chen:
@YouXiang-Wang We've added standalone mode in our new query engine in PR https://github.com/opendistro-for-elasticsearch/sql/pull/491. You can find sample code in https://github.com/opendistro-for-elasticsearch/sql/blob/develop/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/StandaloneIT.java, though our engine code is not available on Maven repo yet. We will evaluate how many users need this and put efforts to make it public.

Issue #461, opened 2020-05-07
Incorrect column in CSV format when select field is function


chloe-zh:
Example: 

Index: `kibana_sample_data_flights`

Query:
```
POST _opendistro/_sql?format=csv
{
    "query": "select log(AvgTicketPrice) from kibana_sample_data_flights limit 5"
}
```

Result:
```
log(AvgTicketPrice),FlightNum,Origin,OriginLocation,DestLocation,FlightDelay,DistanceMiles,FlightTimeMin,OriginWeather,dayOfWeek,AvgTicketPrice,Carrier,FlightDelayMin,OriginRegion,FlightDelayType,DestAirportID,timestamp,Dest,FlightTimeHour,Cancelled,DistanceKilometers,OriginCityName,DestWeather,OriginCountry,DestCountry,DestRegion,OriginAirportID,DestCityName
6.472336294873528,JHB3J5V,Melbourne International Airport,"{lon=144.843002, lat=-37.673302}","{lon=-93.22180176, lat=44.88199997}",false,9441.687300483349,1012.9948537939384,Heavy Fog,4,646.9935382543698,Kibana Airlines,0,SE-BD,No Delay,MSP,2020-04-03T01:03:05,Minneapolis-St Paul International/Wold-Chamberlain Airport,16.883247563232306,false,15194.922806909077,Melbourne,Clear,AU,US,US-MN,MEL,Minneapolis
6.839912084672193,GQTQ1B3,London Gatwick Airport,"{lon=-0.190277994, lat=51.14810181}","{lon=141.6920013, lat=42.77519989}",false,5571.180557608165,471.89189491070294,Heavy Fog,4,934.4069613948874,ES-Air,0,GB-ENG,No Delay,CTS,2020-04-03T14:03:40,New Chitose Airport,7.864864915178383,false,8965.946003303356,London,Cloudy,GB,JP,SE-BD,LGW,Chitose / Tomakomai
5.8188328302884935,80V65VD,Shanghai Hongqiao International Airport,"{lon=121.3359985, lat=31.19790077}","{lon=12.3519, lat=45.505299}",false,5536.339466666921,424.27974774493384,Clear,4,336.5789802442783,ES-Air,0,SE-BD,No Delay,VE05,2020-04-03T06:55:15,Venice Marco Polo Airport,7.071329129082231,false,8909.87470264361,Shanghai,Hail,CN,IT,IT-34,SHA,Venice
[...]

PR #484, opened 2020-05-22
[PPL] Add ES rest client to support standalone mode

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/360

*Description of changes:* In this PR, a new implementation of our `ElasticsearchClient` is added to support running query engine outside Elasticsearch cluster. Meanwhile, a `cleanup` method is added to client so we can make sure resources related to a search, for example scroll context, is released if any exception.

### DIFF
--- a/elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/ElasticsearchIndexScan.java
+++ b/elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/ElasticsearchIndexScan.java
@@ -63,6 +63,8 @@ public class ElasticsearchIndexScan extends TableScanOperator {
 
     @Override
     public void open() {
+        super.open();
+
         // For now pull all results immediately once open
         List<ElasticsearchResponse> responses = new ArrayList<>();
         ElasticsearchResponse response = client.search(request);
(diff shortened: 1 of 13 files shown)


elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/ElasticsearchIndexScan.java
```diff
@@ -83,4 +85,11 @@ public ExprValue next() {
         return ExprValueUtils.fromObjectValue(hits.next().getSourceAsMap());
```
chloe-zh, line 85: Not a problem of this pr, but sometimes the result of method `getSourceAsMap()` is not what we expected, e.g. the source would be all fields when the designated field(s) is an evaluation function (like `abs(balance)`). Here is a similar issue: https://github.com/opendistro-for-elasticsearch/sql/issues/461. Are we likely running into the same problem here?

dai-chen: Yeah, you're right, this is only for the time being. Need to handle different cases in future.

PR #485, opened 2020-05-22
[PPL] Support dedup command

penghuo:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/planner/physical/DedupeOperator.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/planner/physical/DedupeOperator.java
@@ -1,0 +1,146 @@
+/*
+ *   Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.planner.physical;
+
+import com.amazon.opendistroforelasticsearch.sql.data.model.ExprValue;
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/planner/physical/DedupeOperator.java
```diff
@@ -0,0 +1,160 @@
...
+  /**
+   * The Consecutive Deduper monitor the duplicated element with consecutive seen value. It means
+   * only the consecutive duplicated value will be counted.
+   */
+  static class ConsecutiveDeduper<K> implements Deduper<K> {
+    private K lastSeenDedupeKey = null;
+    private Integer consecutiveCount = 0;
+
+    @Override
+    public int seenTimes(K dedupeKey) {
```
dai-chen, line 150: Notice that `HistoricalDeduper` is using `ConcurrentHashMap`, so just wonder if thread safe is also needed here by synchronized or AtomicReference?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/planner/physical/DedupeOperator.java
```diff
@@ -0,0 +1,160 @@
...
+   * @param value
+   * @return true: keep, false: ignore
+   */
+  public boolean keep(ExprValue value) {
+    BindingTuple bindingTuple = value.bindingTuples();
+    ImmutableList.Builder<ExprValue> dedupeKeyBuilder = new ImmutableList.Builder<>();
+    for (Expression expression : dedupeList) {
+      ExprValue exprValue = expression.valueOf(bindingTuple);
+      if (NULL_OR_MISSING.test(exprValue)) {
+        return keepEmpty;
```
dai-chen, line 110: Probably safer to do null check before return in case this `Boolean` is null? Or verify this in constructor and save as `boolean` if it's required non-null?

penghuo: Done. add the @NonNull at the constructor.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/planner/physical/DedupeOperator.java
penghuo: Done. Refactor the Deduper. Get rid of hierarchy

dai-chen: Thanks!

PR #493, opened 2020-05-27
[PPL] Support Count aggregator and OR operator

penghuo:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AggregatorFunction.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AggregatorFunction.java
@@ -41,6 +41,7 @@ public class AggregatorFunction {
     public static void register(BuiltinFunctionRepository repository) {
         repository.register(avg());
         repository.register(sum());
+        repository.register(count());
     }
 
     private static FunctionResolver avg() {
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
```diff
@@ -29,7 +29,7 @@
 import static com.amazon.opendistroforelasticsearch.sql.utils.ExpressionUtils.format;
 
 /**
- * The average aggregator aggregate the value evaluated by the expression.
+ * The count aggregator aggregate the value evaluated by the expression.
```
dai-chen, line 32: Is this meant for JavaDoc on `CountAggregator `?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AggregatorFunction.java
```diff
@@ -54,6 +55,31 @@ private static FunctionResolver avg() {
         );
     }
 
+    private static FunctionResolver count() {
+        FunctionName functionName = BuiltinFunctionName.COUNT.getName();
+        return new FunctionResolver(
+            functionName,
+            new ImmutableMap.Builder<FunctionSignature, FunctionBuilder>()
+                .put(new FunctionSignature(functionName, Collections.singletonList(ExprType.INTEGER)),
```
dai-chen, line 63: Np: since the only difference is input type, this can be simplified into a for loop on input type array.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
penghuo: done.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AggregatorFunction.java
penghuo: totally agree, let's fix the bug firstly, then think how we could better define built in funciton.

dai-chen: Sounds good!

PR #496, opened 2020-05-27
[PPL] Added integration tests for query analysis and search, where, fields commands

chloe-zh:
*Issue #, if available:*

*Description of changes:*

- Changed the default value of `argList` in `Field` from `null` to empty list to avoid `NullPointerException`, and added `hasArgument()` method to `Field`, `Project` and `Aggregation` nodes.
- Added integration tests for query analysis and search, where, fields commands

### DIFF
--- a/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
+++ b/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
@@ -1,0 +1,131 @@
+/*
+ *   Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.ppl;
+
+import com.amazon.opendistroforelasticsearch.sql.exception.SemanticCheckException;
+import java.io.IOException;
+import org.elasticsearch.client.Request;
+import org.elasticsearch.client.Response;
+import org.elasticsearch.client.ResponseException;
+import org.junit.Ignore;
+import org.junit.Test;
+import static com.amazon.opendistroforelasticsearch.sql.esintgtest.TestsConstants.TEST_INDEX_ACCOUNT;
+import static org.elasticsearch.rest.RestStatus.OK;
+
+public class QueryAnalysisIT extends PPLIntegTestCase {
+
+    @Override
+    public void init() throws IOException {
+        loadIndex(Index.ACCOUNT);
+    }
+
+    /**
+     * Valid commands should pass semantic check
+     */
+    @Test
+    public void searchCommandShouldPassSemanticCheck() {
+        String query = String.format("search source=%s age=20", TEST_INDEX_ACCOUNT);
+        queryShouldPassSemanticCheck(query);
+    }
+
+    @Test
+    public void whereCommandShouldPassSemanticCheck() {
+        String query = String.format("search source=%s | where age=20", TEST_INDEX_ACCOUNT);
+        queryShouldPassSemanticCheck(query);
+    }
+
+    @Test
+    public void fieldsCommandShouldPassSemanticCheck() {
+        String query = String.format("search source=%s | fields firstname", TEST_INDEX_ACCOUNT);
+        queryShouldPassSemanticCheck(query);
+    }
+
+    @Ignore("Can't resolve target field yet")
+    @Test
+    public void renameCommandShouldPassSemanticCheck() {
+        String query = String.format("search source=%s | rename firstname as first", TEST_INDEX_ACCOUNT);
+        queryShouldPassSemanticCheck(query);
+    }
+
+    @Test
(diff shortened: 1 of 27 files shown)


integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
```diff
@@ -0,0 +1,131 @@
...
+        String query = String.format("search source=%s | fields name", TEST_INDEX_ACCOUNT);
+        queryShouldThrowSemanticException(query, "can't resolve expression name in type env");
+    }
+
+
+    private void queryShouldPassSemanticCheck(String query) {
+        Request request = buildRequest(query);
+        try {
+            Response response = client().performRequest(request);
+            assertEquals(OK.getStatus(), response.getStatusLine().getStatusCode());
```
dai-chen, line 108: This is same as `executeQuery`? Probably we can just call it and discard return value?

```
    protected String executeQuery(String query) throws IOException {
        Response response = client().performRequest(buildRequest(query));
        Assert.assertEquals(200, response.getStatusLine().getStatusCode());
        return getResponseBody(response, true);
    }
```

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
```diff
@@ -0,0 +1,131 @@
...
+        queryShouldPassSemanticCheck(query);
+    }
+
+
+    /**
+     * Commands that fail semantic analysis should throw {@link SemanticCheckException}
+     */
+    @Test
+    public void unsupportedAggregationShouldFailSemanticCheck() {
+        String query = String.format("search source=%s | stats count(age)", TEST_INDEX_ACCOUNT);
```
dai-chen, line 93: Is count being added in https://github.com/opendistro-for-elasticsearch/sql/pull/493?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
```diff
@@ -0,0 +1,131 @@
...
+        queryShouldThrowSemanticException(query, "Unsupported aggregation function count");
+    }
+
+    @Test
+    public void nonexistentIndexShouldFailSemanticCheck() {
+        String query = String.format("search source=%s | fields name", TEST_INDEX_ACCOUNT);
```
dai-chen, line 99: The test method name should be nonexistent field? But I think we do need another one for index.

chloe-zh: It should be nonexistent field, corrected it now. It seems we have not set semantic check for index, let's add the index check when the index check is enabled.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
chloe-zh: Thanks for the reminder. Changed the case to `range`, is that making sense?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
chloe-zh: Done, thanks!

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
dai-chen: The `ElasticsearchIndex.getFieldTypes()` should throw Elasticsearch IndexNotFoundException. Not sure if the exception is caught without re-throw or for other reason.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
dai-chen: Sure. Thanks!

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
chloe-zh: Yes Elasticsearch throws this exception. I didn't add it here because it's not handled by the semantic check. Maybe we can add this to other test

dai-chen: Sounds good!

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/SearchCommandIT.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void init() throws IOException {
+        loadIndex(Index.BANK);
+    }
+
+    @Test
+    public void testSearchAllFields() throws IOException {
+        String result = executeQuery(String.format("search source=%s", TEST_INDEX_BANK));
+        assertTrue(result.contains("\"total\": 7"));
```
penghuo, line 33: Is it possible to assert all the response instead of just one field. e.g.
schema(contains("name", "string")))
or we can assert the full string match.
Actually, we don't need much data for integration test purpose, IMO, the sample data include all the types, null, missing should be enough for IT purpose.

chloe-zh: I'm considering to add a utils to build result set and schema from the result, so that we can test the behavior in a more accurate way.
For the sample data, what about applying `bank_with_null_values` (already in our resources) to all tests?

Issue #517, opened 2020-06-11
[PPL] Support more readable column name of aggregation with expression as argument


chloe-zh:
Query example: `source=accounts | stats avg(abs(age)*2)`

Actual
```
{
  "schema": [{
    "name": "avg(com.amazon.opendistroforelasticsearch.sql.expression.operator.OperatorUtils$1@ab2ac68)",
    "type": "double"
  }],
  "total": 1,
  "datarows": [[60.342]],
  "size": 1
}
```

Expected
```
{
  "schema": [{
    "name": "avg(abs(age)*2)",
    "type": "double"
  }],
  "total": 1,
  "datarows": [[60.342]],
  "size": 1
}
```

PR #518, opened 2020-06-11
[PPL] Support expression and boolean operators

chloe-zh:
*Issue #, if available:*

* #517 

*Description of changes:*

**1. Support Boolean operators**

Boolean operators supported in the pull request include:
- less than (<)
- less than or equal to (<=)
- greater than (>)
- greater than or equal to (>=)
- not equal to (!=)

Additionally supplemented operators in analyzer:
- NOT in analyzer
- XOR in parser and analyzer

**2. Corrected column name of nested functions/operators** (related: #517)

Example: `source=accounts | stats avg(abs(age))`
Before: returned column name was in pattern `avg(object@#)`
After: column name is `avg(abs(age))`


integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/StatsCommandIT.java
```diff
@@ -81,4 +80,22 @@ public void testStatsCount() throws IOException {
   }
 
   // TODO: each stats aggregate function should be tested here when implemented
+
+  @Test
+  public void testStatsNested() throws IOException {
+    String result =
+        executeQueryToString(
+            String.format("source=%s | stats avg(abs(age)*2) as AGE", TEST_INDEX_ACCOUNT));
+    assertEquals(
```
penghuo, line 89: when should we use plan text comparsion? when shoud with use matchers like verifyColumn?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/OperatorIT.java
```diff
@@ -0,0 +1,225 @@
...
+                TEST_INDEX_BANK));
+    verifyDataRows(result, rows(32), rows(33));
+  }
+
+  @Test
+  public void testModuleOperator() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | where age %s 32 = 0 | fields age",
```
penghuo, line 78: where age %s 32 = 0 ? Moudlo? age % 32

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/OperatorIT.java
```diff
@@ -0,0 +1,225 @@
...
+import static com.amazon.opendistroforelasticsearch.sql.legacy.TestsConstants.TEST_INDEX_BANK;
+import static com.amazon.opendistroforelasticsearch.sql.util.MatcherUtils.rows;
+import static com.amazon.opendistroforelasticsearch.sql.util.MatcherUtils.verifyDataRows;
+
+import java.io.IOException;
+import org.json.JSONObject;
+import org.junit.Ignore;
+import org.junit.jupiter.api.Test;
+
+public class OperatorIT extends PPLIntegTestCase {
```
penghuo, line 27: consider to add null and missing related test cases.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
```diff
@@ -86,12 +85,6 @@ public void evalCommandShouldPassSemanticCheck() {
   /**
    * Commands that fail semantic analysis should throw {@link SemanticCheckException}.
    */
-  @Test
```
penghuo, line 89: why remove this IT? it should fail syntax check, right?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
```diff
@@ -163,6 +168,49 @@ public static void register(BuiltinFunctionRepository repository) {
...
+   */
+  private static Table<ExprValue, ExprValue, ExprValue> notEqualTable =
+      new ImmutableTable.Builder<ExprValue, ExprValue, ExprValue>()
+          .put(LITERAL_NULL, LITERAL_NULL, LITERAL_FALSE)
+          .put(LITERAL_NULL, LITERAL_MISSING, LITERAL_TRUE)
+          .put(LITERAL_MISSING, LITERAL_NULL, LITERAL_TRUE)
+          .put(LITERAL_MISSING, LITERAL_MISSING, LITERAL_FALSE)
+          .build();
+
+  /**
```
penghuo, line 187: The comments is not very related to the valueComparisonTable defined in here. If there is no special handling for NULL and MISSING, the behavior should follow the generic rule defined in here. https://asterixdb.apache.org/docs/0.9.3/sqlpp/manual.html#Operator_expressions

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
```diff
@@ -210,8 +260,85 @@ private static FunctionResolver equal() {
     );
   }
 
-  private static Map<FunctionSignature, FunctionBuilder> predicateFunction(
+  private static FunctionResolver notEqual() {
```
penghuo, line 263: could we define notEqual as the combination of not(equal)

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprValue.java
```diff
@@ -31,6 +31,11 @@
    */
   ExprType type();
 
+  /**
+   * Compare the Object value of itself to the ExprValue v.
+   */
+  int compareTo(ExprValue v);
```
penghuo, line 37: How does this interface been used?

penghuo:
General suggestion, if the PR related to different issue or subject, we should submit serval small PR accordingly instead of using large PR.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/OperatorIT.java
chloe-zh: To escape % character

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/OperatorIT.java
chloe-zh: Ok sure, will do.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
chloe-zh: yes it will fails syntax since I removed unsupported stats functions from grammar file.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
chloe-zh: Looks good, I will refer to this manual and make corresponding changes, thanks!

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/QueryAnalysisIT.java
chloe-zh: Created another PR to add syntax check: https://github.com/opendistro-for-elasticsearch/sql/pull/523

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
chloe-zh: I was also thinking about it didn't figure it out yet

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprValue.java
chloe-zh: To compare the values of instances of the ExprValue subclasses exactly like basic java objects, for example, the comparison between`ExprIntegerValue` instances like 
```
ExprIntegerValue v1 = new ExprIntegerValue(1);
ExprIntegerValue v2 = new ExprIntegerValue(2);
```
Then the comparison between v1 and v2 has the same behavior of the comparison between integers 1 and 2, aka:
```
v1.compareTo(v2) -> -1
v2.compareTo(v1) -> 1
v1.compareTo(v1) -> 0
```

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/StatsCommandIT.java
chloe-zh: To my understanding, we can use verifyColumn like matchers in most of time, but some cases that some other content other than columns and datarows are also important to veriy, so we can flexibly choose text comparison. For example, to verify the formats, or verify the explain results etc. 
As for the case this comment referred to, I would like to verify schema, column names, result set of the aggregation, so I chose to compare text.

Issue #138, opened 2019-08-01
Move the index mapping definitions to standalone json files


arsen-es:
All mappings are currently defined in `src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java` as java strings, which makes it difficult to find/maintain/properly indent when updating/verify validity of json. We need to create one small json per index mapping and load it at run-time.

pakio:
@arsen-es Hi! I tried to fix this issue. if you haven't picked up yet, i'd be grateful if you would check my PR https://github.com/opendistro-for-elasticsearch/sql/pull/526

anirudha:
We Will prioritize it next week , thanks!

dai-chen:
@pakio Thanks for your fix! Resolving the issue.

PR #526, opened 2020-06-18
Move the index mapping definitions to standalone json files

pakio:
*Issue #, if available:* #138 

*Description of changes:*
Move the test index mapping definition, which directly written in code, to standalone json files.

Name of these json files are made referring to the function name.
I referred [this](https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/test/java/com/amazon/opendistroforelasticsearch/sql/esdomain/mapping/FieldMappingsTest.java#L45) code to load each json file placed in `resources/mappings/**'


src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
```diff
@@ -127,566 +130,196 @@ public static Response performRequest(RestClient client, Request request) {
...
-        return "{  \"mappings\": {" +
-                " \"properties\": {\n" +
-                "          \"phrase\": {\n" +
-                "            \"type\": \"text\",\n" +
-                "            \"store\": true\n" +
-                "          }" +
-                "       }"+
-                "   }" +
-                "}";
+        String mappingFile = "mappings/phrase_index_mapping.json";
```
penghuo, line 143: all the method have almost same implementation, the only different is mappingFile name, could you simply it?

penghuo:
Thanks for the change.
Could you sumbit the PR against the develop branch? We plan to make master stable only for release purpose. All the new feature change will push to develop branch.

pakio:
Sure, But I found a lot of differences between master branch and develop branch, and there are three different `TestUtils` (but same code) at develop branch.
Should I apply same changes to all file?

src/test/java/com/amazon/opendistroforelasticsearch/sql/esintgtest/TestUtils.java
pakio: Thank you for the review! I made a function to make logic and error handling common.
Also found `fileToString` method in TestUtils, so i stopped using Resources and used that method.

chloe-zh:
> Sure, But I found a lot of differences between master branch and develop branch, and there are three different `TestUtils` (but same code) at develop branch.
> Should I apply same changes to all file?

Thanks for the changes! There should only one `TestUtils` in dev branch, it's located in `integ-test` module and in the directory `integ-test/src/test/../util`.

chloe-zh:
> Sure, But I found a lot of differences between master branch and develop branch, and there are three different `TestUtils` (but same code) at develop branch.
> Should I apply same changes to all file?

Thanks for the changes! We are migrating the SQL codebase to a new architecture so there are still some pending changes, sorry to make you confused. We will clean up redundant codelines/files in the future.

For the index mapping methods, only `TestUtils` in integ-modul under `src/test/java/../legacy/` is called so it's fine to apply the changes only to this file.

But it looks like you are making changes in the one under `src/test/java/../util/` for now so it's actually skip the mapping methods you added. Could you move the changes toward `integ-test/src/test/java/../legacy/TestUtils` please?

pakio:
Thanks for the detail information! Now I understood what should I do!
I've moved every my changes which was applied to `legacy/..` and index definition files into under `integ
-test/src/test/..`.
Could you please check if my change is alright or need more change?

pakio:
@dai-chen Thanks for reviewing! I've already changed code at TestUtils.java to use these new json files. Is that enough change? or if it's not, I can make more change as needed !

dai-chen:
> @dai-chen Thanks for reviewing! I've already changed code at TestUtils.java to use these new json files. Is that enough change? or if it's not, I can make more change as needed !

I may miss something but I only see the `TestUtils` class in legacy module is removed?

chloe-zh:
> > @dai-chen Thanks for reviewing! I've already changed code at TestUtils.java to use these new json files. Is that enough change? or if it's not, I can make more change as needed !
> 
> 
> 
> I may miss something but I only see the `TestUtils` class in legacy module is removed?

Hi Chen @dai-chen , there are too many code lines removed in `TestUtils` so it doesn’t show up by default. I guess that’s what confused you?

dai-chen:
> > > @dai-chen Thanks for reviewing! I've already changed code at TestUtils.java to use these new json files. Is that enough change? or if it's not, I can make more change as needed !
> > 
> > 
> > I may miss something but I only see the `TestUtils` class in legacy module is removed?
> 
> Hi Chen @dai-chen , there are too many code lines removed in `TestUtils` so it doesn’t show up by default. I guess that’s what confused you?

Yeah, I missed that and thought the whole file was removed :) Thanks!

PR #534, opened 2020-06-26
Support LIKE operator

chloe-zh:
*Issue #, if available:*

*Description of changes:*

**Syntax:** `string like pattern`
The pattern supports string literal with percent(`%`) character for wildcard and underscore(`_`) character for single character. Here are some examples:
`foobar like 'foobar%'` -> true
`foobar like '%bar'` -> true
`foobar like '%BAR'` -> false
`foobar like 'fooba_'` -> true
`foobar like 'foobar_'` -> false
`foobar like '%ba_'` -> true

For null values and missing values in LIKE operation, the results follows [AsterixDB standard](https://asterixdb.apache.org/docs/0.9.3/sqlpp/manual.html#Comparison_operators):
| A | B | A like B |
|---|---|---|
| NULL | NULL | NULL |
| NULL | MISSING | MISSING |
| MISSING | NULL | MISSING |
| MISSING | MISSING | MISSING |

Reference: [**CrateDB**](https://github.com/crate/crate)


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/OperatorUtils.java
```diff
@@ -171,4 +171,69 @@ public String toString() {
...
+
+  /**
+   * Wildcard pattern matcher util.
+   * Percent (%) character for wildcard,
+   * Underscore (_) character for a single character match.
+   * @param pattern string pattern to match.
+   * @return if text matches pattern returns true; else return false.
+   */
+  public static boolean matches(String text, String pattern) {
+    return Pattern.compile(patternToRegex(pattern)).matcher(text).matches();
```
dai-chen, line 183: I checked a few methods in JDK, such as `Pattern.matches`. I think it's more intuitive to put pattern as first argument for this util method as well as like operator argument?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
```diff
@@ -475,4 +513,34 @@ public String toString() {
       }
     };
   }
+
+  private static <T, R> FunctionBuilder likePattern(FunctionName functionName,
+                                                    BiFunction<T, T, R> function,
+                                                    Function<ExprValue, T> observer,
+                                                    ExprType returnType) {
```
dai-chen, line 520: Just did a quick look and found this method is very similar to others, ex. `compareValue`, `equalPredicate` and `binaryPredicate`. The `equalPredicate` method signature looks most generic. I'm curious is there any chance to only keep or reuse a most generic function generator method.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/OperatorUtils.java
chloe-zh: Good point! Will change the argument order in the util method to follow the JDK custom. As for the like operator, I think it would be better to keep consistent with the PPL specification, which is text like pattern.

dai-chen: I see. I wan't aware of that. Thanks!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
chloe-zh: Yes that's also I'm considering, I will take a look into it

chloe-zh:
> Besides IT, could you add a doctest for this?

Thanks @dai-chen ! The doc of operators have not been created/updated, can I add the doc test when adding the experimental doc in another PR?

dai-chen:
> > Besides IT, could you add a doctest for this?
> 
> Thanks @dai-chen ! The doc of operators have not been created/updated, can I add the doc test when adding the experimental doc in another PR?

Sure, no problem! Next time probably we can start from doc to drive the coding and save the trouble of adding all docs at last. :)

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java
chloe-zh: Used generic function bundles generating method `binaryOperator` in `OperatorUtils`, but some for special cases are remained and added java doc for them.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/OperatorUtils.java
```diff
@@ -171,4 +171,69 @@ public String toString() {
...
+   * @param pattern string pattern to match.
+   * @return if text matches pattern returns true; else return false.
+   */
+  public static boolean matches(String pattern, String text) {
+    return Pattern.compile(patternToRegex(pattern)).matcher(text).matches();
+  }
+
+  private static final char DEFAULT_ESCAPE = '\\';
+
+  private static String patternToRegex(String patternString) {
```
penghuo, line 188: 1. could you seperate it from OperatorUtils in case is more focus on SQL regex to Java regex.
2. by define the ESCAPE is defined in SQL language, how do we support it?
3. is it possible to seperate the translation rule from translation logic.
for example, 
TranslationRule.Buidler.put("//%", %").put("%", ".*").build()
Then, apply rule on the input REGEX.

PR #540, opened 2020-06-29
Support mathematical functions ceil/ceiling, exp, floor, ln, log

chloe-zh:
*Issue #, if available:*

*Description of changes:*

Implementation of function expressions is referred to MySQL mathematical functions as the requirement and standard.
Mathematical functions in core will be separated in a couple of pull requests. Functions in this pull request include:

- `ceil/ceiling` 
**usage:** maps a number to the least integer greater than or equal to itself
**argument type:** integer/long/float/double
**return type:** integer

- `exp`
**usage:** maps a number to its natural exponential function
**argument type:** integer/long/float/double
**return type:** double

- `floor`
**usage:** maps a number to the greatest integer less than or equal to itself
**argument type:** integer/long/float/double
**return type:** integer

- `ln`
**usage:** maps a number to its natural logarithm
**argument type:** integer/long/float/double
**return type:** double

- `log`  
**usage:** log(b, x) maps numbers b, x to their logarithm where b is the base, log(x) maps x to the natural logarithm, which is equivalent to ln(x)
**argument type:** integer/long/float/double
**return type:** double

- `log10`  
**usage:** maps a number to its 10-base logarithm
**argument type:** integer/long/float/double
**return type:** double

- `log2`  
**usage:** maps a number to its 2-base logarithm
**argument type:** integer/long/float/double
**return type:** double

**Reference**
- [MySQL Reference Munual](https://dev.mysql.com/doc/refman/8.0/en/)
- [ArterixDB Numeric Functions](https://asterixdb.apache.org/docs/0.9.3/aql/builtins.html#NumericFunctions)


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/OperatorUtils.java
```diff
@@ -104,6 +104,57 @@ public String toString() {
...
+   * @param function     {@link BiFunction}
+   * @param observer1    extract the value of type T from the first argument
+   * @param observer2    extract the value of type U from the second argument
+   * @param returnType   return type
+   * @param <T>          the type of the first argument to the function
+   * @param <U>          the type of the second argument to the function
+   * @param <R>          the type of the result of the function
+   * @return {@link FunctionBuilder}
+   */
+  public static <T, U, R> FunctionBuilder doubleArgFunc(
```
dai-chen, line 123: What's the difference between this new method and `binaryOperator()` above?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
```diff
@@ -0,0 +1,214 @@
...
+    FunctionName functionName = BuiltinFunctionName.CEILING.getName();
+    return new FunctionResolver(
+        BuiltinFunctionName.CEILING.getName(),
+        new ImmutableMap.Builder<FunctionSignature, FunctionBuilder>()
+            .put(
+                new FunctionSignature(functionName, Arrays.asList(ExprType.DOUBLE)),
+                unaryOperator(
+                    functionName, v -> ((long) Math.ceil(v)), ExprValueUtils::getDoubleValue,
+                    ExprType.LONG))
+            .build());
```
dai-chen, line 97: Exact same as `ceil()`?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/ppl/MathematicalFunctionIT.java
```diff
@@ -0,0 +1,109 @@
...
+  @Test
+  public void testLn() throws IOException {
+    JSONObject result =
+        executeQuery(
+            String.format(
+                "source=%s | eval f = ln(age) | fields f", TEST_INDEX_BANK));
+    verifySchema(result, schema("f", null, "double"));
+    verifyDataRows(
+        result, rows(Math.log(32)), rows(Math.log(36)), rows(Math.log(28)), rows(Math.log(33)),
+        rows(Math.log(36)), rows(Math.log(39)), rows(Math.log(34)));
```
dai-chen, line 93: For assertion on floating-point number, please check if we're using matcher like `isCloseTo()` to avoid precision problem.

dai-chen:
Recently I found the function name has to be lowercase otherwise not supported function exception is thrown. Could you check if we can do a quick fix here or a separate PR is needed?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/OperatorUtils.java
chloe-zh: only the `toString` method is different

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
chloe-zh: Yes the same method but with different registered name

dai-chen: I see. I think we can refactor these kind of similar function family later by generating from same method with different argument, such as substr and substring, log and log2 etc.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/OperatorUtils.java
dai-chen: Okay. For now `toString` seems being used for column name in response, I think we can treat arithmetic and function toString same way later when column name is retrieved in other way.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/DSL.java
```diff
@@ -58,6 +58,38 @@ public FunctionExpression abs(Environment<Expression, ExprType> env, Expression.
         repository.compile(BuiltinFunctionName.ABS.getName(), Arrays.asList(expressions), env);
   }
 
+  public FunctionExpression ceil(Environment<Expression, ExprType> env, Expression... expressions) {
+    return (FunctionExpression)
+        repository.compile(BuiltinFunctionName.CEIL.getName(), Arrays.asList(expressions), env);
```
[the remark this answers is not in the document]
chloe-zh: Done

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/util/MatcherUtils.java
```diff
@@ -239,8 +233,29 @@ protected boolean matchesSafely(JSONArray array) {
...
+    return new TypeSafeMatcher<JSONArray>() {
+      @Override
+      protected boolean matchesSafely(JSONArray item) {
+        List<Number> expectedValues = new ArrayList<>(Arrays.asList(values));
+        List<Number> actualValues = new ArrayList<>();
+        item.iterator().forEachRemaining(v -> actualValues.add((Number) v));
+        return actualValues.stream()
+            .map(v -> (valuesAreClose(v, expectedValues.get(actualValues.indexOf(v)))))
+            .reduce((a, b) -> a && b)
+            .get();
```
dai-chen, line 247: np: `allMatch(v -> valuesAreClose...)` ?

chloe-zh: Done, thanks!

Issue #227, opened 2019-10-18
max()/min() not supported for String and Date types


jordanw-bq:
Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT max(Origin) AS usr_Calculation_462181953493630977_nk FROM kibana_sample_data_flights

**Result:**
Raw response received: {
  "error": {
    "reason": "Invalid SQL query",
    "details": "Function [MAX] cannot work with [KEYWORD]. Usage: MAX(NUMBER T) -> T",
    "type": "SemanticAnalysisException"
  },
  "status": 400
}

**Expectation:**
Tableau, using other SQL connectors, allows users to use these functions on these types. Adding support for this in Elasticsearch will allow for full support for these functions when used through Tableau.

dai-chen:
This exception is thrown from the type checker in semantic analysis. Will disable it to see if we do have the support for `max/min` on String and Date. If so, that means the type check is too strict and needs to relax for this case. Otherwise we need to add the support before relaxing the restriction.

dai-chen:
Just disabled semantic analysis but found ES max aggregation can only work with numeric value. Need to add support for this and add overloaded function spec to type checker in new parser.

Explain for `MAX` with string:

```
{
  "from" : 0,
  "size" : 0,
  "_source" : {
    "includes" : [
      "MAX"
    ],
    "excludes" : [ ]
  },
  "aggregations" : {
    "MAX(firstname.keyword)" : {
      "max" : {
        "field" : "firstname.keyword"
      }
    }
  }
}
```

Exception thrown from ES engine:

```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "Expected numeric type on field [firstname.keyword], but got [keyword]"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "accounts",
        "node": "jn9G8trERRikgYiF3_eVng",
        "reason": {
          "type": "illegal_argument_exception",
          "reason": "Expected numeric type on field [firstname.keyword], but got [keyword]"
        }
      }
    ],
    "caused_by": {
      "type": "illegal_argument_exception",
[...]

PR #541, opened 2020-06-30
Support aggregations min, max

chloe-zh:
*Issue #, if available:*

*Description of changes:*

* Added max and min aggregations and corresponding tests


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
```diff
@@ -0,0 +1,106 @@
...
+          break;
+        default:
+          throw new ExpressionEvaluationException(
+              String.format("unexpected type [%s] in min aggregation", type));
+      }
+    }
+
+    @Override
+    public ExprValue result() {
+      return isNullResult ? ExprNullValue.of() : minResult;
```
dai-chen, line 103: So if any field is `NULL`, then null is returned as result? Are we following some spec here? I checked MySQL and ANSI SQL, it seems "All aggregate functions except COUNT(*) will ignore NULL values when computing their results." Unless all fields are null, result won't be null eventually.

For example, the result is 2 for the following test data and query.
```
CREATE TABLE Test(age INT);

INSERT INTO Test VALUES(1);
INSERT INTO Test VALUES(null);
INSERT INTO Test VALUES(2);

SELECT MAX(age) FROM Test
```

penghuo: +1, we should follow https://asterixdb.apache.org/docs/0.9.3/sqlpp/manual.html#Aggregation_functions ARRAY_XXXX
Our previous implementation of SUM, AVG, COUNT also need to be changed.

dai-chen:
Could you also check if old issues like this can be fixed here? https://github.com/opendistro-for-elasticsearch/sql/issues/227

chloe-zh:
> Could you also check if old issues like #227 can be fixed here?

Yes, added new signature to support MIN and MAX with string. Thanks!

dai-chen:
> > Could you also check if old issues like #227 can be fixed here?
> 
> Yes, added new signature to support MIN and MAX with string. Thanks!

Awesome! Let's resolve old issues if fixed in new engine. Thanks.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
```diff
@@ -63,19 +62,19 @@ public String toString() {
...
-    public AvgState() {
+    AvgState() {
       this.count = 0;
       this.total = 0d;
     }
 
     @Override
     public ExprValue result() {
-      return isNullResult ? ExprNullValue.of() : ExprValueUtils.doubleValue(total / count);
+      return isEmptyCollection ? ExprNullValue.of() : ExprValueUtils.doubleValue(total / count);
```
dai-chen, line 77: Can this `isEmptyCollection` be replaced by `count == 0`? Because I notice `count` already represents the count of non-null/missing value only. Same question for all other aggregators.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
```diff
@@ -0,0 +1,111 @@
...
+import com.amazon.opendistroforelasticsearch.sql.expression.Expression;
+import com.amazon.opendistroforelasticsearch.sql.expression.function.BuiltinFunctionName;
+import com.amazon.opendistroforelasticsearch.sql.storage.bindingtuple.BindingTuple;
+import java.util.List;
+
+/**
+ * The minimum aggregator aggregate the value evaluated by the expression.
+ * If the expression evaluated result is NULL or MISSING, then the result is NULL.
+ */
+public class MinAggregator extends Aggregator<MinAggregator.MinState> {
```
dai-chen, line 43: Just a thought: if we see more aggregate functions like MIN and MAX later (ex. window functions), probably we can consider using a single class for generalization. Because I notice only several lines different between `MinAggregator` and `MaxAggregator`.

coderReview:
hello @dai-chen and @chloe-zh does this also allows min/max for Date fields? I think DSL supports that already.

dai-chen:
> hello @dai-chen and @chloe-zh does this PR also allow min/max for Date fields? I think DSL supports that already.

Thanks for your comment! I think only min/max for string is covered additionally in this PR. FYI, in this new query engine, we haven't pushed down the aggregation to Elasticsearch DSL yet. The aggregation is implemented and happening in memory for now. @chloe-zh is it possible to support date too?

coderReview:
Don't need to that in memory. Here https://discuss.elastic.co/t/min-date-and-max-date/125174

dai-chen:
> Don't need to that in memory. Here https://discuss.elastic.co/t/min-date-and-max-date/125174

Thanks for the info! Yeah, just to clarify, the post-processing in memory capability is to make sure our new engine can handle queries that unable to push down to ES DSL simply. For example, in future, we may support min/max aggregation on result of another subquery.

I think it's good to consider date support at this moment because our current aggregate operator is using JDK `Math.min/max` which assumes numeric input. Actually it could be generalized to support all data type that implements `Comparable` interface.

chloe-zh:
Hi @coderReview  @dai-chen , this pr supports min/max aggregation date and time types now

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
chloe-zh: made this change in avg/count/sum, but for min and max, we don't have to count the number so still using `isEmptyCollection`.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
```diff
@@ -0,0 +1,100 @@
...
+    private ExprValue minResult;
+    private boolean isEmptyCollection;
+
+    MinState(ExprCoreType type) {
+      minResult = isNumber(type) ? doubleValue(Double.MAX_VALUE) : LITERAL_NULL;
+      isEmptyCollection = true;
+    }
+
+    public void min(ExprValue value) {
+      minResult = minResult.isNull() ? value : minResult.compareTo(value) < 0 ? minResult : value;
```
dai-chen, line 87: np: could you add parentheses or split it into multi-line?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
```diff
@@ -0,0 +1,100 @@
...
+  public String toString() {
+    return String.format("min(%s)", format(getArguments()));
+  }
+
+  protected static class MinState implements AggregationState {
+    private ExprValue minResult;
+    private boolean isEmptyCollection;
+
+    MinState(ExprCoreType type) {
+      minResult = isNumber(type) ? doubleValue(Double.MAX_VALUE) : LITERAL_NULL;
```
dai-chen, line 82: Trying to understand why we need different initial values for `minResult`. Can we just use literal null?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
```diff
@@ -0,0 +1,100 @@
...
+      isEmptyCollection = true;
+    }
+
+    public void min(ExprValue value) {
+      minResult = minResult.isNull() ? value : minResult.compareTo(value) < 0 ? minResult : value;
+    }
+
+    @Override
+    public ExprValue result() {
+      return isEmptyCollection ? ExprNullValue.of() : minResult;
```
dai-chen, line 92: So `isEmptyCollection` is true only when `minResult` is not updated initially right? Could we remove this flag and just return `minResult`?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
chloe-zh: Yes sounds good

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
chloe-zh: Done, thanks!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/MinAggregator.java
chloe-zh: Done, thanks!

dai-chen (approved):
LGTM. Could you also add some comparison test from SQL side?

chloe-zh:
> LGTM. Could you also add some comparison test from SQL side?

Done, thanks!

penghuo:
Please update the docs.
https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/experiment/ppl/cmd/stats.rst

chloe-zh:
> Please update the docs.
> https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/experiment/ppl/cmd/stats.rst

Done, thanks!

penghuo (approved):
Thanks for the change.

PR #550, opened 2020-07-08
Redefine the Expression interface and Type

penghuo:
*Issue #, if available:*

*Description of changes:*
## 1.Expression

In current design, the Expression has two observers, valueOf and type.
```
public interface Expression {

  /**
   * Evaluate the value of expression in the value environment.
   */
  ExprValue valueOf(Environment<Expression, ExprValue> valueEnv);

  /**
   * Evaluate the type of expression in the type environment.
   */
  ExprType type(Environment<Expression, ExprType> typeEnv);
}
```
The Expression is construct during the analyzer stage with the type environment provided by the storage engine. so the type observer should not rely on the type environment during the execution stage. 
The interface of Expression should be changed as follows
```
public interface Expression {

  /**
   * Evaluate the value of expression in the value environment.
   */
  ExprValue valueOf(Environment<Expression, ExprValue> valueEnv);

  /**
   * Retrieve the type of expression.
   */
  ExprType type();
}
```
## 2.Expression Type (ExprType)

The ExprType define the Type of expression in the QueryEngine. In current design, ExprType is enum which can't been extended. 
In this PR, we define the ExprType as the interface and define the ExprCoreType which extends the ExprType.


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/type/WideningTypeRule.java
```diff
@@ -0,0 +1,87 @@
...
+ * STRUCT       STRUCT
+ */
+@UtilityClass
+public class WideningTypeRule {
+  public static final int IMPOSSIBLE_WIDENING = Integer.MAX_VALUE;
+  public static final int TYPE_EQUAL = 0;
+
+  /**
+   * The widening distance is calculated from the leaf to root.
+   * e.g. distance(INTEGER, FLOAT) = 2, but distance(FLOAT, INTEGER) =
```
dai-chen, line 44: Unfinished comment "distance(FLOAT, INTEGER) =" ?

penghuo: done.

PR #560, opened 2020-07-10
add date and time support

penghuo:
## Description of changes
1. Define DATE, TIME, TIMESTAMP as the core type of query engine.
2. Add the date, time and timestamp Literal.
3. Add the ```INTEGER dayofmonth(DATE)``` function.
4. Add the doc. https://github.com/penghuo/sql/blob/datetype/docs/user/dql/expressions.rst

## To Reviewer
1. Missing the correct test, because the JDBC doesn't support TIMESTAMP and TIME type, and also has the wrong mapping between DATE and TIMESTMAP.

### DIFF
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/ExpressionAnalyzer.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/ExpressionAnalyzer.java
@@ -78,7 +78,8 @@ public class ExpressionAnalyzer extends AbstractNodeVisitor<Expression, Analysis
 
   @Override
   public Expression visitLiteral(Literal node, AnalysisContext context) {
-    return DSL.literal(ExprValueUtils.fromObjectValue(node.getValue()));
+    return DSL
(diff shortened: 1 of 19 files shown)


penghuo:
> Would it be better if the value() of DATE, TIME, TIMESTAMP are made to return LocalDate, LocalTime, Instant types rather than String type? Since they should work in some arithmetic operations like calculating the date/time difference fo example DATE ‘2020-07-09’ - DATE ‘2020-07-08’, and should also work in adding and subtracting INTERVAL like https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_timestampadd

The value() interface has two purpose now,
(1) Used to get the original Java Data and then perform the calculation.
(2) Used by Protocol to get the value for result.
For primitive data type like integer, string, it works fine. But for other types, like date, the format should be controlled by the value itself.
Currently, for ExprDateValue, the value() interface is used by the Protocol to get the serialized result, and I add getDate() method in ExprDateValue performing calculation.
I am also thinking deprecated the value interface and add integer(), long(), and date(), time() interface instead.

PR #562, opened 2020-07-10
Add Cypress testing for SQL Workbench

joshuali925:
*Issue #, if available:*

*Description of changes:*
- Add cypress for Workbench sanity test, add test cases


chloe-zh:
Could you present an example of what the generated cypress test looks like?

Issue #565, opened 2020-07-10
ODBC: Disable catalog support through driver


jordanw-bq:
Catalog support will need to be removed in order to keep applications like Excel from generating queries containing catalog qualifiers.

PR #566, opened 2020-07-10
ODBC: Remove catalog support from driver

jordanw-bq:
*Issue #, if available:* #565 

*Description of changes:*
- return error when catalogs/schemas are queried
- test & info call fixes


sql-odbc/src/odfesqlodbc/es_info.cpp
```diff
@@ -729,6 +734,15 @@ ESAPI_Tables(HSTMT hstmt, const SQLCHAR *catalog_name_sql,
         std::string query;
         GenerateTableQuery(query, flag, table_name, result_type, table_valid);
 
+        // Get list of columns with SELECT * query since columns doesn't match
```
rupal-bq, line 737: IIRC, this won't be applicable here since for SQLTables is using SHOW TABLES query which doesn't have column information.

rupal-bq: And list_of_columns isn't used in this function afterwards

jordanw-bq: Misplaced changes; will remove

jordanw-bq: Removed

Issue #331, opened 2019-12-30
JSON does not allow non-finite numbers


dai-chen:
An error "JSON does not allow non-finite numbers" thrown when using math function. This may be caused by float number with too many digits after decimal point. No error stack found in ES log.

```
POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT acos(FlightDelayMin) FROM kibana_sample_data_flights"
}

POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT asin(FlightDelayMin) FROM kibana_sample_data_flights"
}

POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT log(FlightDelayMin) FROM kibana_sample_data_flights"
}

POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT (log(FlightDelayMin)/log(10)) FROM kibana_sample_data_flights"
}
```

dai-chen:
The root cause is trig functions only accepts [-1, 1]. For other cases, the result is NaN which causes the exception above in org.json library.

```
[2020-04-24T17:11:55,160][ERROR][c.a.o.s.e.AsyncRestExecutor] [186590df9563.ant.amazon.com] Error
org.json.JSONException: JSON does not allow non-finite numbers.
    at org.json.JSONObject.testValidity(JSONObject.java:2214) ~[json-20180813.jar:?]
    at org.json.JSONArray.put(JSONArray.java:1050) ~[json-20180813.jar:?]
    at com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.dataEntry(Protocol.java:255) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
    at com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.getDataRowsAsJson(Protocol.java:245) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
    at com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.outputInJdbcFormat(Protocol.java:159) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
    at com.amazon.opendistroforelasticsearch.sql.executor.format.Protocol.format(Protocol.java:135) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
[...]

PR #577, opened 2020-07-15
Support mathematical functions: conv, crc32, mod, pow/power, round, sign, sqrt, truncate

chloe-zh:
*Issue #, if available:*

*Description of changes:*

- `conv` 
**usage:** conv(x, a, b) converts the number x from a base to b base
**argument type:** x: string, a: integer, b: integer
**return type:** string

- `crc32`
**usage:** calculates a cyclic redundancy check value and returns a 32-bit unsigned value
**argument type:** string
**return type:** long

- `mod`
**usage:** mod(n, m) calculates the remainder of the number n divided by m
**argument type:** integer/long/float/double
**return type:** widen type between types of n and m

- `pow/power`
**usage:** pow(x, y) calculates the value of x raised to the power of y
**argument type:** integer/long/float/double
**return type:** double

- `round`
**usage:** round(x, d) rounds the argument x to d decimal places, d defaults to 0 if not specified
**argument type:** integer/long/float/double
**return type map:**
(integer/long [,integer]) -> long
(float/double [,integer]) -> double

- `sign`
**usage:** returns the sign of the argument as -1, 0, or 1, depending on whether the number is negative, zero, or positive
**argument type:** integer/long/float/double
**return type:** integer

- `sqrt`
**usage:** calculate the square root of a nonnegative number
**argument type:** integer/long/float/double
**return type map:**
(nonnegative) integer/long/float/double -> double
(negative) integer/long/float/double -> null

- `truncate`
**usage:** truncate(x, d) returns the number x, truncated to d decimal places
**argument type:** integer/long/float/double
**return type map:**
integer/long -> long
float/double  -> double

**Reference**
- [MySQL Reference Munual](https://dev.mysql.com/doc/refman/8.0/en/)
- [ArterixDB Numeric Functions](https://asterixdb.apache.org/docs/0.9.3/aql/builtins.html#NumericFunctions)


dai-chen:
Could you check if we can avoid and fix this issue in new engine? https://github.com/opendistro-for-elasticsearch/sql/issues/331

dai-chen:
I think the description of this PR should be added to our doc. Could you add it and make this a doctest file: https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/dql/functions.rst

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/MathematicalFunctionIT.java
```diff
@@ -0,0 +1,143 @@
...
+    result = executeQuery("select round(-56, -1)");
+    verifySchema(result, schema("round(-56, -1)", null, "long"));
+    verifyDataRows(result, rows(-60));
+  }
+
+  /**
+   * Test sign function with double value.
+   */
+  @Test
+  public void testSign() throws IOException {
```
dai-chen, line 103: Is this case and all others in this IT unable to cover in comparison test?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
```diff
@@ -97,6 +110,58 @@ private static FunctionResolver ceiling() {
...
+  private static FunctionResolver conv() {
+    FunctionName functionName = BuiltinFunctionName.CONV.getName();
+    return new FunctionResolver(
+        functionName,
+        new ImmutableMap.Builder<FunctionSignature, FunctionBuilder>()
+            .put(
+                new FunctionSignature(functionName,
+                    Arrays.asList(ExprCoreType.STRING, ExprCoreType.INTEGER, ExprCoreType.INTEGER)),
+                tripleArgFunc(
+                    functionName, (t, u, v) -> Integer.toString(Integer.parseInt(t, u), v),
```
dai-chen, line 128: Could we use more readable argument name such as `(num, fromBase, toBase) ->` ?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
```diff
@@ -182,6 +247,186 @@ private static FunctionResolver log2() {
...
+  /**
+   * Definition of mod(x, y) function.
+   * Calculate the remainder of x divided by y
+   * The supported signature of mod function is
+   * (INTEGER, INTEGER) -> INTEGER
+   * (LONG, LONG) -> LONG
+   * (FLOAT, FLOAT) -> FLOAT
+   * (DOUBLE, DOUBLE) -> DOUBLE
+   */
+  private static FunctionResolver mod() {
```
penghuo, line 259: does mod support LONG and FLOAT?

dai-chen: MySQL seems support float and long? https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_mod

@chloe-zh Btw, does Java native '%' cover this case in MySQL doc above? "MOD(N,0) returns NULL."

integ-test/src/test/resources/correctness/expressions/functions.txt
```diff
@@ -10,6 +10,19 @@ ceil(-1)
...
+power(2, -2.1)
+power(abs(2), 2)
+sign(0)
+sign(-1)
+sign(1)
+sign(abs(1))
+sqrt(0)
+sqrt(1)
+sqrt(1.1)
+sqrt(abs(1))
```
dai-chen, line 25: Any reason why not test `mod` here? I can see the support in H2: http://www.h2database.com/html/functions.html#mod

chloe-zh: currently the schema does not match with H2

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
chloe-zh: sounds good, will do

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/MathematicalFunctionIT.java
chloe-zh: the `sign` function in comparison test is covered by H2, but the schema doesn't match for now

integ-test/src/test/resources/correctness/expressions/functions.txt
dai-chen: Is name or type different?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
chloe-zh: LONG type should be good as says in MySQL manual (as BIGINT), but it doesn't mention about FLOAT.

ppl/src/main/antlr/OpenDistroPPLLexer.g4
```diff
@@ -137,20 +137,29 @@ DC:                                 'DC';
...
+TRUNCATE:                           'TRUNCATE';
 
 // LITERALS AND VALUES
 //STRING_LITERAL:                     DQUOTA_STRING | SQUOTA_STRING | BQUOTA_STRING;
 ID:                                 ID_LITERAL;
 INTEGER_LITERAL:                    DEC_DIGIT+;
 DECIMAL_LITERAL:                    (DEC_DIGIT+)? '.' DEC_DIGIT+;
 
-fragment ID_LITERAL:                [A-Z_$0-9@]*?[A-Z_$\-]+?[A-Z_$\-0-9]*;
+fragment ID_LITERAL:                [A-Z_]+[A-Z_$0-9@\-]*;
```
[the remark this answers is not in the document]
chloe-zh: The original `ID_LITERAL` regex allows digital numbers and special char `@`, `\-` at the starting place. The negative numbers are also recognized as IDs when I tested out some cases in functions. So I changes the rule to only allow letters and underscore(`_`) as the start character.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/arthmetic/MathematicalFunction.java
chloe-zh: @dai-chen Good point! yes currently it also throws arithmetic exception. Will change its behavior to return NULL instead.

integ-test/src/test/resources/correctness/expressions/functions.txt
chloe-zh: @dai-chen It's name. Currently all tests of functions have different name in schema with H2. For example the query `SELECT SIGN(1.1)` has a column name of `SIGN(1.1)` while H2 has the result as the column name, which looks like:
```
"schema": [{
    "name": "1",
    "type": "INTEGER"
}]
```

There are still a couple of functions that have different behavior compared to SQLite, while has the same behavior with H2, like the `sign` function above but schema name is different. These cases are also added to manual test rather than the comparison test cases for now.

dai-chen: I see. Sorry I forgot that! We should be able to overcome this once we support column alias.

ppl/src/main/antlr/OpenDistroPPLLexer.g4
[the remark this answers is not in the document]
dai-chen: Sure, I will check this issue too. Because I'm changing this to define what is identifier, will see if digit should be not allowed as prefix. You can keep this change in your PR.

chloe-zh:
> Could you check if we can avoid and fix this issue in new engine? #331

MySQL returns `null` for functions with illegal arguments, like `ln(-1)`, `log(-1)` etc, and for the result of infinity, and we keep consistent with this behavior. But for special cases like `pow(0, 0)` (Java behavior is returning 1), I didn't find any reference in MySQL manual, so for now it still follows the java behaviors.

dai-chen:
> > Could you check if we can avoid and fix this issue in new engine? #331
> 
> MySQL returns `null` for functions with illegal arguments, like `ln(-1)`, `log(-1)` etc, and for the result of infinity, and we keep consistent with this behavior. But for special cases like `pow(0, 0)` (Java behavior is returning 1), I didn't find any reference in MySQL manual, so for now it still follows the java behaviors.

Sounds good! I just tested `pow(0,0)` in MySQL. It returns 1 same as Java. :)

docs/user/dql/functions.rst
```diff
@@ -139,6 +139,20 @@ Description
...
+====
+
+Description
+-----------
+
+Usage: CONV(x, a, b) converts the number x from a base to b base
+
+Argument type: x: string, a: integer, b: integer
+
+Return type: string
```
dai-chen, line 153: Could you add example for each function? Reference: https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_mod

For example,

```
    od> SELECT MOD(5, 2), MOD(-5, 2), MOD(5, 0);
    fetched rows / total rows = 1/1
    +---------+-------------+-------------+
    ...
```

chloe-zh: Sorry I forgot to add examples here. Will do

PR #591, opened 2020-07-17
Support mathematical functions rand and constants e, pi

chloe-zh:
*Issue #, if available:*

*Description of changes:*

- `RAND`
**Usage:** `rand()`/`rand(N)` returns a random floating-point value in the range 0 <= value < 1.0. If integer N is specified, the seed is initialized prior to execution. One implication of this behavior is with identical argument N, `rand(N)` returns the same value each time, and thus produces a repeatable sequence of column values.
**Argument type:** x: STRING, a: INTEGER, b: INTEGER
**Return type:** STRING

- `E`
**Usage:** `e()` returns the Euler's number
**Return type:** DOUBLE

- `PI`
**Usage:** `pi()` returns the constant pi
**Return type:** DOUBLE


sql/src/main/antlr/OpenDistroSQLParser.g4
```diff
@@ -166,16 +166,16 @@ scalarFunctionName
...
+    | RAND | ROUND | SIGN | SQRT | TRUNCATE
     ;
 
 dateTimeFunctionName
     : DAYOFMONTH
     ;
 
 functionArgs
-    : functionArg (COMMA functionArg)*
+    : (functionArg (COMMA functionArg)*)?
```
dai-chen, line 178: is this not necessary because of `functionArgs?` in `functionCall`? And I think it's more appropriate to mark this optional in `functionCall` for PPL too?

docs/user/dql/functions.rst
```diff
@@ -616,10 +636,21 @@ RAND
...
 -----------
 
-Specifications: 
+Usage: RAND()/RAND(N) returns a random floating-point value in the range 0 <= value < 1.0. If integer N is specified, the seed is initialized prior to execution. One implication of this behavior is with identical argument N, rand(N) returns the same value each time, and thus produces a repeatable sequence of column values.
+
+Argument type: x: STRING, a: INTEGER, b: INTEGER
 
-1. RAND() -> NUMBER
-2. RAND(NUMBER T) -> T
+Return type: STRING
```
dai-chen, line 643: return type is float?

docs/user/dql/functions.rst
```diff
@@ -616,10 +636,21 @@ RAND
 Description
 -----------
 
-Specifications: 
+Usage: RAND()/RAND(N) returns a random floating-point value in the range 0 <= value < 1.0. If integer N is specified, the seed is initialized prior to execution. One implication of this behavior is with identical argument N, rand(N) returns the same value each time, and thus produces a repeatable sequence of column values.
+
+Argument type: x: STRING, a: INTEGER, b: INTEGER
```
dai-chen, line 641: what does this mean?

sql/src/main/antlr/OpenDistroSQLParser.g4
chloe-zh: Yes functions with no argument will return an empty collection as `functionArgs` in parser. It has also been refracted in PPL.

docs/user/dql/functions.rst
chloe-zh: Typo, will correct it, thanks!

docs/user/dql/functions.rst
chloe-zh: It should be `INTEGER`

sql/src/main/antlr/OpenDistroSQLParser.g4
dai-chen: I was asking is this change necessary because:

```
functionCall
    : scalarFunctionName LR_BRACKET functionArgs? RR_BRACKET        #scalarFunctionCall
    ;
```

docs/user/dql/functions.rst
```diff
@@ -291,9 +291,19 @@ E
...
 
-1. E() -> DOUBLE
+Return type: DOUBLE
+
+Example::
+
+    od> SELECT E()
+    fetched rows / total rows = 1/1
+    +-------------------+
+    | e()               |
```
dai-chen, line 303: Sorry for last minute comment. Since we haven't added many functions yet, could you check if the column name required to be all upper case (ex. `E()`, `PI()`) or not? In MySQL or SQL standard.

chloe-zh: Good point! Checked it in MySQL, it should be consistent with the input cases

dai-chen: Thanks for confirming. So the result column name is always lowercase for now? We can fix this separately if change is major.

chloe-zh: Yes all lowercase for now. I looked into this issue, it might be a major change. Created related issue: https://github.com/opendistro-for-elasticsearch/sql/issues/606

dai-chen: Cool. Thanks!

Issue #603, opened 2020-07-21
Microsoft Excel on Mac adds a semicolon at the end of query


rupal-bq:
Microsoft Excel creates queries with semicolon which fails on SQL plugin.

![image](https://user-images.githubusercontent.com/56703525/88004245-7e025a80-cabb-11ea-9609-2ff0c6be3388.png)

rupal-bq:
The error returned by SQL plugin:
```
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "Failed to parse query due to offending symbol [;] at: 'SELECT * FROM kibana_sample_data_flights;' <--- HERE... More details: extraneous input ';' expecting <EOF>",
    "type": "SyntaxAnalysisException"
  },
  "status": 400
}
```

PR #609, opened 2020-07-22
Support queries end with semi colon

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/603

*Description of changes:* Changed ANTLR grammar in both old and new engine to support optional semicolon. Removed semicolon before sending to old Druid parser (which is handwritten without grammar file to change).

*Testing*: Added new UT and IT. Based on our new test workflow, each legacy IT runs twice with/without new engine enabled.

*Documentation*: Added a sentence to declare the support for queries end with semicolon: https://github.com/dai-chen/sql/blob/support-semi-colon/docs/user/dql/basics.rst#syntax


dai-chen:
> LGTM, thanks for the changes! BTW, we also need to support semi colon in workbench to adapt this changes since the workbench is recognizing the semi colon as the separators among queries.

Semicolon at the end of a query is optional. Is there anything possibly broken in SQL Workbench in this case?

Issue #595, opened 2020-07-20
ODBC: Create Power BI M connector for connecting to Open Distro For Elasticsearch


rupal-bq:

coderReview:
Hello @rupal-bq what is the roadmap for stable PowerBI connector? Thank you.

dai-chen:
> Hello @rupal-bq what is the roadmap for stable PowerBI connector? Thank you.

Hi @anirudha , do you have the answer for this ?

dai-chen:
Found Ani's comment on this: https://github.com/opendistro-for-elasticsearch/sql-odbc/issues/145#issuecomment-663176871

PR #610, opened 2020-07-23
ODBC: Adding BASIC & AWS_SIGV4 auth in M Connector 

rupal-bq:
*Issue #595*

*Description of changes:*
- Added BASIC auth
- Added AWS_SIGV4 auth
- Added unit test for checking the connection
- Added error handling 


sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
```diff
@@ -1,42 +1,214 @@
...
+                    rows = Table.TransformRows(source, each Diagnostics.LogValue("SQLColumns", _)),
+                    toTable = Table.FromRecords(rows)
+                in
+                    Value.ReplaceType(toTable, Value.Type(source))
+            else
+                source,
+             
+        // Add support for `LIMIT` and `OFFSET` clauses (rather than `TOP`)
+        AstVisitor = [
+            // format is "LIMIT [<skip>,]<take>" - ex. LIMIT 2,10 or LIMIT 10
```
jordanw-bq, line 117: Can you verify whether ODFE SQL supports this `[skip,]take` syntax? If it doesn't, then we may want to change this to `LIMIT <take> OFFSET <skip>`

sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
```diff
@@ -1,42 +1,214 @@
...
-            Port = Port
+            Driver = "ODFE SQL ODBC Driver"
         ],
-        OdbcDatasource = Odbc.DataSource(ConnectionString)
+
+        SqlCapabilities = Diagnostics.LogValue("SqlCapabilities_Options", [
+            GroupByCapabilities = ODBC[SQL_GB][SQL_GB_COLLATE]
+        ]),
+
+        SQLGetInfo = Diagnostics.LogValue("SQLGetInfo_Options", [
```
jordanw-bq, line 85: These are values that we can change at the driver level. Unless those changes break other applications, lets set Capabilities and GetInfo values there instead

Can you remove these lines (SQLCapabilities & SQLGetInfo) and verify if we get the same connection behaviour?

sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
rupal-bq: `[skip,]take` syntax is supported

sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
rupal-bq: Removed SQLCapabilities & SQLGetInfo since all required values are defined at the driver level

sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
```diff
@@ -1,42 +1,203 @@
...
+                            else
+                                Number.ToText(skip) & ","
+                    in
+                        [
+                            Text = Text.Format("LIMIT #{0}#{1}", { skip, take }),
+                            Location = "AfterQuerySpecification"
+                        ]
+        ],
+
+        OdbcDatasource = Odbc.DataSource(ConnectionString & Server & CredentialConnectionString, [
```
jduo, line 126: Shouldn't this concatenate EncryptedConnectionString too?

sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
```diff
@@ -1,42 +1,203 @@
...
+            Label = "BASIC"
+        ],
+        Key = [
+            Label = "AWS_SIGV4",
+            KeyLabel = "Region"
+        ]
     ],
+
+    // Enable Encryption
+    SupportsEncryption = true,
```
jduo, line 199: Note that if you set this to true, PBI will always try encrypted connections. You need to add an error handler that tells PBI to try an unencrypted connection if it gets an SSL error. See the ImpalaODBC example on the PowerQuery SDK github.

rupal-bq: Yes, will include that in separate PR since some changes are required in driver code to return proper error code for SSL.

sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
rupal-bq: added

Issue #601, opened 2020-07-21
ODBC: Excel for Mac does not return an error message on an invalid query


jordanw-bq:
**Problem**
Excel currently doesn't show an error message when an invalid query is entered into the Data Preview window when selecting tables.

**Cause**
ODFE SQL ODBC currently is missing support for SQLError(W), which is a deprecated ODBC 2.x function used to return error information. Exposing this method through the driver should resolve the issue.

PR #611, opened 2020-07-23
ODBC: Report error from Excel when executing an invalid query

jordanw-bq:
*Issue #, if available:* #601 

*Description of changes:*
- save error type and details on connection errors
- start checking error type when reporting error (first: query syntax)

### DIFF
--- a/sql-odbc/src/odfesqlodbc/odbcapi.c
+++ b/sql-odbc/src/odfesqlodbc/odbcapi.c
@@ -1378,16 +1378,16 @@ RETCODE SQL_API SQLError(SQLHENV EnvironmentHandle, SQLHDBC ConnectionHandle,
 
     MYLOG(ES_TRACE, "entering\n");
 
-    if (EnvironmentHandle) {
+    if (StatementHandle) {
         ret =
-            ESAPI_EnvError(EnvironmentHandle, RecNumber, Sqlstate, NativeError,
+            ESAPI_StmtError(StatementHandle, RecNumber, Sqlstate, NativeError,
                            MessageText, BufferLength, TextLength, 0);
     } else if (ConnectionHandle) {
         ret = ESAPI_ConnectError(ConnectionHandle, RecNumber, Sqlstate,
                                  NativeError, MessageText, BufferLength,
                                  TextLength, 0);
-    } else if (StatementHandle) {
-        ret = ESAPI_StmtError(StatementHandle, RecNumber, Sqlstate, NativeError,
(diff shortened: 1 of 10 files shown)


sql-odbc/src/odfesqlodbc/odbcapi.c
```diff
@@ -1367,4 +1367,33 @@ SQLRETURN SQL_API SQLColAttributes(SQLHSTMT StatementHandle,
...
+RETCODE SQL_API SQLError(SQLHENV EnvironmentHandle, SQLHDBC ConnectionHandle,
+                         SQLHSTMT StatementHandle, SQLCHAR *Sqlstate,
+                         SQLINTEGER *NativeError, SQLCHAR *MessageText,
+                         SQLSMALLINT BufferLength, SQLSMALLINT *TextLength) {
+    RETCODE ret;
+    SQLSMALLINT RecNumber = 1;
+
+    MYLOG(ES_TRACE, "entering\n");
+
+    if (EnvironmentHandle) {
```
jduo, line 1381: You should check this Statement, connection, then environment rather than the other way around. The most recently constructed handle gets priority.

sql-odbc/src/odfesqlodbc/odbcapiw.c
```diff
@@ -1005,3 +1006,63 @@ RETCODE SQL_API SQLSetConnectOptionW(HDBC ConnectionHandle, SQLUSMALLINT Option,
...
+                              NativeError, (SQLCHAR *)mtxt, buflen, &tlen, 0);
+    } else {
+        ret = SQL_ERROR;
+    }
+
+    if (SQL_SUCCEEDED(ret)) {
+        if (Sqlstate)
+            utf8_to_ucs2(qstr_ansi, -1, Sqlstate, 6);
+        if (mtxt && tlen <= BufferLength) {
+            SQLULEN ulen = utf8_to_ucs2_lf(mtxt, tlen, FALSE, MessageText,
```
jduo, line 1045: SQLCHAR inputs aren't guaranteed to be UTF-8. On Windows it depends on the OS codepage for example (usually codepage CP1252). This should likely be a separate issue to fix.

jordanw-bq: :+1: Will log another issue for this

sql-odbc/src/odfesqlodbc/odbcapi.c
jordanw-bq: Reversed order of handle checks

PR #620, opened 2020-07-27
Add Text and Keyword Data Type

penghuo:
## Description of changes
1. Add the Text and Keyword Data Type in Elasticsearch Storage Engine.
2. Redefine ExprValue interface.
    1.add the observer method to retrieve the value.
    2. add the compareTo method.
    3. add the equlas method.
3. Simplify the equal and notEqual operator definition in the BinaryPredicateOperator.
4. Add the doc for Data Types. https://github.com/penghuo/sql/blob/text-keyword-data-type/docs/user/general/datatype.rst

## To Reviewers
1.  I am planning an follow up PR to redefine all the operators in BinaryPredicateOperator. The formal doc for operator will been added accordingly.


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/FunctionDSL.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Unary Function Implementation.
+   *
+   * @param function   {@link ExprValue} based unary function.
+   * @param returnType return type.
+   * @param argsType   argument type.
+   * @return Unary Function Implementation.
+   */
+  public SerializableFunction<FunctionName, Pair<FunctionSignature, FunctionBuilder>> impl(
```
dai-chen, line 80: np: the `impl()` for binary function returns `Function` as below rather than `SerializableFunction`?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/FunctionDSL.java
```diff
@@ -0,0 +1,160 @@
...
+          new FunctionSignature(functionName, Collections.singletonList(argsType));
+      FunctionBuilder functionBuilder =
+          arguments -> new FunctionExpression(functionName, arguments) {
+            @Override
+            public ExprValue valueOf(Environment<Expression, ExprValue> valueEnv) {
+              ExprValue value = arguments.get(0).valueOf(valueEnv);
+              if (value.isMissing()) {
+                return ExprValueUtils.missingValue();
+              } else if (value.isNull()) {
+                return ExprValueUtils.nullValue();
```
dai-chen, line 96: I may miss it somewhere. Where is this MISSING/NULL check logic for the `impl()` below?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/FunctionDSL.java
penghuo: Done.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/FunctionDSL.java
penghuo: Good points.

Remove the NULL/MISSING generic handling logic in FunctionDSL.impl, and provide another utilit method nullMissingHandling to the client. The reason for doing that is (1) most of operator/function follow the generic missing/null handling logic. (2) but some doesn't follow that e.g. and/or/xor. 
```
  public SerializableFunction<ExprValue, ExprValue> nullMissingHandling(
      SerializableFunction<ExprValue, ExprValue> function) {
    return value -> {
      if (value.isMissing()) {
        return ExprValueUtils.missingValue();
      } else if (value.isNull()) {
        return ExprValueUtils.nullValue();
      } else {
        return function.apply(value);
      }
    };
  }
```
The client could make decision whetehr it want to follow the generic rule or not.
```
FunctionDSL.impl(FunctionDSL.nullMissingHandling(DateTimeFunction::exprDayOfMonth),
            INTEGER, DATE)
```

PR #627, opened 2020-07-28
ODBC: Fix for data loading failure in Power BI Desktop

rupal-bq:
*Issue #595*

*Description of changes:*
- Added data type & SQL data type values in SQLColumns
- Tested with Power BI Desktop (using import mode)


sql-odbc/src/IntegrationTests/ITODBCResults/test_odbc_results.cpp
```diff
@@ -936,7 +937,9 @@ TEST_F(TestSQLDescribeCol, SingleColumnMetadata) {
     EXPECT_EQ(single_col, m_column_name);
     EXPECT_EQ(single_col_name_length, m_column_name_length);
     EXPECT_EQ(single_col_data_type, m_data_type);
-    EXPECT_EQ(single_col_column_size, m_column_size);
+    // Value changes according to pagination setup on server
```
kylepbit, line 940: Why is this?

sql-odbc/src/odfesqlodbc/es_info.cpp
```diff
@@ -476,8 +492,28 @@ void SetTableTuples(QResultClass *res, const TableResultSet res_type,
         // information for its Data Preview window.
         std::string catalog("");
         bind_tbl[TABLES_CATALOG_NAME]->UpdateData((void *)catalog.c_str(), 0);
-        for (size_t i = 0; i < binds.size(); i++)
-            binds[i]->AssignData(&tuple[i]);
+
+        for (size_t i = 0; i < binds.size(); i++) {
+            // Add tuples for SQLColumns
+            if (binds.size() > COLUMNS_SQL_DATA_TYPE) {
```
kylepbit, line 498: Should we just report the extra columns regardless?

rupal-bq: The same method is used while adding tuples for SQLTables as well. SQLTables has only 5 columns whereas SQLColumns has 18. 
So this check ensures that the data type column is updated for SQLColumns only. If we remove this check then we get Memory Corruption since SQLTables will try to access column which doesn't exist.

sql-odbc/src/IntegrationTests/ITODBCResults/test_odbc_results.cpp
rupal-bq: Not sure why. Created an issue #628 to investigate this further since I got one another value as 23 for server `1.9.0`. 

Will revert changes in this PR and handle this separately as it's not a blocker for fixing the data loading issue in Power BI.

sql-odbc/src/odfesqlodbc/es_parse_result.cpp
```diff
@@ -29,6 +29,7 @@
 #pragma clang diagnostic pop
 #endif  // __APPLE__
 #include "statement.h"
+#include "es_types.h"
```
jordanw-bq, line 32: Already included on L22 (might need to expand GH code preview)

sql-odbc/src/odfesqlodbc/es_info.cpp
jordanw-bq: Almost wondering if it would be worth it to split some of this logic up, now that we're handling special cases for SQLTables & SQLColumns

sql-odbc/src/IntegrationTests/ITODBCCatalog/test_odbc_catalog.cpp
```diff
@@ -85,6 +85,9 @@ const std::vector< std::string > flights_data_type = {
     "float",   "keyword", "integer", "keyword", "keyword", "keyword", "date",
     "keyword", "keyword", "boolean", "float",   "keyword", "keyword", "keyword",
     "keyword", "keyword", "keyword", "keyword"};
+const std::vector< short > flights_sql_data_type = {
```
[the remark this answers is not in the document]
rupal-bq: added

sql-odbc/src/odfesqlodbc/es_info.cpp
rupal-bq: Created issue #630 to revisit the logic of assigning data for SQLTables & SQLColumns

sql-odbc/src/odfesqlodbc/es_parse_result.cpp
rupal-bq: removed

PR #635, opened 2020-07-29
Add comparsion operator for SQL

penghuo:
## Description of changes:*
1. Add Comparsion operator support for SQL.
https://github.com/penghuo/sql/blob/pr-operator-enhancement/docs/user/dql/expressions.rst#comparison-operators
2. Refactor the BinaryPredicateOperator and UnaryPredicateOperator.

### DIFF
--- a/core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperatorTest.java
+++ b/core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperatorTest.java
@@ -30,6 +30,7 @@ import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtil
 import static com.amazon.opendistroforelasticsearch.sql.data.type.ExprCoreType.BOOLEAN;
 import static com.amazon.opendistroforelasticsearch.sql.data.type.ExprCoreType.INTEGER;
 import static com.amazon.opendistroforelasticsearch.sql.data.type.ExprCoreType.STRING;
+
 import static com.amazon.opendistroforelasticsearch.sql.utils.ComparisonUtil.compare;
 import static com.amazon.opendistroforelasticsearch.sql.utils.OperatorUtils.matches;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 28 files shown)


docs/user/index.rst
```diff
@@ -22,6 +22,8 @@ Open Distro for Elasticsearch SQL enables you to extract insights out of Elastic
   - `Identifiers <general/identifiers.rst>`_
   - `Data Types <general/datatype.rst>`_
 
+  - `Data Types <general/datatype.rst>`_
```
dai-chen, line 25: Duplicate?

docs/user/dql/expressions.rst
```diff
@@ -92,6 +92,83 @@ Here is an example for different type of arithmetic expressions::
...
+| IS MISSING     | MISSING value test             |
++----------------+--------------------------------+
+| IS NOT MISSING | NOT MISSING value test         |
++----------------+--------------------------------+
+
+
+Basic Comparison Operator
+-------------------------
+
+Here is an example for different type of comparison operators::::
```
dai-chen, line 136: np: should be `::` here and below?

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperatorTest.java
```diff
@@ -458,46 +458,46 @@ public void test_less_null() {
     FunctionExpression less = dsl.less(DSL.literal(1),
         DSL.ref(INT_TYPE_NULL_VALUE_FIELD, INTEGER));
     assertEquals(BOOLEAN, less.type());
-    assertEquals(LITERAL_NULL, less.valueOf(valueEnv()));
+    assertEquals(LITERAL_FALSE, less.valueOf(valueEnv()));
```
dai-chen, line 461: So `1 < NULL` returns `FALSE` now?

docs/user/index.rst
penghuo: Done.

docs/user/dql/expressions.rst
penghuo: Done.

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperatorTest.java
penghuo: This is the default order in our system.
NULL equal to NULL
NULL large than MISSING
NULL less than all others.

In Order By, the user can also specify the order of NULL and MISSING.

penghuo: Got your point, change the logic, follow the MySQL way now.
https://dev.mysql.com/doc/refman/8.0/en/working-with-null.html

Issue #651, opened 2020-07-31
Bind parameter error while using advanced editor in Power BI Desktop (import mode)


rupal-bq:
Loaded data into Power BI Desktop using import mode successfully.

Used following query in **Transform Data** > **Advanced Editor**
```
let
    Source = OdfeSqlOdbcPBIConnector.Contents("localhost", 9200),
    kibana_sample_data_flights_null_null = Source{[Item="kibana_sample_data_flights",Schema=null,Catalog=null]}[Data],
    test_and = Table.SelectRows(kibana_sample_data_flights_null_null, each [OriginWeather] = "Rain" and [OriginCountry] = "CA" ) 
in
    test_and
```
But got following error while applying the changes.

![image](https://user-images.githubusercontent.com/56703525/89070478-ed811100-d329-11ea-8553-1f565d1fd68d.png)

SQL.LOG shows following error
```
Microsoft.Mashu 20c0-463c	EXIT  SQLBindParameter  with return code -1 (SQL_ERROR)
		HSTMT               0x000001DCA0A0BDA0
		UWORD                        1 
		SWORD                        1 <SQL_PARAM_INPUT>
		SWORD                       -8 <SQL_C_WCHAR>
		SWORD                       -9 <SQL_WVARCHAR>
		SQLULEN                    4
		SWORD                        0 
		PTR                0x000001DCBAF424A8
		SQLLEN                    16
		SQLLEN *            0x000001DCBAF424A0
		DIAG [HYC00] Elasticsearch does not support parameters. (10) 
```

PR #652, opened 2020-07-31
ODBC: Use literals instead of parameters in Power BI data connector

rupal-bq:
*Issue #651:*

*Description of changes:*
- added support for literals instead of parameters
- disabled direct query
- updated user docs


sql-odbc/src/PowerBIConnector/OdfeSqlOdbcPBIConnector.pq
```diff
@@ -106,6 +106,20 @@ OdfeSqlOdbcPBIConnectorImpl = (Host as text, optional Port as number) as table =
...
+        SQLGetFunctions = Diagnostics.LogValue("SQLGetFunctions_Options", [
+            SQL_API_SQLBINDPARAMETER = false
+        ]),
+
+        SqlCapabilities = Diagnostics.LogValue("SqlCapabilities_Options", [
+            SupportsTop = false,
+            LimitClauseKind = LimitClauseKind.LimitOffset,
+            SupportsNumericLiterals = true,
+            SupportsStringLiterals = true,
+            SupportsOdbcDateLiterals = true,
```
jordanw-bq, line 119: Can you confirm these options for date/time/timestamp literals? Wondering if this refers to the `{dt ...}`-type syntax we've seen before, which ODFE SQL plugin doesn't support

rupal-bq: Yes, this refers to `{d ..}` syntax. But if we don't set this to true then date values will always be specified using Parameter Binding (https://docs.microsoft.com/en-us/power-query/odbc) which is also not supported.

I will check if it is possible to report the escape sequence syntax unsupported.

rupal-bq: ODBC driver sets `SQL_DATETIME_LITERALS` to 0 in SQLGetInfo and a value of "0" means that SQL-92 datetime literals are not supported. (https://docs.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetinfo-function?view=sql-server-ver15)

jordanw-bq: Verified with @rupal-bq that timestamp values work with the current PBI connector; will look back at this when we include DirectQuery support

PR #663, opened 2020-08-05
Expression pushdown optimization

dai-chen:
*Issue #, if available:*

*Description of changes:* 

 1. **New script language**: Register the expression in new query engine as a script language in Elasticsearch. In this case, we don't need to maintain painless source for each expression.
 2. **The first half**: serialize the expression and send it along with DSL request.
 3. **The second half**: the code in DSL is deserialized to run on each node as if built-in painless script.
 4. **Expression visitor**: Add expression node visitor because both first and second half needs to visit expression tree.

*Limitation*:

 1. **Lucene query optimization**: For now, the entire expression would be pushed down to a script query. There will be another PR for Lucene query API pushdown soon.
 2. **Multi-field name**: Assumption is made that the inner field name is always "keyword", so "text.keyword" is used to get doc value for multi-field.
 3. **Serialization**: for now JDK serializer is in use. Due to not all expressions serializable, ex. arithmetic expression, pipelining filter operator will be used as before if serialization error happens.

*Testing*: Add UT and pass all ITs. Because of explain API missing in new engine, new IT will be added later.


elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/data/value/ElasticsearchExprValueFactory.java
```diff
@@ -121,6 +124,49 @@ private ExprValue construct(String field, JsonNode value) {
...
+    } else if (type.equals(TIMESTAMP)) {
+      if (value instanceof Number) {
+        return constructTimestamp((Long) value);
+      } else if (value instanceof Instant) {
+        return constructTimestamp((Instant) value);
+      } else {
+        return constructTimestamp(String.valueOf(value));
+      }
+    } else if (ES_TEXT.isCompatible(type)) {
+      return new ElasticsearchExprTextValue((String) value);
```
penghuo, line 162: Add some comments?

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/ElasticsearchIndex.java
```diff
@@ -87,6 +92,20 @@ public PhysicalPlan implement(LogicalPlan plan) {
      * index scan.
      */
     return plan.accept(new DefaultImplementor<ElasticsearchIndexScan>() {
+          @Override
+          public PhysicalPlan visitFilter(LogicalFilter node, ElasticsearchIndexScan context) {
```
penghuo, line 96: What happen if there are multiple filter, e.g. source=index | filter a = 1 | filter b = 1

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/ExpressionScriptEngine.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import lombok.RequiredArgsConstructor;
+import org.elasticsearch.script.FilterScript;
+import org.elasticsearch.script.ScriptContext;
+import org.elasticsearch.script.ScriptEngine;
+
+/**
+ * Custom expression script engine that supports using our expression code in DSL
```
penghuo, line 32: our expression -> core engine expression?

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
```diff
@@ -0,0 +1,169 @@
...
+                                SearchLookup lookup,
+                                LeafReaderContext context,
+                                Map<String, Object> params) {
+    super(params, lookup, context);
+    this.expression = expression;
+  }
+
+  @Override
+  public boolean execute() {
+    // Check we ourselves are not being called by unprivileged code.
```
penghuo, line 69: we ourserlves?

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
```diff
@@ -0,0 +1,169 @@
...
+    } else {
+      return value;
+    }
+  }
+
+  private ExprValue evaluateExpression(Map<String, ExprValue> valueEnv) {
+    ExprTupleValue tupleValue = ExprTupleValue.fromExprValueMap(valueEnv);
+    ExprValue result = expression.valueOf(tupleValue.bindingTuples());
+
+    if (result.isNull()) {
```
penghuo, line 157: Missing handling?

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
```diff
@@ -0,0 +1,169 @@
...
+
+  @Override
+  public boolean execute() {
+    // Check we ourselves are not being called by unprivileged code.
+    SpecialPermission.check();
+
+    return AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> {
+      Set<ReferenceExpression> fields = extractFields(expression);
+      ElasticsearchExprValueFactory valueFactory = buildValueFactory(fields);
+      Map<String, ExprValue> valueEnv = buildValueEnv(fields, valueFactory);
```
penghuo, line 75: NP, Could we impl BindingTuple and encapsulate these logic in that to achive the same concept of valueEnv.

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/data/value/ElasticsearchExprValueFactory.java
dai-chen: Sure, will add a new `ExprValue` class to differentiate text and multi-field.

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/ElasticsearchIndex.java
dai-chen: Good point. Will double check and add IT to make sure this can be handled properly. Thanks!

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
dai-chen: Since no missing concept in JAVA, will add a `MISSING` constant in value factory. And will add test case for `null` field.

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
dai-chen: Sure, will add a new `Environment` class to encapsulate the map data structure. Thanks!

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/ExpressionScriptEngine.java
dai-chen: Done!

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
dai-chen: Done!

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/ExpressionFilterScript.java
dai-chen: I added missing value check here. But I found that there seems no way to differentiate null and missing in doc value. I tried the following index. Both `null_field` and `missing_field` in 2nd doc gives a script doc value object with empty list inside. Will confirm and come back to this later.

```
PUT null_and_missing/_doc/1
{
  "null_field": "abc",
  "missing_field": 456
}

PUT null_and_missing/_doc/2
{
  "null_field": null
}
```

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/ElasticsearchIndex.java
dai-chen: Added logic to handle multiple filter operator pushdown and UT/IT to test. We may need to improve our planner soon since current index scan operator assumes all push down is around original data source. I guess if we do `where | stats | where`, the result would be wrong.

PR #671, opened 2020-08-07
Lucene query pushdown optimization

dai-chen:
*Issue #, if available:*

*Description of changes:* The major changes in this PR are new added `LuceneQuery` abstraction and its subclass to implement different Lucene query APIs. To clarify, Lucene query here means Lucene query via Elasticsearch DSL. We don't bypass DSL and call Lucene directly.

*Problem Statement*: In PR https://github.com/opendistro-for-elasticsearch/sql/pull/663, we register our expression as new script language and optimize expression evaluation by pushdown to script query in Elasticsearch DSL. However, what was not covered yet is that how we can leverage Lucene API for expression that can be optimized. For certain expression, we want to optimize filtering expression further by pushing down to Lucene query fully or partially.

*Solution*: The core logic is in `FilterQueryBuilder.visitFunction` method which is executed in the top down way:

 1. `AND`, `OR`, `NOT`: Translate to bool query and visit left/right side recursively
 2. Functions that Lucene may support: Translate to Lucene query if left side is a reference (field name) and right is a literal. Otherwise go to 3.
 3. Functions that Lucene doesn't support or argument is not valid as mentioned in 2: Serialize the expression (subtree with current function node as root) and translate to script query.

Here is an example to help understand. `AND` is translated to bool filter query directly (case #1). The `name = 'John'` is eligible because `=` can be translated to Lucene term query and left side (first argument) is a reference and right side is a literal (case #2). `ABS(age) = 30` is translated to a script query (case #3).

lucene-pushdown

*Testing*: new UT and PPL IT can pass. Since we don't have explain API yet, SQL IT and doctest will be added later.


elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/filter/lucene/TermQuery.java
```diff
@@ -0,0 +1,39 @@
...
+import org.elasticsearch.index.query.QueryBuilders;
+
+/**
+ * Lucene query that build term query for equality comparison.
+ */
+public class TermQuery extends LuceneQuery {
+
+  @Override
+  protected QueryBuilder doBuild(String fieldName, ExprType fieldType, ExprValue literal) {
+    if (fieldType == ES_TEXT_KEYWORD) { // Assume inner field name is always "keyword"
```
penghuo, line 33: could you add doc for this limitation?

dai-chen: Good idea! Added a `limitation` folder with a single doc for now. Copy-pasted the up-to-date content from ODFE doc. Will make it doctest with example in explain API PR. Thanks!

1. MySQL reference: https://dev.mysql.com/doc/mysql-reslimits-excerpt/8.0/en/
2. ODFE doc: https://opendistro.github.io/for-elasticsearch-docs/docs/sql/limitation/
3. New limitation doc: https://github.com/dai-chen/sql/blob/lucene-pushdown-optimization/docs/user/limitations/limitations.rst

PR #687, opened 2020-08-15
Add temporal interval data type and support interval clause as function

chloe-zh:
*Issue #, if available:*

*Description of changes:*

Doc: https://github.com/chloe-zh/sql/blob/interval/docs/user/general/datatypes.rst

- Added `INTERVAL` data type in the SQL language structure. Syntax: `INTERVAL expr unit`. The corresponding expression value is `ExprIntervalValue` that can encapsulate the temporal amount as the property value.

- Quantity expression `expr` in interval clause can be any value expression, including the basic numeric values, field expression, functions etc. which has been revealed in the parser.

- Interval units to support for phase 1: microsecond, second, minute, hour, day, week, month, quarter, year.

- Added docs for date and time data types.

- *[TODO]* Current PR does not contain examples of date and time types in the reference manual yet. These examples will be added in the coming PR for datetime functions where functions `date`, `time`, `datetime`, `timestamp` and `interval` are implemented.

**Note:** 

Interval data type maps to `TemporalAmount` in Java runtime, it has two types of amount `Duration` subclass that can be mapped with microseconds to days, where all amount would be converted to a value with `second` as the unit; and another amount `Period` can be mapped with days to years, where the values would be converted to some value with `day` as the the unit. This rule is similar to H2: https://www.h2database.com/html/datatypes.html#interval_type 

In this PR, these two types are sharing the same syntax, but will differ in the storage method, and is only comparable with the same type of interval. For example, the comparison between `INTERVAL 1 week` with `INTERVAL 1 month` will result in 1 week is shorter than 1 month, but an attempt to compare `INTERVAL 1 week` with `INTERVAL 1 minute` will fail and throw an exception. Which is also similar to H2 behavior.


docs/user/general/datatypes.rst
```diff
@@ -0,0 +1,206 @@
...
++==========+==================================+==============================================================+
+| Datetime | 'yyyy-MM-dd hh:mm:ss[.fraction]' | '0001-01-01 00:00:00.000000' to '9999-12-31 23:59:59.999999' |
++----------+----------------------------------+--------------------------------------------------------------+
+
+
+
+Timestamp
+---------
+
+A timestamp instance is an absolute instant independent of timezone or convention. For example, for a given point of time, if we set the timestamp of this time point into another timezone, the value should also be different accordingly. Bides, the storage of timestamp type is also different from the other types. The timestamp is converted from the current timezone to UTC for storage, and is converted back to the set timezone from UTC when retrieving.
```
penghuo, line 138: Typo, Bides -> ?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprValueUtils.java
```diff
@@ -172,4 +171,8 @@ public static Boolean getBooleanValue(ExprValue exprValue) {
   public static ZonedDateTime getDateValue(ExprValue exprValue) {
     return exprValue.dateValue();
   }
+
+  public static TemporalAmount getIntervalValue(ExprValue exprValue) {
```
penghuo, line 175: do we need this method?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/expression/DataType.java
```diff
@@ -29,13 +29,15 @@
...
   INTEGER(ExprCoreType.INTEGER),
+  LONG(ExprCoreType.LONG),
   DOUBLE(ExprCoreType.DOUBLE),
   STRING(ExprCoreType.STRING),
   BOOLEAN(ExprCoreType.BOOLEAN),
 
   DATE(ExprCoreType.DATE),
   TIME(ExprCoreType.TIME),
-  TIMESTAMP(ExprCoreType.TIMESTAMP);
+  TIMESTAMP(ExprCoreType.TIMESTAMP),
```
penghuo, line 39: Datetime is not implemented yet?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/FunctionDSL.java
```diff
@@ -56,7 +57,7 @@ public FunctionResolver define(FunctionName functionName,
    * @param functions    a list of function implementation.
    * @return FunctionResolver.
    */
-  public FunctionResolver define(FunctionName functionName,
+  public static FunctionResolver define(FunctionName functionName,
```
penghuo, line 60: @UtilityClass will add static automatically.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/IntervalClause.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private ExprValue year(ExprValue value) {
+    return new ExprIntervalValue(Period.ofYears(getIntegerValue(value)));
+  }
+
+  /**
+   * Interval argument implementation as function.
+   */
+  private static SerializableFunction
```
penghuo, line 140: could we reuse the FunctionDSL.impl?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprValueUtils.java
chloe-zh: yes, i've also added getTimeValue, getDatetimeValue, getTimestampValue in the other PR #694  to stay consistent. We can also remove all these methods including the getDateValue actually

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/expression/DataType.java
chloe-zh: it's in #694

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/IntervalClause.java
chloe-zh: `toString()` method needs to be overridden

docs/user/general/datatypes.rst
chloe-zh: Corrected it, thanks

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/IntervalClause.java
dai-chen: > `toString()` method needs to be overridden

Where is the `toString` method used? I recall it was required earlier because we didn't have original field name in `SELECT`?

dai-chen:
So in this PR the changes regarding parser only covers PPL language?

chloe-zh:
> So in this PR the changes regarding parser only covers PPL language?

Added parser in SQL in the latest update, thanks!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/FunctionDSL.java
chloe-zh: Removed static keywords

dai-chen:
> > So in this PR the changes regarding parser only covers PPL language?
> 
> Added parser in SQL in the latest update, thanks!

Thanks for adding! In this case please add comparison test case to make sure it can work with JDBC driver together.

PR #694, opened 2020-08-18
Added datetime data type and basic functions for the conversion from string type to date and time types

chloe-zh:
*Issue #, if available:*

*Description of changes:*

- Added datetime type
- Modified the properties and methods in expression values of date, time, timestamp types
- Added function `date`, `datetime`, `time`, `timestamp` as the conversion method to convert string to date and time types


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/DateTimeFunction.java
```diff
@@ -58,4 +135,30 @@ private FunctionResolver dayOfMonth() {
...
   }
+
+  /**
+   * Time implementation for ExprValue.
+   * @param exprValue ExprValue of Time type or String.
+   * @return ExprValue.
+   */
+  private ExprValue exprTime(ExprValue exprValue) {
+    if (exprValue instanceof ExprStringValue) {
+      return new ExprTimeValue(getStringValue(exprValue));
```
penghuo, line 146: suggest to use exprValue.string() in future to get rid of the static getStringValue method.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprValueUtils.java
```diff
@@ -166,12 +173,33 @@ public static Boolean getBooleanValue(ExprValue exprValue) {
...
    */
-  public static ZonedDateTime getDateValue(ExprValue exprValue) {
+  public static LocalDate getDateValue(ExprValue exprValue) {
     return exprValue.dateValue();
   }
 
+  /**
+   * Get {@link LocalTime} from ExprValue of Time type.
+   */
+  public static LocalTime getTimeValue(ExprValue exprValue) {
```
penghuo, line 185: could we get rid of the static method incase we already have the method impl in ExprValue.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprDatetimeValue.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Override
+  public int compare(ExprValue other) {
+    return datetime.compareTo(other.datetimeValue());
+  }
+
+  @Override
+  public boolean equal(ExprValue other) {
+    return datetime.equals(other.datetimeValue());
+  }
```
dai-chen, line 78: Just wonder any reason not use lombok annotation for equal and hashCode?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprValueUtils.java
chloe-zh: yes, will remove all similar static methods and use the impl instead

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprDatetimeValue.java
chloe-zh: for the equal method, because  it is an implementation of abstract method

chloe-zh:
#709

PR #716, opened 2020-08-27
Support date, time, timestamp types in JDBC

chloe-zh:
*Issue #, if available:*

*Description of changes:*

Changed mappings between es and types of date, time, timestamp.
----
Previously:

| ES type from mapping | to JDBC type |
| --- | --- |
| Date | Timestamp |

| from JDBC type | to ES type (schema) |
| --- | --- |
| Date | Date |
|Timestamp | Date | 

Currently:

| Mapping in ES | JDBC | Resulted schema |
| --- | --- | --- |
| Date | Timestamp | Timestamp |
|  | Date | Date |
|  | Time | Time |

Fixed the `dayofmonth` function
----
1. The implementation of `dayofmonth` for `ExprValue` was incorrect, fixed it to the correct one.
2. Allows the function with string type argument, and grants it as the date type to stay consistent with the other databases. For example, previously we only support `dayofmonth(date('2020-08-27'))` and return `27` as the result, but `dayofmonth('2020-08-27')` throws semantic exception; now we support both of these two inputs and return `27`.
3. Added comparison test for `dayofmonth` function.

### DIFF
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/DateTimeFunction.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/DateTimeFunction.java
@@ -78,7 +78,8 @@ public class DateTimeFunction {
   private FunctionResolver dayOfMonth() {
     return define(DAYOFMONTH.getName(),
         impl(nullMissingHandling(DateTimeFunction::exprDayOfMonth),
-            INTEGER, DATE)
+            INTEGER, DATE),
+        impl(nullMissingHandling(DateTimeFunction::exprDayOfMonth), INTEGER, STRING)
     );
   }
 
@@ -128,6 +129,10 @@ public class DateTimeFunction {
    * @return ExprValue.
    */
   private ExprValue exprDayOfMonth(ExprValue date) {
+    if (date instanceof ExprStringValue) {
+      return new ExprIntegerValue(
+          new ExprDateValue(date.stringValue()).dateValue().getMonthValue());
+    }
     return new ExprIntegerValue(date.dateValue().getMonthValue());
   }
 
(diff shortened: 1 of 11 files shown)


penghuo:
after we support dayofmonth('2020-08-27'), do it means the dayofmonth("no date") will pass the semantic analysis and throw exception during execution stage?

sql-jdbc/src/main/java/com/amazon/opendistroforelasticsearch/jdbc/types/ElasticsearchType.java
```diff
@@ -71,6 +73,8 @@
     NESTED(JDBCType.STRUCT, null, 0, 0, false),
     OBJECT(JDBCType.STRUCT, null, 0, 0, false),
     DATE(JDBCType.TIMESTAMP, Timestamp.class, 24, 24, false),
```
penghuo, line 75: DATE to TIMESTAMP?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/DateTimeFunction.java
```diff
@@ -78,7 +78,8 @@ private FunctionResolver date() {
   private FunctionResolver dayOfMonth() {
     return define(DAYOFMONTH.getName(),
         impl(nullMissingHandling(DateTimeFunction::exprDayOfMonth),
-            INTEGER, DATE)
+            INTEGER, DATE),
+        impl(nullMissingHandling(DateTimeFunction::exprDayOfMonth), INTEGER, STRING)
```
penghuo, line 82: could you confirm this? dayofmonth doesn't accept DATE? e.g. dayofmonth(CURDATE()) should work.

chloe-zh: It accepts date, this line is to addd string as argument as well

sql-jdbc/src/main/java/com/amazon/opendistroforelasticsearch/jdbc/types/ElasticsearchType.java
chloe-zh: DATE to DATE breaks the test by failing to load the `kibana_sample_data_flights` index. So I temporarily changed it back to TIMESTAMP. Will look into the issue

dai-chen (approved):
Thanks for the changes!

penghuo (approved):

Issue #717, opened 2020-08-27
Bool field gets integer value after aggregation


dai-chen:
Most aggregate functions works well with bool field:

```
POST _opendistro/_sql
{
  "query": "SELECT male, AVG(age) FROM bool-test GROUP BY male"
}
{
  "schema": [
    {
      "name": "male",
      "type": "boolean"
    },
    {
      "name": "AVG(age)",
      "type": "double"
    }
  ],
  "total": 1,
  "datarows": [[
    true,
    30
  ]],
  "size": 1
}
```

However, if `COUNT(*)` in use, the bool field value is changed to an integer `1` instead of `true`. I found that `COUNT(male)` or `COUNT(age)` doesn't have this issue either.

```
POST _opendistro/_sql
{
  "query": "SELECT male, COUNT(*) FROM bool-test GROUP BY male"
}
{
  "schema": [
    {
      "name": "male",
      "type": "boolean"
    },
    {
      "name": "COUNT(*)",
      "type": "integer"
    }
  ],
  "total": 1,
  "datarows": [[
    1,
    1
  ]],
  "size": 1,
  "status": 200
}
```

dai-chen:
This is actually happening in old engine.

dai-chen:
The issue has been fixed in new query engine (which is experimental and disabled by default). You can enable it to preview by changing the plugin setting: https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst#opendistro-sql-engine-new-enabled. Thanks!

Issue #643, opened 2020-07-30
"select `column` from `table` group by `column` order by `column` LIMIT 5" is not supported


rupal-bq:
Following query fails for columns with data type `text`,`nested` and `object`
Query:
```
"select `type_text` from `kibana_sample_data_types` group by `type_text` order by `type_text` LIMIT 5"
```
Response:
```
{
  "error": {
    "reason": "Error occurred in Elasticsearch engine: all shards failed",
    "details": "Shard[0]: java.lang.IllegalArgumentException: field [type_object] isn't a leaf field\n\nFor more details, please send request for Json format to see the raw response from elasticsearch engine.",
    "type": "SearchPhaseExecutionException"
  },
  "status": 400
}
```

The same query works for data type `keyword`.
Query:
```
"select `type_keyword` from `kibana_sample_data_types` group by `type_keyword` order by `type_keyword` LIMIT 5"
```
Response:
```
{
  "schema": [
    {
      "name": "type_keyword",
      "type": "keyword"
    }
  ],
  "total": 2,
  "datarows": [
    [
      "goodbye"
    ],
    [
      "hello"
    ]
  ],
  "size": 2,
  "status": 200
}
```

dai-chen:
For text field, I tested it by `accounts` index. I think this is the same issue as https://github.com/opendistro-for-elasticsearch/sql/issues/401#issuecomment-614324103. The root cause is that text field name (if it's a multi-field) is rewritten to its inner keyword field name because aggregation on text field requires fielddata enabled otherwise. However, field name in SELECT is not rewritten accordingly.

```
POST _opendistro/_sql
{  
  "query": """
      select `address` from `accounts` group by `address` order by `address` LIMIT 5
      """
}
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "invalid value operation on MISSING_VALUE",
    "type": "IllegalStateException"
  },
  "status": 500
}
```

This worked if we rewrote `address` in SELECT to `address.keyword` manually to match `GROUP BY` item:

```
POST _opendistro/_sql
{  
  "query": """
      select `address`.`keyword` from `accounts` group by `address`.`keyword` order by `address`.`keyword` LIMIT 5
      """
}
{
  "schema": [{
    "name": "address.keyword",
    "type": "keyword"
  }],
  "total": 5,
  "datarows": [
    ["100 Gatling Place"],
    ["101 Bush Street"],
    ["102 Barwell Terrace"],
[...]

Issue #430, opened 2020-04-14
Tableau query not working


abbashus:
**JDBC driver:** opendistro-sql-jdbc-1.4.0.0.jar
**SQL plugin:** opendistro-1.6 release
**Elasticsearch version:** 7.6.1 OSS

Query:

```
POST _opendistro/_sql?
{
  "query": """SELECT `accounts`.`gender` AS `gender`,
      SUM(`accounts`.`balance`) AS `sum_balance_ok`
      FROM `accounts`
      GROUP BY `accounts`.`gender`
"""
}
```

Response:
```
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "invalid value operation on MISSING_VALUE",
    "type": "IllegalStateException"
  },
  "status": 500
}
```

elasticsearch.log
```
[2020-04-14T10:27:47,759][WARN ][c.a.o.s.e.f.PrettyFormatRestExecutor] [f45c898a2ba7.ant.amazon.com] Error happened in pretty formatter
java.lang.IllegalStateException: invalid value operation on MISSING_VALUE
	at com.amazon.opendistroforelasticsearch.sql.expression.model.ExprValue.value(ExprValue.java:23) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.format.BindingTupleResultSet.lambda$buildDataRows$1(BindingTupleResultSet.java:55) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195) ~[?:?]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1621) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[?:?]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578) ~[?:?]
[...]

Issue #401, opened 2020-03-30
Group By * Query ERROR


EveyWu:
SQL By ``bin/elasticsearch-sql-cli `` is correct

## Test Data：
```
curl -X PUT "localhost:9200/guide/_bulk?refresh&pretty" -H 'Content-Type: application/json' -d'
{"index":{"_id": "358e4f2c-d796-4634-8f7e-ac014b71ab90"}}
{"regionCode":"440600","itemType":"05","orgName":"微博","id":"110253L3053","title":"热搜","region":"柬埔寨","orgNumber":"53","onlineHandle":"0"}
{"index":{"_id": "9653eedf-ed4d-4ebf-8e48-9cb86a317ca8"}}
{"regionCode":"441323106006","itemType":"21","orgName":"新浪","handleUrl":"http://weibo.com","id":"064421","title":"热度","region":"泰国","orgNumber":"","onlineHandle":"1"}
'
```

## Query:
```
http://localhost:9200/_opendistro/_sql?sql=SELECT COUNT(*), regionCode FROM guide GROUP BY regionCode
```
## Error :
```
Internal Server Error. Reason: There was internal problem at backend. Type: IllegalStateException. Details: invalid value operation on MISSING_VALUE
	at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.executeQueryRequest(StatementImpl.java:78)
	at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.executeQueryX(StatementImpl.java:61)
	at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.executeQuery(StatementImpl.java:54)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #724, opened 2020-08-31
Support group by in new sql parser

dai-chen:
*Issue #, if available:*

1. https://github.com/opendistro-for-elasticsearch/sql/issues/280: Chained function call or expression
2. https://github.com/opendistro-for-elasticsearch/sql/issues/401, https://github.com/opendistro-for-elasticsearch/sql/issues/430, https://github.com/opendistro-for-elasticsearch/sql/issues/643: Group by multi field
3. https://github.com/opendistro-for-elasticsearch/sql/issues/373, https://github.com/opendistro-for-elasticsearch/sql/issues/389: function name changed to lower case
4. https://github.com/opendistro-for-elasticsearch/sql/issues/288: group by function
5. https://github.com/opendistro-for-elasticsearch/sql/issues/717: bool field returns integer

*Description of changes:* This PR is to change the grammar in new SQL parser and integrate with core engine. The basic idea is:

1. Build `Aggregation` node with aggregators, so the output aggregation of SQL is same as PPL AST builder.
2. Only perform "static" validation in SQL frontend and leave more analysis and validations in analyzer in core engine.

To make 1st item happen, the major change is adding a new domain class `QuerySpecification` and separate AST building process into 2 phases. Copy and pasting JavaDoc for your reference. Please see JavaDoc on `QuerySpecification` and `AstAggregationBuilder` class.

```
* AST aggregation builder that builds AST aggregation node for the following scenarios:
 * 1. Explicit GROUP BY
 *     1.1 Group by column name or scalar expression:
 *          SELECT ABS(age) FROM test GROUP BY ABS(age)
 *     1.2 Group by alias in SELECT AS clause:
 *          SELECT state AS s FROM test GROUP BY s
 *     1.3 Group by ordinal referring to select list:
 *          SELECT state FROM test GROUP BY 1
 *  2. Implicit GROUP BY
 *     2.1 No non-aggregated item (only aggregate functions):
 *          SELECT AVG(age), SUM(balance) FROM test
 *     2.2 Non-aggregated item exists:
 *          SELECT state, AVG(age) FROM test
 *
 *  For 1.1 and 2.1, Aggregation node is built with aggregators.
 *  For 1.2 and 1.3, alias and ordinal is replaced first and then
 *    Aggregation is built same as above.
 *  For 2.2, Exception thrown for now. We may support this by different SQL mode.
```

*Documentation*: No impact from user perspective so the old doc still up-to-date.

*Testing*: Add comparison test for basic cases and all bug fixes above. 3 legacy IT has to be skipped and the reason is listed in `@Ignore` annotation.


chloe-zh (approved):

penghuo (approved):
Thanks for the change.

PR #743, opened 2020-09-17
Fix bug in SqlRequestFactoryTest

penghuo:
## Description of changes:
Fix bug in SqlRequestFactoryTest, the LocalClusterState should be initialized to avoid NullPointerException.


dai-chen (approved):
Thanks for the fix! Is the Mockito upgrade in this PR also required or not? https://github.com/opendistro-for-elasticsearch/sql/pull/740

penghuo:
> Thanks for the fix! Is the Mockito upgrade in this PR also required or not? #740

not necessary.

chloe-zh (approved):

Issue #745, opened 2020-09-18
Keywords as identifiers failed new engine and fell back to old engine


chloe-zh:
Symptom: when field name or alias is a keyword of SQL/PPL specification, it cannot pass the new engine semantic analysis and falls back to old engine. For example:

data set: `kibana_sample_data_fligths`

query: `SELECT timestamp FROM kibana_sample_data_flights`

expected schema:
```
  "schema": [
    {
      "name": "timestamp",
      "type": "timestamp"
    }
  ]
```

actual result:
```
  "schema": [{
    "name": "timestamp",
    "type": "date"
  }]
```
log: 
```
[2020-09-18T11:19:50,669][INFO ][c.a.o.s.l.p.RestSqlAction] [a483e711985a.ant.amazon.com] [f282d5e0-b3f1-4adb-bd68-2953131e486b] Incoming request /_opendistro/_sql?pretty=true: ( SELECT identifier FROM table )
[2020-09-18T11:19:50,728][WARN ][stderr                   ] [a483e711985a.ant.amazon.com] line 1:17 no viable alternative at input 'timestamp FROM'
```
---
This is because "timestamp" is parsed as a function keyword in the new engine and the query then goes into the old engine. If we quote the field `timestamp` using backticks:
query: 
```
SELECT `timestamp` FROM kibana_sample_data_flights
```
result schema:
```
  "schema": [
    {
      "name": "timestamp",
      "type": "timestamp"
    }
  ]
```
log:
```
[2020-09-18T11:24:07,106][INFO ][c.a.o.s.l.p.RestSqlAction] [a483e711985a.ant.amazon.com] [0022fc4a-a21e-4801-9fc3-195ed85d982e] Request SQLQueryRequest(jsonContent={"query":"SELECT `timestamp` FROM kibana_sample_data_flights"}, query=SELECT `timestamp` FROM kibana_sample_data_flights, path=/_opendistro/_sql, format=jdbc) is handled by new SQL query engine
```

PR #747, opened 2020-09-19
Fix: let keywords eligible to be parsed as identifiers

chloe-zh:
*Issue #, if available:*

#745 

*Description of changes:*

- Fixed the data type conversion of JDBC date and time types
- Added keywords as eligible identifiers in SQL and PPL ANTLR parser
- Excluded `AggregationExpressionIT` in legacy integration test due to inconsistency in JDBC schema between old engine and new engine


sql/src/main/antlr/OpenDistroSQLIdentifierParser.g4
```diff
@@ -44,10 +44,18 @@ alias
 
 qualifiedName
     : ident (DOT ident)*
+    | keywordsCanBeId
```
dai-chen, line 47: Can we also add functionNameBase here as before? https://github.com/opendistro-for-elasticsearch/sql/blob/master/legacy/src/main/antlr/OpenDistroSqlParser.g4#L237. Although this is a separate g4 file, I guess it may have access to rules in parent g4.

sql-jdbc/src/main/java/com/amazon/opendistroforelasticsearch/jdbc/types/ElasticsearchType.java
```diff
@@ -72,7 +72,7 @@
     IP(JDBCType.VARCHAR, String.class, 15, 0, false),
     NESTED(JDBCType.STRUCT, null, 0, 0, false),
     OBJECT(JDBCType.STRUCT, null, 0, 0, false),
-    DATE(JDBCType.TIMESTAMP, Timestamp.class, 24, 24, false),
```
dai-chen, line 75: Are the changes in UT meant for this? I can see we're changing `date` type in test data and code to `timestamp` in UT?

ppl/src/main/java/com/amazon/opendistroforelasticsearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -191,6 +191,9 @@ public UnresolvedExpression visitEvalFunctionCall(EvalFunctionCallContext ctx) {
    */
   @Override
   public UnresolvedExpression visitQualifiedName(QualifiedNameContext ctx) {
+    if (ctx.keywordsCanBeId() != null) {
+      return new QualifiedName(ctx.keywordsCanBeId().getText());
+    }
```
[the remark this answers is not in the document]
chloe-zh: Yes we can, that introduces more changes so I didn't add label. But yes I can use label since it makes code more readable

sql/src/main/antlr/OpenDistroSQLIdentifierParser.g4
chloe-zh: It looks like it cannot read the `functionNameBase` so I added them separately. Do they have heritance relation, I only see `OpenDistroSQLParser` imports Identifier parser file

sql-jdbc/src/main/java/com/amazon/opendistroforelasticsearch/jdbc/types/ElasticsearchType.java
chloe-zh: Yes, all timestamp type were taken as date type in JDBC, here it's corrected back to timestamp and so are the UTs

sql/src/main/antlr/OpenDistroSQLIdentifierParser.g4
dai-chen: I see. Thanks!

dai-chen (approved):
Thanks for the fix!

penghuo (approved):
Thanks

PR #748, opened 2020-09-19
Support expression over aggregation

penghuo:
## Description of changes:
Enhancement the aggregation support in SQL.
https://github.com/penghuo/sql/blob/poc-ref-reduce/docs/user/dql/aggregations.rst


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/AggregationExpressionOptimizer.java
```diff
@@ -0,0 +1,113 @@
...
+ * Project[abs(age), sub(sum(age), avg(age))
+ *  Agg(agg=[sum(age), avg(age)], group=[abs(age)]]
+ *   Relation
+ * The sum(age) and avg(age) in the Project could be replace by the analyzed reference, the
+ * LogicalPlan should be
+ * LogicalProject[Ref("abs(age)"), sub(Ref("sum(age)"), Ref("avg(age)"))
+ *  LogicalAgg(agg=[sum(age), avg(age)], group=[abs(age)]]
+ *   LogicalRelation
+ */
+public class AggregationExpressionOptimizer
```
dai-chen, line 47: Could we name this more abstract? Because we may need to replace expression with reference elsewhere than aggregation, for example, collect expression in window operator and replace it in project list.

dai-chen (approved):
Just one minor comment. Thanks for the changes!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/AggregationExpressionOptimizer.java
penghuo: Done.

chloe-zh (approved):

Issue #754, opened 2020-09-23
PPL Sort command return empty result when count=0


penghuo:
## Current result
```
source=kibana_sample_data_flights | sort 0 FlightDelayMin | fields FlightDelayMin

{
  "schema": [
    {
      "name": "FlightDelayMin",
      "type": "integer"
    }
  ],
  "datarows": [],
  "total": 0,
  "size": 0
}
```
## Expectation
* Count=0, should return all the result,
* Count=$number, should return $number result,
* Missing Count argument, should return default size = 1000

PR #755, opened 2020-09-23
Bug fix, sort command should return all the result when count = 0

penghuo:
*Issue #, if available:* #754 


dai-chen (approved):
Thanks for the fix!

chloe-zh (approved):

PR #761, opened 2020-09-29
[UI] Separate SQL and PPL pages

chloe-zh:
*Issue #, if available:*

*Description of changes:*

- Separate to SQL and PPL pages that can be switched by the SQL/PPL toggle
- Added `PPL Documentation` link popout
- Corrected the download routes for PPL results
- Added modal with error message when download ppl results of json/csv/raw formats since these formats are unsupported yet

Video Walkthrough


workbench/public/components/Main/main.tsx
```diff
@@ -203,6 +207,9 @@ export class Main extends React.Component<MainProps, MainState> {
     };
 
     this.httpClient = this.props.httpClient;
+    this.updateSQLQueries = _.debounce(this.updateSQLQueries, 250).bind(this);
```
joshuali925, line 210: Would there be a delay if you debounce state update when editor updates?

workbench/public/components/QueryResults/QueryResultsBody.tsx
```diff
@@ -127,15 +130,16 @@ class QueryResultsBody extends React.Component<QueryResultsBodyProps, QueryResul
...
-      isDownloadPopoverOpen: false
+      isDownloadPopoverOpen: false,
+      isModalVisible: false,
+      downloadErrorModal: {}
     };
 
     this.expandedRowColSpan = 0;
     this.items = [];
     this.columns = [];
     this.panels = [];
```
joshuali925, line 141: This line seems unnecessary

joshuali925 (approved):
Why is it necessary to have separate SQLPage and PPLPage components? Right now they seem to be the same component with different prop/state variable names

workbench/public/components/Main/main.tsx
chloe-zh: Yes there would be a minor delay of 250 millisecond from every call of this function, I think it's necessary and 250 ms is acceptable.

workbench/public/components/QueryResults/QueryResultsBody.tsx
chloe-zh: > Why is it necessary to have separate SQLPage and PPLPage components? Right now they seem to be the same component with different prop/state variable names

It's because we will have different components in SQL and PPL pages in our latest ux design, but not refracted in this PR yet. We will elaborate the details later.

chloe-zh: > This line seems unnecessary

Removed this line, thanks!

chloe-zh:
> Why is it necessary to have separate SQLPage and PPLPage components? Right now they seem to be the same component with different prop/state variable names

It's because we will have different components in SQL and PPL pages in our latest ux design, but not refracted in this PR yet. We will elaborate the details later.

joshuali925:
> It's because we will have different components in SQL and PPL pages in our latest ux design, but not refracted in this PR yet. We will elaborate the details later.

Got it, thanks!

anirudha (approved):

Issue #710, opened 2020-08-25
[draft] Support string functions for new engine


chloe-zh:
- [ ] ascii
- [ ] bin
- [ ] bit_length
- [ ] char
- [ ] char_length
- [ ] character_length
- [ ] concat
- [ ] concat_ws
- [ ] elt
- [ ] export_set
- [ ] field
- [ ] find_in_set
- [ ] format
- [ ] from_base64
- [ ] hex
- [ ] insert
- [ ] instr
- [ ] lcase
- [ ] left
- [ ] length
- [ ] like
- [ ] load_file
- [ ] locate
- [ ] lower
- [ ] lpad
- [ ] ltrim
- [ ] make_set
- [ ] match
- [ ] mid
- [ ] not_like
- [ ] not_regexp
- [ ] oct
- [ ] octet_length
- [ ] ord
- [ ] position
- [ ] quote
- [ ] regexp
- [ ] regexp_instr
- [ ] regexp_like
- [ ] regexp_replace
- [ ] regexp_substr
- [ ] repeat
- [ ] replace
- [ ] reverse
- [ ] right
- [ ] rlike
- [ ] rpad
- [ ] rtrim
- [ ] soundex
- [ ] sounds_like
- [ ] space
- [ ] strcmp
- [ ] substr
- [ ] substring
- [ ] substring_index
- [ ] to_base64
- [ ] trim
- [ ] ucase
- [ ] unhex
- [ ] upper
- [ ] weight_string

dai-chen:
An old issue for `DIV` function support: https://github.com/opendistro-for-elasticsearch/sql/issues/228

PR #764, opened 2020-09-30
DATE_FORMAT function

lyndonb-bq:
*Issue #710 

*Description of changes:*
Added implementation, unit tests, manual integration tests, and documentation for date_format function.


chloe-zh (approved):
LGTM!

rupal-bq (approved):

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,131 @@
...
+  private DateTimeFormatterUtil() {
+  }
+
+  /**
+   * Format the date using the date format String.
+   * @param dateExpr the date ExprValue of Date/Datetime/Timestamp/String type.
+   * @param formatExpr the format ExprValue of String type.
+   * @return Date formatted using format and returned as a String.
+   */
+  static ExprValue getFormattedDate(ExprValue dateExpr, ExprValue formatExpr) {
```
penghuo, line 87: Not sure the use case. Do you try to do something like this format.replace("%\\W", () -> replace).
If so, i recommended to consider the https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#appendReplacement(java.lang.StringBuffer,%20java.lang.String)
which provide the more clean code structure.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/datetime/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+    return SUFFIX_CONVERTER.getOrDefault(val % 10, SUFFIX_SPECIAL_TH);
+  }
+
+  /**
+   * Goes through format String and replaces any %x with 'x' where x is unmapped.
+   * @param format Incoming format String without mapping completed.
+   * @return Outgoing format String with mapping completed.
+   */
+  private static String literalReplace(String format) {
```
penghuo, line 121: What is the expected result if the user input the format string which we doesn't support?

lyndonb-bq: Do you mean a format we don't support, such as %z? 
According to the documentation for MySQL it should come out as just z, so using the DateTimeFormatter specification, to get a literal out, we wrap in ', so %z => 'z'.

penghuo: Got it, does it covered in the test cases also?

lyndonb-bq: Yes

penghuo (approved):
Thanks for the change!

Issue #772, opened 2020-10-08
SQL sum boolean value return 0


penghuo:
When SQL aggregate the boolean value, return 0 which is unexpected.
```
POST /_opendistro/_sql
{
  "query": """
    select SUM(CAST(Cancelled AS INT)) from kibana_sample_data_flights 
  """
}
{
  "schema": [{
    "name": "SUM(com.alibaba.druid.sql.ast.expr.SQLCastExpr@93fd2342)",
    "type": "double"
  }],
  "total": 1,
  "datarows": [[0]],
  "size": 1,
  "status": 200
}
```

penghuo:
Run against the explained DSL, return 0

```
GET /kibana_sample_data_flights/_search
{
  "from" : 0,
  "size" : 0,
  "_source" : {
    "includes" : [
      "SUM"
    ],
    "excludes" : [ ]
  },
  "aggregations" : {
    "SUM_0" : {
      "sum" : {
        "script" : {
          "source" : "def cast_1 = (doc['Cancelled'].value instanceof boolean) ? (doc['Cancelled'].value ? 1 : 0) : Double.parseDouble(doc['Cancelled'].value.toString()).intValue();",
          "lang" : "painless"
        }
      }
    }
  }
}
```

penghuo:
The painless script missing the return clause

```
GET /kibana_sample_data_flights/_search
{
  "from" : 0,
  "size" : 0,
  "_source" : {
    "includes" : [
      "SUM"
    ],
    "excludes" : [ ]
  },
  "aggregations" : {
    "SUM_0" : {
      "sum" : {
        "script" : {
          "source" : "def cast_1 = (doc['Cancelled'].value instanceof boolean) ? (doc['Cancelled'].value ? 1 : 0) : Double.parseDouble(doc['Cancelled'].value.toString()).intValue();return cast_1",
          "lang" : "painless"
        }
      }
    }
  }
}
```

dai-chen:
It seems return statement will be added to `MethodField` if it's on group by list (term and other aggregations), though no such logic if it's in aggregate function (metric aggregations).

Check another function and result is 0 too:

```
PUT _opendistro/_sql/settings
{
  "transient" : {
    "opendistro.sql.query.analysis.enabled" : "false"
  }
}

POST _opendistro/_sql
{
  "query": """
    SELECT SUM(ABS(FlightDelayMin)) FROM kibana_sample_data_flights 
  """
}
{
  "schema": [{
[...]

PR #773, opened 2020-10-08
Fix cast statement in aggregate function return 0

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/772

*Description of changes:* As per investigation in the issue, the root cause is return statement is missing in the painless script generated. Please the comment on the code changes for more details.

*TO-DO*: Other built-in functions also have this issue if we disable semantic analyzer (which disallow this) and do "SUM(ABS(a))". I tried a quick fix but it broke existing ITs. Here we only fix aggregate function on `CAST` statement. The issue should go away once new query engine enabled.


penghuo (approved):
Thanks for the fix.

chloe-zh (approved):

Issue #725, opened 2020-09-01
Describe statement doesn't work for index alias


dai-chen:
When `DESCRIBE` statement is used against an index alias, no result is returned.

```
PUT dropdown-region-marketplace
{
  "mappings": {
    "properties": {
      "balance": {
        "type": "float"
      },
      "age": {
        "type": "integer"
      }
    }
  }
}

# Work without issue
POST _opendistro/_sql
{
  "query": "DESCRIBE TABLES LIKE dropdown-region-marketplace"
}

# Empty result returned
PUT /dropdown-region-marketplace/_alias/ddregionmarketplace
POST _opendistro/_sql
{
  "query": "DESCRIBE TABLES LIKE ddregionmarketplace"
}
{
 ...
  "total": 0,
  "datarows": [],
  "size": 0,
  "status": 200
}
```

dai-chen:
I debugged code and found that we tried to match the index name returned in ES get mapping API and pattern in query. In the case of index alias, index name rather than alias is returned in get mapping API response. Thus it is unable to match the original pattern text in query.

PR #775, opened 2020-10-09
Support describe index alias

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/725

*Description of changes:* To understand the root cause and its fix, first let's see how we implemented SHOW/DESCRIBE:

1. Given SQL LIKE pattern (probably with wildcard `%` or `_`)
2. Convert to ES search pattern: replace `%` with ES wildcard `*`.
3. Get index metadata by ES pattern above: fetch all if including `_` because ES only supports `*`
4. Convert to regex pattern: replace `%` with `.*` and replace `_` with `.`
5. Filter the ES response by regex pattern above: because we may fetch all if pattern includes `_`, it's required to filter out the actual matched index name by regex again.

So the root cause is that alias doesn't necessarily match its index name so it's ruled out during regex matching. Ideally, we should fetch alias info and determine if LIKE pattern is an alias. However, this incurs extra ES metadata reading which may need additional permission and testing.

To quick fix the issue, changes in this PR is to skip the regex pattern matching if LIKE pattern doesn't include any wildcard. In this way, the matching process between an index name and its alias (regex pattern converted from pattern in LIKE) won't happen. Thus the index name will be present in the final result.

*Documentation*: Update existing doc with support for alias: https://github.com/dai-chen/sql/blob/support-describe-index-alias/docs/user/dql/metadata.rst#example-2-show-specific-index-information

*Testing*: Add UT and IT. Here is the 4 typical cases described in UT.

```
  /**
   * Case #1:
   * LIKE 'test%' is converted to:
   *  1. Regex pattern: test.*
   *  2. ES search pattern: test*
   * In this case, what ES returns is the final result.
   */
  ...

  /**
   * Case #2:
   * LIKE 'test_123' is converted to:
   *  1. Regex pattern: test.123
   *  2. ES search pattern: (all)
   * Because ES doesn't support single wildcard character, in this case, none is passed
   * as ES search pattern. So all index names are returned and need to be filtered by
   * regex pattern again.
   */
  ...

  /**
   * Case #3:
   * LIKE 'acc' has same regex and ES pattern.
   * In this case, only index name(s) aliased by 'acc' is returned.
   * So regex match is skipped to avoid wrong empty result.
   * The assumption here is ES won't return unrelated index names if
   * LIKE pattern doesn't include any wildcard.
   */
  ...

  /**
   * Case #4:
   * LIKE 'test.2020.10' has same regex pattern. Because it includes dot (wildcard),
   * ES search pattern is all.
   * In this case, all index names are returned. Because the pattern includes dot,
   * it's treated as regex and regex match won't be skipped.
   */
```

### DIFF
--- a/docs/user/dql/metadata.rst
+++ b/docs/user/dql/metadata.rst
@@ -55,7 +55,7 @@ Result set:
 Example 2: Show Specific Index Information
 ------------------------------------------
 
-Here is an example that searches metadata for index name prefixed by 'acc'
+Here is an example that searches metadata for index name prefixed by 'acc'. Besides index name and pattern with wildcard characters, searching by index alias is also supported. So you can ``SHOW`` or ``DESCRIBE`` an index alias which gives you same result as doing this with an index name.
 
 SQL query::
 
--- a/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/legacy/MetaDataQueriesIT.java
+++ b/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/legacy/MetaDataQueriesIT.java
@@ -29,7 +29,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import org.elasticsearch.client.Request;
-import org.elasticsearch.client.Response;
 import org.hamcrest.Description;
 import org.hamcrest.TypeSafeMatcher;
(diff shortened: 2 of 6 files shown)


penghuo (approved):
Thanks for the fix!

chloe-zh (approved):

PR #776, opened 2020-10-12
Change sql-release-workflow to use new staging bucket for artifacts

camerski:
*Description of changes:*
The infrastructure team is separating the production and staging locations into different AWS accounts. Plugins need to modify their workflows to publish to the new locations.

This PR changes the sql-release-workflow to add a build number and write the zip, deb, and rpm plugin artifacts to ```staging.artifacts.opendistroforelasticsearch.amazon.com/snapshots/elasticsearch-plugins/sql/```. The write to S3 currently fails because the secrets have not been updated, however a dummy run can be seen here: https://github.com/opendistro-for-elasticsearch/sql/actions/runs/296411387

This is the first change to the SQL plugin workflows. If this change is approved, I will update all the remaining workflows in a second PR. Secrets will be updated at the same time the PRs are merged.

See #777 for the second change.


abbashus (approved):
Looks good to me.

abbashus:
@camerski  I would suggest creating a PR with a fork repo workflow in future instead of directly creating a branch on OpenDistro SQL repo.

camerski:
> @camerski I would suggest creating a PR with a fork repo workflow in future instead of directly creating a branch on OpenDistro SQL repo.

Sure, no problem.

penghuo (approved):
Looks good to me!
Could you help us understand how the GITHUB_RUN_NUMBER help the release workflow?

camerski:
> Looks good to me!
> Could you help us understand how the GITHUB_RUN_NUMBER help the release workflow?

The motivation for this change is that with the current workflows it is possible for an artifact to be overwritten during or after an ODFE release. That makes it impossible to guarantee that the ODFE release actually contains the artifacts that were tested, because they could be overwritten between the testing and final build phases. Even after a release, the artifacts can be overwritten which means we cannot reliably reproduce an ODFE build.

This change guarantees that each build is unique, making it much easier to provide reproducible ODFE builds and guarantee that we're releasing the right artifacts.

Long term, we plan to decouple ODFE distributions from plugin releases and treat it more like a Linux distribution: An ODFE release will just cherry-pick whatever public version of your plugin you tell us to use. This will allow you to define your own release process and not be dependent on the infrastructure team.

joshuali925 (approved):

PR #777, opened 2020-10-12
Change release workflows to use new staging bucket

camerski:
Change sql-cli-, sql-jdbc-, sql-odbc-, and sql-workbench-release-workflow to use new staging bucket for artifacts

*Description of changes:*
This is a follow-up PR for https://github.com/opendistro-for-elasticsearch/sql/pull/776

The infrastructure team is separating the production and staging locations into different AWS accounts. Plugins need to modify their workflows to publish to the new locations.

This PR changes the rest of the release workflows, excluding sql-release-workflow, to add a build number and write the artifacts to staging.artifacts.opendistroforelasticsearch.amazon.com. The write to S3 currently fails because the secrets have not been updated.

This is the second change to the SQL plugin workflows. Secrets will be updated at the same time that this PR and #776 are merged.


joshuali925 (approved):

penghuo (approved):
LGTM

Issue #674, opened 2020-08-10
order by is not working as expected when more than one group by is used


dheerajravindranath:
`{"query": "select id, sum(marks) marks from mark_tables group by id order by marks desc"}` This is working as expected! But when we give more than one column in group by is not working fine! `{"query": "select id, name, sum(marks) marks from mark_tables group by id, name order by marks desc"}`. Please, give me some suggestions on this. Is there any workaround for this?

penghuo:
Thanks for reporting the issue.

Could you provide some test data for debuging? Have you try to explain the query by using the [explain](https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/interfaces/endpoint.rst#explain) API. It will help to know what is DSL of this SQL.

dai-chen:
It seems both Term Aggregation and Composite Aggregation API can only support sort inside bucket.

One workaround is to adjust the aggregation order based on order by clause:

```
SELECT OriginCountry, OriginCityName FROM flights GROUP BY OriginCountry, OriginCityName:
{
  "size" : 0,
  "_source" : false,
  "stored_fields" : "_none_",
  "aggregations" : {
    "groupby" : {
      "composite" : {
        "size" : 1000,
        "sources" : [
          {
            "da66945c" : {
              "terms" : {
                "field" : "OriginCountry.keyword",
                "missing_bucket" : true,
                "order" : "asc"
              }
            }
          },
          {
            "1af15150" : {
              "terms" : {
                "field" : "OriginCityName.keyword",
                "missing_bucket" : true,
                "order" : "asc"
              }
            }
          }
        ]
      }
    }
  }
}

SELECT OriginCountry, OriginCityName FROM flights GROUP BY OriginCountry, OriginCityName ORDER BY OriginCityName DESC
{
  "size" : 0,
  "_source" : false,
  "stored_fields" : "_none_",
  "aggregations" : {
    "groupby" : {
      "composite" : {
        "size" : 1000,
        "sources" : [
          {
            "1af15150" : {
              "terms" : {
                "field" : "OriginCityName.keyword",
                "missing_bucket" : true,
                "order" : "desc"
              }
            }
          },
          {
            "da66945c" : {
              "terms" : {
                "field" : "OriginCountry.keyword",
                "missing_bucket" : true,
                "order" : "asc"
              }
            }
          }
        ]
      }
    }
  }
}
```

dai-chen:
[...]

Issue #423, opened 2020-04-09
Query with order by aggregation function fails to return jdbc format output


zhongnansu:
#### Issue description
The query I tested is in [README: SQL Usage](https://github.com/opendistro-for-elasticsearch/sql/blob/master/README.md#sql-usage). It fails to return default jdbc format output.
```
SELECT COUNT(*),SUM(age),MIN(age) as m, MAX(age),AVG(age)
  FROM account
  GROUP BY gender
  HAVING m >= 20
  ORDER BY SUM(age), m DESC
```
#### Error message
```
{
  "error": {
    "reason": "Error occurred in Elasticsearch engine: all shards failed",
    "details": """Shard[0]: [account/LFh7b508QPeT_pKu8VjBEg] QueryShardException[No mapping found for [SUM(age)] in order to sort on]

For more details, please send request for Json format to see the raw response from elasticsearch engine.""",
    "type": "SearchPhaseExecutionException"
  },
  "status": 400
}
``` 
But it works fine after applying any of the following change:
1. Change to `format=json`
2. Give `SUM(age)` an alias, and `ORDER BY` its alias

#### Test data
[accounts.json](https://github.com/opendistro-for-elasticsearch/sql/blob/master/src/test/resources/doctest/testdata/accounts.json)

dai-chen:
The issue has been fixed in new query engine (which is for now experimental and disabled by default). You can enable it to preview by changing the plugin setting: https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst#opendistro-sql-engine-new-enabled. Thanks!

Issue #402, opened 2020-03-30
Group By Oder by * SQL ERROR


EveyWu:
SQL By ``bin/elasticsearch-sql-cli`` is correct

## Demo SQL of ReadMe.txt
```
	String aggregate_sql = "SELECT COUNT(*), SUM(age), MIN(age) as m, MAX(age), AVG(age)" +
			"  FROM bank" +
			"  GROUP BY gender" +
			"  HAVING m >= 20" +
			"  ORDER BY SUM(age), m DESC";
```
## ERROR
```
{
  "error" : {
    "reason" : "Error occurred in Elasticsearch engine: all shards failed",
    "details" : "Shard[0]: [bank/ngwB8uhERG-nAzDgIV7suA] QueryShardException[No mapping found for [SUM(age)] in order to sort on]\n\nFor more details, please send request for Json format to see the raw response from elasticsearch engine.",
    "type" : "SearchPhaseExecutionException"
  },
  "status" : 400
}
```

## ES ERROR LOG
```
java.lang.RuntimeException: Failed to execute phase [query], all shards failed; shardFailures {[yo7D69ncSpm5FCsX6xyjUQ][bank][0]: RemoteTransportException[[YIWENWU-MB1][127.0.0.1:9300][indices:data/read/search[phase/query]]]; nested: QueryShardException[No mapping found for [SUM(age)] in order to sort on]; }
	at com.amazon.opendistroforelasticsearch.sql.query.planner.core.BindingTupleQueryPlanner.execute(BindingTupleQueryPlanner.java:55) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.adapter.QueryPlanRequestBuilder.execute(QueryPlanRequestBuilder.java:37) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.QueryActionElasticExecutor.executeQueryPlanQueryAction(QueryActionElasticExecutor.java:65) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.executor.QueryActionElasticExecutor.executeAnyAction(QueryActionElasticExecutor.java:109) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
[...]

Issue #277, opened 2019-11-08
Aggregation with Sort doesn't work as expected


penghuo:
The SQL with aggreation and sort doesn't work as expected.

### SQL
```
    SELECT COUNT(*) as count
    FROM account 
    GROUP BY age 
    ORDER BY COUNT(*) 
```

### Actual Result
```
{
  "error": {
    "root_cause": [
      {
        "type": "query_shard_exception",
        "reason": "No mapping found for [COUNT(*)] in order to sort on",
        "index_uuid": "KRpeXu1zSleGXYXB7_hRNg",
        "index": "account"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "account",
        "node": "fD0Y7LubSCm1ff-cOeunQw",
        "reason": {
          "type": "query_shard_exception",
          "reason": "No mapping found for [COUNT(*)] in order to sort on",
          "index_uuid": "KRpeXu1zSleGXYXB7_hRNg",
          "index": "account"
        }
      }
    ]
  },
  "status": 400
}
```

dai-chen:
The issue has been fixed in new query engine (which is for now experimental and disabled by default). You can enable it to preview by changing the plugin setting: https://github.com/opendistro-for-elasticsearch/sql/blob/master/docs/user/admin/settings.rst#opendistro-sql-engine-new-enabled. Thanks!

PR #782, opened 2020-10-16
Support ORDER BY in new SQL engine

dai-chen:
*Issue #, if available:*

1. https://github.com/opendistro-for-elasticsearch/sql/issues/123: Order by aggregate function even if it's not in SELECT
2. https://github.com/opendistro-for-elasticsearch/sql/issues/674: Order by inner field of group by
3. https://github.com/opendistro-for-elasticsearch/sql/issues/277: Order by aggregate function
4. https://github.com/opendistro-for-elasticsearch/sql/issues/250: similar as above
5. https://github.com/opendistro-for-elasticsearch/sql/issues/423: similar as above
6. https://github.com/opendistro-for-elasticsearch/sql/issues/402: similar as above

*Description of changes:*

1. Relax `QualifiedName` to `UnresolvedExpression` in `Field` so we can support order by any expression for SQL.
2. Add `AstSortBuilder` that shares `replaceIfAliasOrOrdinal()` logic with `AstAggregationBuilder`.
3. Improve test framework to support comparison test for `ORDER BY` queries.

With these changes, we can support **ORDER BY field, scalar function, aggregate function, alias and ordinal**. Please see the documentation below for more details.

*Documentation*: https://github.com/dai-chen/sql/blob/support-order-by-in-new-sql-engine/docs/user/dql/basics.rst#example-3-ordering-by-aggregate-functions

*Testing*:

1. Add new UT and comparison test for basic cases
2. Add comparison tests for issues above for regression test
3. Bypass legacy `OrderIT` and `dateFunctionNameCaseInsensitiveTest()`: the former tests explain output and the latter tests `DATE_FORMAT` which has different signature in legacy and new engine.


sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/context/QuerySpecification.java
```diff
@@ -86,6 +99,52 @@ public void collect(QuerySpecificationContext query, String queryString) {
...
+          "Non-integer constant [%s] found in GROUP BY clause", expr));
+    }
+    return true;
+  }
+
+  private UnresolvedExpression getSelectItemByOrdinal(UnresolvedExpression expr) {
+    int ordinal = (Integer) ((Literal) expr).getValue();
+    if (ordinal <= 0 || ordinal > selectItems.size()) {
+      throw new SemanticCheckException(StringUtils.format(
+          "Group by ordinal [%d] is out of bound of select item list", ordinal));
```
penghuo, line 134: It is also could be sort by, right?

penghuo (approved):
Thanks for the change.

chloe-zh (approved):

PR #798, opened 2020-10-23
Support HAVING in new SQL engine

dai-chen:
*Issue #, if available:*

*Description of changes:*

1. Change SQL grammar and build `Filter` AST node to integrate with analyzer in core engine. (To replace alias in `HAVING` condition, add a new `AstHavingFilterBuilder` which extends `AstExpressionBuilder` so it doesn't need to replace and clone conditions later)
2. Optimize filter conditions in analyzer because now there may be aggregator in condition that needs to be replaced with a reference.

*Documentation*: https://github.com/dai-chen/sql/blob/support-having-in-new-sql-engine/docs/user/dql/aggregations.rst#having-clause

*Testing*: Add new UT and comparison IT.


penghuo (approved):
Thanks for the change!

chloe-zh (approved):

PR #804, opened 2020-10-28
Updated Mac version for GitHub action build

lyndonb-bq:
*Description of changes:*
Updated workflow to a different mac version because the new latest has a different C++ compiler that breaks the current build.

### DIFF
--- a/.github/workflows/sql-odbc-main.yml
+++ b/.github/workflows/sql-odbc-main.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:    
   build-mac:
-    runs-on: macos-latest
+    runs-on: macos-10.15
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


.github/workflows/sql-odbc-main.yml
```diff
@@ -11,7 +11,7 @@ env:
 
 jobs:    
   build-mac:
-    runs-on: macos-latest
+    runs-on: macos-10.15
```
rupal-bq, line 14: might need the same change in `sql-odbc-release-workflow.yml`

dai-chen (approved):
Thanks for the fix!

rupal-bq (approved):

chloe-zh (approved):

PR #818, opened 2020-11-11
Support CASE clause in new engine

dai-chen:
*Issue #, if available:* N/A

*Description of changes:* Add support for `CASE` clause in new SQL engine.

1. Add AST and expression node for `CASE` and `WHEN` clause.
2. Implement evaluation and type check. Because CASE clause is different from regular functions, it's not registered into function repository which usually performs the type check.

*TODO*: Case function for PPL is not done yet. Reference: https://docs.splunk.com/Documentation/Splunk/8.1.0/SearchReference/ConditionalFunctions

*Documentation*: https://github.com/dai-chen/sql/blob/support-case-statement-in-new-engine/docs/user/dql/functions.rst#case

*Testing*:

1. Add UT.
2. Add comparison test in https://github.com/dai-chen/sql/blob/support-case-statement-in-new-engine/integ-test/src/test/resources/correctness/expressions/conditionals.txt
3. Ignore and fix broken legacy test cases.


docs/user/dql/functions.rst
```diff
@@ -1832,3 +1832,82 @@ Description
...
+----
+
+Description
+>>>>>>>>>>>
+
+``CASE`` statement has two forms with slightly different syntax: Simple Case and Searched Case.
+
+Simple case syntax compares a case value with each compare value in ``WHEN`` clause and return its result if matched. Otherwise, result in ``ELSE`` clause is returned (or ``NULL`` if absent). Note that case value, compare value and result can be any expression::
+
+   CASE case_value
```
penghuo, line 1847: The case_value and compare_value could be expression, right? Could we refer the T-SQL definition https://docs.microsoft.com/en-us/sql/t-sql/language-elements/case-transact-sql?view=sql-server-ver15

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/ExpressionAnalyzer.java
```diff
@@ -163,6 +169,43 @@ public Expression visitCompare(Compare node, AnalysisContext context) {
...
+      if (node.getCaseValue() == null) {
+        whens.add((WhenClause) analyze(when, context));
+      } else {
+        // Merge case value and condition (compare value) into a single equal condition
+        whens.add((WhenClause) analyze(
+            new When(
+                new Function("=", Arrays.asList(node.getCaseValue(), when.getCondition())),
+                when.getResult()
+            ), context));
+      }
```
penghuo, line 185: Can we merge two different case...when logic in the AstBuilder?

penghuo (approved):
Thanks for the change!

chloe-zh (approved):

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/ExpressionAnalyzer.java
dai-chen: yeah, we can, I was thinking make AST close to original form and do this equivalent transform and type check in analyzer together.

docs/user/dql/functions.rst
dai-chen: sure, will do. thanks!

Issue #375, opened 2020-03-05
Aggregation subquery in FROM clause runs into wrong code path


dai-chen:
Subquery with aggregate function or `GROUP BY` is being handled by recently added post processing logic unexpectedly. This only happens for default JDBC format:

Example copied from `SubqueryIT`:

```
POST _opendistro/_sql
{
  "query": """
    SELECT t.TEMP1 as count, t.TEMP2 as balance
    FROM (SELECT COUNT(*) as TEMP1, SUM(balance) as TEMP2
          FROM accounts) t
  """
}

{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "unsupported expr: t.TEMP1",
    "type": "RuntimeException"
  },
  "status": 503
}
```

Exception in ES log:

```
[2020-03-05T13:31:36,187][ERROR][c.a.o.s.p.RestSqlAction  ] [186590df9563.ant.amazon.com] 4e34c9da-f63f-4de8-bcd7-ffc22641d630 Server side error during query execution
java.lang.RuntimeException: unsupported expr: t.TEMP1
	at com.amazon.opendistroforelasticsearch.sql.query.planner.converter.SQLExprToExpressionConverter.convert(SQLExprToExpressionConverter.java:96) ~[opendistro_sql-1.4.0.0.jar:1.4.0.0]
	at com.amazon.opendistroforelasticsearch.sql.query.planner.converter.SQLAggregationParser.parseExprInSelectList(SQLAggregationParser.java:155) ~[opendistro_sql-1.4.0.0.jar:1.4.0.0]
	at com.amazon.opendistroforelasticsearch.sql.query.planner.converter.SQLAggregationParser.parse(SQLAggregationParser.java:66) ~[opendistro_sql-1.4.0.0.jar:1.4.0.0]
	at com.amazon.opendistroforelasticsearch.sql.query.planner.converter.SQLToOperatorConverter.visit(SQLToOperatorConverter.java:61) ~[opendistro_sql-1.4.0.0.jar:1.4.0.0]
	at com.alibaba.druid.sql.dialect.mysql.ast.statement.MySqlSelectQueryBlock.accept0(MySqlSelectQueryBlock.java:255) ~[druid-1.0.15.jar:1.0.15]
	at com.alibaba.druid.sql.dialect.mysql.ast.statement.MySqlSelectQueryBlock.accept0(MySqlSelectQueryBlock.java:246) ~[druid-1.0.15.jar:1.0.15]
[...]

PR #822, opened 2020-11-13
Support subquery in FROM clause in new engine

chloe-zh:
*Issue #, if available:*
#375 

*Description of changes:*

- Added subquery in FROM as table source in grammar and parser
- Created new tree node `RelationSubquery` in AST standing for subquery as table
- Added test cases
- Skipped a broken test case in `SubqueryIT.java` caused by different schema in new engine


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/tree/RelationSubquery.java
```diff
@@ -0,0 +1,64 @@
...
+  private String alias;
+
+  /**
+   * Take subquery alias as table name.
+   */
+  public String getAliasAsTableName() {
+    if (alias == null) {
+      throw new SemanticCheckException(
+          String.format(Locale.ROOT, "missing table alias for subquery %s", toString()));
+    }
```
dai-chen, line 46: Since alias is not optional in ANTLR grammar, is this check necessary?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/Analyzer.java
```diff
@@ -118,6 +119,17 @@ public LogicalPlan visitRelation(Relation node, AnalysisContext context) {
     return new LogicalRelation(node.getTableName());
   }
 
+  @Override
+  public LogicalPlan visitRelationSubquery(RelationSubquery node, AnalysisContext context) {
+    context.push();
```
dai-chen, line 124: Since we can support subquery now, do we need to pop up context somewhere?

sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/context/QuerySpecification.java
```diff
@@ -170,9 +171,10 @@ public QuerySpecificationCollector(String queryString) {
     }
 
     @Override
-    public Void visitQuerySpecification(QuerySpecificationContext ctx) {
-      // TODO: avoid collect sub-query
-      return super.visitQuerySpecification(ctx);
+    public Void visitSubqueryAsRelation(SubqueryAsRelationContext ctx) {
+      // skip collecting subquery for current layer
+      visit(ctx.alias());
```
dai-chen, line 176: Is this necessary?

dai-chen:
Can you add examples for new features that we can support now? https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/dql/complex.rst#example-2-subquery-in-from-clause

For example, two levels of subquery in FROM, a subquery with GROUP BY etc.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/tree/RelationSubquery.java
penghuo: +1

chloe-zh: forgot to remove this after changing the antlr grammar. Will remove it, thanks!

dai-chen (approved):
Thanks for the changes! Btw, in the doc, you can doctest, no need to follow the old format.

penghuo (approved):

chloe-zh:
> Thanks for the changes! Btw, in the doc, you can doctest, no need to follow the old format.

Thanks, done!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/analysis/Analyzer.java
dai-chen: Just took another look, so context is pushed first when visit RelationSubquery, should we do analyzer recursion before this? In this way, symbol defined by outer won't be accessible to inner, ex. select ... from (select ... from B) as A, is A accessible when visit subquery B?

PR #836, opened 2020-11-19
Seperate the logical plan optimization rule from core to storage engine

penghuo:
## Why making the change
The PR seperate the logical plan optimization into two categories.
1. Pure relation operator optimization which independent of storage engine.
2. Storage engine related optimization. e.g. In Elasticsearch, the aggregation operator could merge with relation operator as IndexAgg operator which is sepeific to Elasticsearch.

## Description of changes
Add the new optimize function to Table, the storage could override this interface to perform customized optimization. By default, there is no optimization.
```
  default LogicalPlan optimize(LogicalPlan plan) {
    return plan;
  }
```


chloe-zh (approved):

dai-chen (approved):
Thanks for the changes!

PR #840, opened 2020-11-20
Migrate Query Workbench to 7.10

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Migrate query workbench to 7.10. Tested by running both `SQL` and `PPL` queries and testing features such as `Explain`, `Download`, and sorting on columns. 

Related PR to update workflows: #844 


workbench/kibana.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "opendistroQueryWorkbench",
```
joshuali925, line 2: in odfe 1.12 workbench plugin changed to `opendistroQueryWorkbenchKibana`, could you update here?

joshuali925 (commented):
Github action failed, could you update node version?
Also artifact file name changed in 7.10, needs to rename it in github actions. You can use reporting as a reference https://github.com/opendistro-for-elasticsearch/kibana-reports/blob/dev/.github/workflows/kibana-reports-test-and-build-workflow.yml#L49

workbench/kibana.json
davidcui-amzn: Yes, will do

chloe-zh:
`error opendistro-query-workbench@1.12.0.0: The engine "node" is incompatible with this module. Expected version "10.22.1". Got "10.22.0"`
Looks like the node version does not match in the workbench workflows, could you also change the node version in `.github/workflows/sql-test-and-build-workflow.yml` and `.github/workflows/sql-workbench-release-workflow.yml`?

chloe-zh (approved):
LGTM, thanks for the changes!

joshuali925 (approved):
Thanks

PR #843, opened 2020-11-20
Support NULLS FIRST/LAST in new engine

dai-chen:
*Issue #, if available:* N/A

*Description of changes:*

1. Change SQL grammar to add `NULLS FIRST` and `NULLS LAST` clause.
2. Build AST Sort node with sort options given.
3. Analyze sort options by determining default nulls order if `NULLS` clause absent.
4. Change comparison test data and framework to support nulls.

*TODO*: Add support in window function `OVER` clause. Because AST window function is unresolved expression instead of operator, regular `ORDER BY` and that in over clause cannot share same logic.

*Documentation*:

1. Add doctest: https://github.com/dai-chen/sql/blob/support-order-by-null-order-2/docs/user/dql/basics.rst#example-2-specifying-order-for-null
2. Update README with all features only available in new engine to avoid user confusion: https://github.com/dai-chen/sql/tree/support-order-by-null-order-2#experimental

*Testing*:

1. Add UT and comparison test
2. Remove one comparison test query because it is ordering on SUM(FlightDelayMin) which is 0 for many rows and the order is not determined.
3. Change alias names in `can_build_from_subquery()` because new SQL doesn't support keyword as alias name.


penghuo (approved):
Thanks for the change.

chloe-zh (approved):

PR #846, opened 2020-11-23
Fix bug of nested field format issue in JDBC response

penghuo:
*Description of changes:*
1.Fix the bug of nested filed output.
2.Fix the bug for Unit Test.


dai-chen (approved):
Thanks for the fix!

chloe-zh (approved):

PR #848, opened 2020-11-23
Sort field push down

penghuo:
## New added features.
1. Add the fitler-sort push down optimizaiton rule in core engine. Rule: PushFilterUnderSort
`e.g. Filter - Sort - AnyChild will be optimized as Sort - Filter - AnyChild`
2. Add the sort operator push down on Elasticseach storage engine. Rule: MergeSortAndIndexScan and MergeSortAndIndexRelation
`e.g. Sort - Relation will be optimized as IndexScan(sort)`

## Limitation
Currently, the storage engine layer optimzation depend on the Elasticsearch DSL to push down to sort expression. But there are some limitations of Elasticsearch DSL.
* Regardless the sort order (ASC/DESC), the doc has the field with NULL value always come before the doc missing the field.
* The script based sort can’t handle MISSING/NULL value
* The script based sort doesn’t support MISSING/NULL first/last paramater.

Thus, currently, the Sort Operator will push down to Elasticsearch query only when the sort by list include fields only. if the Sort operator include sort expression, the optimization will do nothing, it will still fall back to default memory based implementation. 
`e.g. source=index | sort abs(a)` the expression `abs(a)` can't be push down to Elasticsearch DSL.

## Changes
1. Remove the count from PPL sort command.
2. Remove the count from LogicalSort/PhysicalSort

## Todo
1. Aggregation with Sort.


dai-chen (approved):
LGTM. Thanks!

chloe-zh (approved):

Issue #852, opened 2020-11-24
New engine: Sort order keyword ASC/DESC is case sensitive


chloe-zh:
Sort order ASC/DESC works well when the keyword ASC/DESC is in upper case, but it is not recognized if using lower case, for example:

```
POST _opendistro/_sql
{
  "query": "select age from accounts order by age desc"
}

{
  "error": {
    "reason": "Invalid SQL query",
    "details": "No enum constant com.amazon.opendistroforelasticsearch.sql.ast.tree.Sort.SortOrder.desc",
    "type": "IllegalArgumentException"
  },
  "status": 400
}
```

expected output:
```
{
  "schema": [
    {
      "name": "age",
      "type": "long"
    }
  ],
  "datarows": [
    [
      36
    ],
    [
      33
    ],
    [
      32
    ],
    [
      28
    ]
  ],
  "total": 4,
  "size": 4
}
```

PR #853, opened 2020-11-24
Fix issue: sort order keyword is case sensitive

chloe-zh:
*Issue #, if available:*
#852 

*Description of changes:*
- Replaced the sort order keyword to upper case when collecting the query information in `QuerySpecification`
- Added tests


penghuo (approved):

dai-chen (approved):
Thanks for the fix!

PR #854, opened 2020-11-25
Add JDBC formatter in new SQL engine

dai-chen:
*Issue #, if available:* N/A

*Description of changes:* Add a new JDBC formatter to format new engine response as before. This is intend to avoid impact on user after new SQL engine enabled in future. The main differences between this and default JSON formatter include:

1. `alias` field in schema
2. `status` field is added
3. Data type: convert to legacy type name for compatibility, including ES_TEXT, ES_TEXT_KEYWORD, STRUCT, ARRAY.
4. Error response: follow exactly same format as before. Treat syntax and query engine exception as client error.

**Note that this is only for SQL and will take effect without `format` parameter or with `format=jdbc`** New PPL doesn't need this.

*TODO*:

1. Add `version` field and do version check on client side: simply adding new field causes JDBC driver broken, so need to change both sides together later.
2. For now `total` field is always same as `size`: need to use total number of hits as before

*Testing*:

1. Add new `JdbcFormatIT`
2. Add back legacy IT removed earlier due to response format issue: `AggregationExpressionIT` and `QueryIT`


protocol/src/test/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/JdbcResponseFormatterTest.java
```diff
@@ -0,0 +1,121 @@
...
+        "{"
+            + "\"schema\":["
+            + "{\"name\":\"name\",\"alias\":\"name\",\"type\":\"keyword\"},"
+            + "{\"name\":\"address1\",\"alias\":\"address1\",\"type\":\"text\"},"
+            + "{\"name\":\"address2\",\"alias\":\"address2\",\"type\":\"text\"},"
+            + "{\"name\":\"location\",\"alias\":\"location\",\"type\":\"object\"},"
+            + "{\"name\":\"employer\",\"alias\":\"employer\",\"type\":\"nested\"},"
+            + "{\"name\":\"age\",\"alias\":\"age\",\"type\":\"integer\"}"
+            + "],"
+            + "\"datarows\":[[\"John\",20]],"
```
penghuo, line 66: How to handle the missing value?

dai-chen: I missed other columns here. Will add and check other formatter's UT. Thanks!

chloe-zh (approved):

protocol/src/test/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/JdbcResponseFormatterTest.java
dai-chen: Fixed this mismatch and added another case for missing and null value.

penghuo (approved):

PR #855, opened 2020-11-25
For ODFE 1.12 change position for sql workbench plugin

zhyuanqi:
*Issue #, if available:*

*Description of changes:*
Change position of SQL workbench plugin in side bar. 
The order is designed by UX designer and in step of 1000(https://github.com/elastic/kibana/blob/master/docs/development/core/public/kibana-plugin-core-public.appcategory.md). 
The sql workbench is the first in the row. So I put 1000 for the order number. 
image (4)


joshuali925 (approved):
Thanks

seraphjiang (approved):
LGTM

PR #866, opened 2020-11-30
Added keywords option as alias identifier in SQL parser

chloe-zh:
*Issue #, if available:*

*Description of changes:*
- Added keywords option as alias in SQL parser
- Added tests


sql/src/main/antlr/OpenDistroSQLIdentifierParser.g4
```diff
@@ -40,6 +40,7 @@ columnName
 
 alias
     : ident
+    | keywordsCanBeId
```
dai-chen, line 43: Should we move this `keywordCanBeId` both here and in `qualifiedName` to under `ident`? Otherwise I recall keyword in qualified name with dots will still fail, ex. SELECT a.count ...

chloe-zh: Good point! Will change

dai-chen (approved):
Could you add FIRST and LAST to `keywordsCanBeId ` too? Thanks for the changes!

penghuo (approved):

Issue #148, opened 2019-08-09
CSV and JDBC export are 2 different execution paths


galkk:
CSV and JDBC result formatting have 2 parallel codebases, so fix in one wouldn't work with another and one needs to always remember to double check. Example: CSV formatter supported all aggregations (Percentile, Numeric SimpleValue, Numeric Multiple), and JDBC supported only one: Numeric SimpleValue

Recommendation: the CSV formatter should use results of JDBC formatter and do simple transformation of results. The logic, that is currently not presented in JDBC formatter must be brought there

CSV formatting: CSVResultsExtractor.java
JDBC/RAW: SelectResultSet.java

abbashus:
+1.

dai-chen:
This has been resolved in new query engine.

Issue #117, opened 2019-07-16
JDBC and CSV format do not support column aliases


galkk:
The output for both JDBC and CSV doesn't respect alias in query

```
POST _opendistro/_sql?format=jdbc
{
  "query": "SELECT date_format(utc_time, 'YYYY') as year, date_format(utc_time,'MM') as month, count(*) FROM kibana_sample_data_logs GROUP BY date_format(utc_time, 'YYYY'), date_format(utc_time, 'MM')"
}


{
  "schema": [
    {
      "name": "date_format_1257426103",
      "type": "text"
    },
    {
      "name": "date_format_1868425714",
      "type": "text"
    },
    {
      "name": "COUNT(*)",
      "type": "long"
    }
  ],
  "total": 3,
  "datarows": [
    [
      "2018",
      "08",
      7132
    ],
    [
      "2018",
      "09",
      4552
    ],
    [
      "2018",
      "07",
      2321
    ]
  ],
  "size": 3,
  "status": 200
}
```

```
POST _opendistro/_sql?format=csv
{
  "query": "SELECT date_format(utc_time, 'YYYY') as year, date_format(utc_time,'MM') as month, count(*) FROM kibana_sample_data_logs GROUP BY date_format(utc_time, 'YYYY'), date_format(utc_time, 'MM')"
}

date_format_1153263583,date_format_184649509,COUNT(*)
2018,08,7132.0
2018,09,4552.0
2018,07,2321.0
```

dai-chen:
Same situation as https://github.com/opendistro-for-elasticsearch/sql/issues/100#issuecomment-571699967

Issue #100, opened 2019-06-28
Wrong order of returned fields when function is used in select list


galkk:
When using function in select list, the order of fields in returned csv doesn't match the order in the query

```
POST _opendistro/_sql?format=csv
{
  "query": "SELECT ip, utc_time, date_format(utc_time, 'dd-MM-YYYY') date from kibana_sample_data_logs"
}
```

Expected field order: ip, utc_time, date
Actual field order: date,ip,utc_time
22-07-2018,157.4.76.38,2018-07-22T12:23:52.803Z

dai-chen:
JDBC format seems able to handle this case. So this should be fixed once we merge the different code paths to one in https://github.com/opendistro-for-elasticsearch/sql/issues/148.

Issue #546, opened 2020-07-01
Improve raw formatter with more options


dai-chen:
Recently we secured our CSV formatter by sanitizing special characters for Microsoft Excel: https://github.com/opendistro-for-elasticsearch/sql/issues/449. One of the side effect is that cell with potential executable code will be prefixed by a single quote. As our CSV format targets Microsoft Excel, we want to improve our raw formatter so user can have more flexibility to control how their data exported to a file.

Here is an example command in MySQL exporter which allows user to choose different field and line separator:

```
SELECT * FROM Test
INTO OUTFILE '/tmp/userdata.csv'
FIELDS TERMINATED BY ','
LINES TERMINATED BY '\n';
```

dai-chen:
A new param `santize` is added to CSV formatter in new engine, so this improvement is not necessary: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/interfaces/protocol.rst#csv-format

PR #870, opened 2020-12-01
Add CSV formatter in new engine

chloe-zh:
*Issue #, if available:*
#148 #117 #100 

*Description of changes:*

- Added csv formatter `CsvResponseFormatter` in new engine. The formatter formats the `QueryResult` to csv format and by default it sanitizes the result by (1) inserting single-quote in front of header/data if it starts with special characters including `+`, `-`, `=` and `@`; (2) quoting the header/data if one or more commas (`,`) in the header/data. The sanitizing can be skipped by adding an `sanitize` param and set it to `false` value in the request. 

- User manual updates: https://github.com/chloe-zh/sql/blob/csv-formatter/docs/user/interfaces/protocol.rst

- Added tests for the formatter; ignored a couple of legacy tests that are not applicable for the new engine


dai-chen:
Could you add the following related issue to the description? I think we should be able to close them after this PR merged.

1. https://github.com/opendistro-for-elasticsearch/sql/issues/148
2. https://github.com/opendistro-for-elasticsearch/sql/issues/117
3. https://github.com/opendistro-for-elasticsearch/sql/issues/100
4. https://github.com/opendistro-for-elasticsearch/sql/issues/546

chloe-zh:
> Could you add the following related issue to the description? I think we should be able to close them after this PR merged.
> 
> 1. #148
> 2. #117
> 3. #100
> 4. #546

Added but the last issue #546 because this PR didn't enable custom separators yet

docs/user/interfaces/protocol.rst
```diff
@@ -299,6 +299,39 @@ Result set::
...
+
+Result set::
+
+    '+firstname,'=lastname,address
+    'Hattie,'@Bond,"671 Bristol Street, Dente, TN"
+
+
+To escape the sanitize and keep the csv result as the original, you can add the escape param and set it to true value. For example::
+
+	>> curl -H 'Content-Type: application/json' -X POST localhost:9200/_opendistro/_sql?format=csv&escape=true -d '{
```
dai-chen, line 327: I feel `escape` is not very clear. Could we use a different name? ex. `sanitize=false` means skip sanitizing operation.

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/CsvResponseFormatter.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.stream.Collectors;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Singular;
+
+@RequiredArgsConstructor
+public class CsvResponseFormatter implements ResponseFormatter<QueryResult> {
+  private static final String INLINE_SEPARATOR = ",";
+  private static final String INTERLINE_SEPARATOR = "\n";
```
dai-chen, line 34: Could you double check in CSV should we use JDK's platform independent separator or not? https://stackoverflow.com/questions/207947/how-do-i-get-a-platform-dependent-new-line-character

dai-chen:
> > Could you add the following related issue to the description? I think we should be able to close them after this PR merged.
> > 
> > 1. #148
> > 2. #117
> > 3. #100
> > 4. #546
> 
> Added but the last issue #546 because this PR didn't enable custom separators yet

Got it, you can close that one once this merged. Thanks!

legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/plugin/RestSQLQueryAction.java
```diff
@@ -109,6 +121,13 @@ public RestChannelConsumer prepareRequest(SQLQueryRequest request, NodeClient no
       return NOT_SUPPORTED_YET;
     }
 
+    Format format = request.format();
+    if (format.equals(Format.CSV)) {
+      formatter = new CsvResponseFormatter(request.escape());
+    } else {
+      formatter = new JdbcResponseFormatter(PRETTY);
+    }
```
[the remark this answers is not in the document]
chloe-zh: Ok, but in this case I extracted another property `request: SqlRequest` in `RestSQLQueryAction` to pass the sql request properties (including the format and sanitize option) to the listener

legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/plugin/RestSqlAction.java
```diff
@@ -154,6 +154,10 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
                                                                     sqlRequest.getSql(),
                                                                     request.path(),
                                                                     format.getFormatName());
+                // set escape option if csv format
+                if (format.equals(Format.CSV)) {
+                    newSqlRequest.escape(SqlRequestParam.getEscapeOption(request.params()));
+                }
```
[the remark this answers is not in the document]
chloe-zh: Sure, will do

docs/user/interfaces/protocol.rst
chloe-zh: Sounds making more sense, will do

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/CsvResponseFormatter.java
chloe-zh: Good point, will do!

legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/plugin/RestSQLQueryAction.java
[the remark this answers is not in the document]
dai-chen: I just realized this `formatter` is actually only required by `createQueryResponseListener` instead of both execute and explain. In this case, how about we just pass `request` and create specific formatter in `createQueryResponseListener`?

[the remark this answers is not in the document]
chloe-zh: Yes that's also what I'm thinking about, the changes are done

dai-chen (approved):
Thanks for the changes!

docs/user/interfaces/protocol.rst
```diff
@@ -299,6 +299,39 @@ Result set::
 	Amber,Duke,32
 	Dale,Adams,33
 	Hattie,Bond,36
+
+
+The formatter sanitizes the csv result with the following rules:
+1. If a header cell or data cell is starting with special character including '+', '-', '=' , '@', the sanitizer will insert a single-quote at the start of the cell.
```
penghuo, line 305: It should in seperate paragraph, right?

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/CsvResponseFormatter.java
```diff
@@ -0,0 +1,137 @@
...
+  private static final Set<String> SENSITIVE_CHAR = ImmutableSet.of("=", "+", "-", "@");
+
+  private final boolean sanitize;
+
+  public CsvResponseFormatter() {
+    this.sanitize = true;
+  }
+
+  @Override
+  public String format(QueryResult response) {
```
penghuo, line 44: In case we already have CsvResult abstraction. Could we make the  code like following
```
CsvResult result = new CsvResult(response)
return result.getCsv()
```

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/CsvResponseFormatter.java
```diff
@@ -0,0 +1,137 @@
...
+    CsvResult result = buildCsvResult(response);
+    String headers = String.join(INLINE_SEPARATOR, result.getHeaders());
+    ImmutableList.Builder<String> dataLines = new ImmutableList.Builder<>();
+    result.getData().forEach(line -> dataLines.add(String.join(INLINE_SEPARATOR, line)));
+    return String.join(
+        INTERLINE_SEPARATOR, headers, String.join(INTERLINE_SEPARATOR, dataLines.build()));
```
penghuo, line 50: could we simply the logic as 
```
Stream<String> headerAndData = (?);
headerAndData.stream().collect(Collectors.joining(INTERLINE_SEPARATOR))
```

penghuo (approved):

PR #886, opened 2020-12-03
Fix workbench bugs from plugin platform upgrade

joshuali925:
*Issue #, if available:*

*Description of changes:*
- update download json api to be `/api/sql_console/sqljson` instead of `/api/sql_console/queryjson`
- update cypress tests
- changed POST body to ``{ body: `{ "query": "${query}" }` }`` from `{ query: query }` so we no longer need `convertQueryToString` (ref: https://github.com/opendistro-for-elasticsearch/sql/pull/812#discussion_r521506109)
- use request instead of actual query as parameter when calling `client.asScoped(request)`, so it works with security plugin
- fix margin of results tabs in 7.10

image


davidcui-amzn (approved):
LGTM, thanks for the fix

chloe-zh (approved):
LGTM, thanks for the changes!

Issue #877, opened 2020-12-03
Fix CASE clause pushdown issue


dai-chen:
The new `CaseClause` was added as a separate expression class other than `FunctionExpression`. This caused it cannot be handled by optimizer.

Issue #582, opened 2020-07-17
case statement is not working properly


dheerajravindranath:
Elasticsearch Verison: 7.4

{
    "query": "select distinct(case when event='test' then user_id end) as testing from events_testing where event='test' and team_id=xxxxx"
}

{
  "error": {
    "reason": "Invalid SQL query",
    "details": "",
    "type": "NullPointerException"
  },
  "status": 400
}

The same query is working fine in other DB. Could someone help me to resolve this?

dai-chen:
@dheerajravindranath Thanks for reporting the issue! I think we only support simple `DISTINCT(* )` and `DISTINCT(field)`. Will let you know if we can fix this or workaround found for your case.

dheerajravindranath:
Thanks, @dai-chen It would be great if we get this resolved asap!

dai-chen:
This issue has been resolved in our new SQL engine. Please find more details here: https://github.com/opendistro-for-elasticsearch/sql#experimental.

Test queries verified in PR above: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/integ-test/src/test/resources/correctness/bugfixes/582.txt

```
SELECT DISTINCT CASE WHEN Carrier = 'ES-Air' Then FlightDelayMin END AS delay FROM kibana_sample_data_flights
SELECT DISTINCT CASE WHEN FlightDelay = true Then Carrier END AS name FROM kibana_sample_data_flights
SELECT DISTINCT CASE WHEN FlightDelay = true Then Carrier WHEN FlightDelay = false THEN NULL END AS name FROM kibana_sample_data_flights
```

PR #895, opened 2020-12-04
Fix CASE clause pushdown issue

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/877, https://github.com/opendistro-for-elasticsearch/sql/issues/582

*Description of changes:* Because of the difference between CASE clause and normal scalar function, `CaseClause` implemented `Expression` interface rather than `FunctionExpression`. This caused CASE clause pushdown issue since both filter and aggregation optimizer handles FunctionExpression only. 

This PR is to:

1. Make `CaseClause` and `WhenClause` extend FunctionExpression to support pushdown optimization.
2. Ignore NULL in `THEN` and `ELSE` when type check for CASE clause, ex. CASE WHEN a = 1 THEN 'test' ELSE NULL is valid.
3. Fix NPE issue in `Literal.toString()` when the literal represent a null literal with null `value`.

*Testing:* add UT and new comparison test for WHERE and GROUP BY CASE clause.


penghuo (approved):
Thanks for the chagne.

chloe-zh (approved):

PR #896, opened 2020-12-07
Update utils.ts

SKlimov:
*Issue #, if available:*

*Description of changes:*
Changed encode function for saving to file.
encodeURI doesn't work very well in current versions of Chrome. This function won't escape symbols like "#".
So, all requested data that should be saved in a file will be truncated by the first # symbol in Chrome browser, in Safari this feature works fine.


chloe-zh (approved):
LGTM, thanks for the fix!

anirudha (approved):

Issue #867, opened 2020-11-30
group by 1 isn't working in new engine


rupal-bq:
Query:
```
 SELECT DATE_ADD(`calcs`.`date2`, INTERVAL 1 MONTH) AS `TEMP(Test)(2799254343)(0)` FROM `calcs` GROUP BY 1 
```
Error:
```
{
    "error": {
        "type": "IllegalStateException",
        "reason": "Unsupported type DATETIME to construct expression value from object for field: DATE_ADD(calcs.date2, Interval(value=1, unit=MONTH)), value: 1972-08-12.",
        "details": "Unsupported type DATETIME to construct expression value from object for field: DATE_ADD(calcs.date2, Interval(value=1, unit=MONTH)), value: 1972-08-12."
    },
    "status": 503
}
```
Was able to get a successful response after removing `group by 1` from the query.

Issue #858, opened 2020-11-25
functions in group by clause aren't supported 


rupal-bq:
The query

```
SELECT DATE(`calcs`.`date2`) AS `TEMP(Test)(3817907367)(0)` FROM `calcs` GROUP BY DATE(`calcs`.`date2`)
```
returns the following error while using new engine.
```
{
    "error": {
        "type": "IllegalStateException",
        "reason": "Unsupported type DATE to construct expression value from object for field: DATE(calcs.date2), value: 1972-07-12.",
        "details": "Unsupported type DATE to construct expression value from object for field: DATE(calcs.date2), value: 1972-07-12."
    },
    "status": 503
}
```

After modifying the query to 
```
 SELECT DATE(`calcs`.`date2`) AS `TEMP(Test)(3817907367)(0)` FROM `calcs` GROUP BY `calcs`.`date2`  
```
got the successful response as below.
```
{
    "schema": [
        {
            "name": "TEMP(Test)(3817907367)(0)",
            "alias": "TEMP(Test)(3817907367)(0)",
            "type": "date"
        }
    ],
    "datarows": [
        [
            "1972-07-12"
        ],
        [
            "1974-03-17"
        ],
        [
            "1974-05-03"
        ],
        [
            "1976-09-09"
        ],
        [
            "1977-02-08"
        ],
        [
            "1977-04-20"
        ],
        [
            "1980-07-26"
        ],
        [
            "1980-11-07"
        ],
        [
            "1988-01-05"
        ],
        [
            "1994-04-20"
        ],
        [
            "1995-06-04"
        ],
        [
            "1995-09-03"
        ],
        [
            "1997-05-30"
        ],
        [
            "1997-09-19"
        ],
        [
            "1998-08-12"
        ],
        [
            "2001-02-04"
        ],
        [
            "2002-04-27"
        ]
    ],
    "total": 17,
    "size": 17,
    "status": 200
}
```

PR #898, opened 2020-12-08
Fix for ExprValueFactory construct issue

rupal-bq:
*Issue #858, #867*

*Description of changes:*
- added date, time, and datetime in ExprValueFactory
- added unit tests 
- added manual integration tests (SQLite returned the wrong type in a schema and the comparison test failed).


dai-chen:
Is https://github.com/opendistro-for-elasticsearch/sql/issues/867 same issue?

rupal-bq:
> Is #867 same issue?

Thanks for noticing, it's the same issue. I will add a test for that case as well.

chloe-zh:
Could you also add test cases in comparison test?

chloe-zh (approved):

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/data/value/ElasticsearchExprValueFactory.java
```diff
@@ -177,14 +183,22 @@ public ExprValue construct(String field, Object value) {
       return constructString((String) value);
     } else if (type.equals(BOOLEAN)) {
       return constructBoolean((Boolean) value);
-    } else if (type.equals(TIMESTAMP)) {
+    } else if (type.equals(TIMESTAMP) || type.equals(DATE) || type.equals(TIME)
```
penghuo, line 186: Quesiton,  In which case the field will in type of DATA/TIME?
The main purpose of ExprValueFactory is extract data from Elasticsearch response. Elasticsearch only support date type and it is mapped to TIMESTAMP.

penghuo: Never mind, this is the aggregation result. I see the problem.

penghuo (approved):
Thanks for the change.

PR #906, opened 2020-12-10
Fix Round fix issue when input is negative and end with .5

harold-wang:
*Issue #908 
Round function did not round negative number correctly in case input is negative and end with .5. 
issue: Round(-3.5) = 3.0 instead of 4.0

*Description of changes:*
Change Round(double) to Round(double, 0)

Test: IT added and 100% pass


penghuo (approved):
Thanks for the change.

penghuo:
Is it related to any open issue? if so, please also link it.

dai-chen (approved):
Thanks for the fix!

chloe-zh (approved):

harold-wang:
> Is it related to any open issue? if so, please also link it.

Linked, thanks

Issue #404, opened 2020-03-31
SHOW/DESCRIBE explain throws exception


dai-chen:
Explain `SHOW/DESCRIBE` throws exception. It is supposed to either disable explain for `SHOW/DESCRIBE` or return friendly message to user.

```
POST _opendistro/_sql/_explain
{  
  "query": "SHOW TABLES LIKE %"
}

{
  "error" : {
    "root_cause" : [
      {
        "type" : "json_parse_exception",
        "reason" : "Unrecognized token 'org': was expecting ('true', 'false' or 'null')\n at [Source: java.io.StringReader@358b216d; line: 1, column: 4]"
      }
    ],
    "type" : "json_parse_exception",
    "reason" : "Unrecognized token 'org': was expecting ('true', 'false' or 'null')\n at [Source: java.io.StringReader@358b216d; line: 1, column: 4]"
  },
  "status" : 400
}
```

penghuo:
```
POST _opendistro/_sql/_explain
{  
  "query": """
    SHOW TABLES LIKE %
  """
}

{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[TABLE_CAT, TABLE_SCHEM, TABLE_NAME, TABLE_TYPE, REMARKS, TYPE_CAT, TYPE_SCHEM, TYPE_NAME, SELF_REFERENCING_COL_NAME, REF_GENERATION]"
    },
    "children": [
      {
        "name": "FilterOperator",
        "description": {
          "conditions": """like(TABLE_NAME, "%")"""
        },
        "children": [
          {
            "name": "ElasticsearchSystemIndexScan",
            "description": {
              "request": "ElasticsearchCatIndicesRequest{}"
            },
            "children": []
          }
        ]
      }
    ]
  }
}
```

Issue #324, opened 2019-12-17
Extend support for SHOW query to apply filter on `TABLE_CAT`, `TABLE_SCHEM` and `TABLE_TYPE`


rupal-bq:
Expected Result: List of filtered table, catalog, schema names, and table types. 

Currently, filter is available only for `TABLE_NAME`

For example, `SHOW TABLES LIKE kib*` returns a list of filtered `TABLE_NAME`
```
{
  "schema": [
    {
      "name": "TABLE_CAT",
      "type": "keyword"
    },
    {
      "name": "TABLE_SCHEM",
      "type": "keyword"
    },
    {
      "name": "TABLE_NAME",
      "type": "keyword"
    },
    {
      "name": "TABLE_TYPE",
      "type": "keyword"
    },
    {
      "name": "REMARKS",
      "type": "keyword"
    },
    {
      "name": "TYPE_CAT",
      "type": "keyword"
    },
    {
      "name": "TYPE_SCHEM",
      "type": "keyword"
    },
    {
      "name": "TYPE_NAME",
      "type": "keyword"
    },
    {
      "name": "SELF_REFERENCING_COL_NAME",
      "type": "keyword"
    },
    {
      "name": "REF_GENERATION",
      "type": "keyword"
    }
  ],
  "total": 4,
  "datarows": [
    [
      "odfe-cluster",
      null,
      "kibana_sample_data_ecommerce",
      "BASE TABLE",
      null,
      null,
      null,
      null,
      null,
      null
    ],
    [
      "odfe-cluster",
      null,
      "kibana_sample_data_flights",
      "BASE TABLE",
      null,
      null,
      null,
      null,
      null,
      null
    ],
    [
      "odfe-cluster",
      null,
      "kibana_sample_data_logs",
      "BASE TABLE",
      null,
      null,
      null,
      null,
      null,
      null
    ],
    [
      "odfe-cluster",
      null,
      "kibana_sample_data_types",
      "BASE TABLE",
      null,
      null,
      null,
      null,
      null,
      null
    ]
  ],
  "size": 4,
  "status": 200
}
```
Similar support is required for `TABLE_CAT`, `TABLE_SCHEM` and `TABLE_TYPE`.

Issue #238, opened 2019-10-18
<catalog>.<table> syntax not supported


jordanw-bq:
Low priority issue; we could not find this syntax specifically used with Tableau, but other BI tools and JDBC applications may use this type of syntax.

Somewhat similar to #213, but this appears in the `FROM` clause.

Using Docker image `amazon/opendistro-for-elasticsearch:1.1.0`
**Dataset:** Flights sample data (loaded through Kibana)
**SQL Plugin version:** 1.1.0.1

**Example Query:**
SELECT Carrier FROM odfe-cluster.kibana_sample_data_flights

**Results**
```
Raw response received: {
  "error": {
    "reason": "Invalid SQL query",
    "details": "Field [odfe-cluster.kibana_sample_data_flights] cannot be found or used here.",
    "type": "SemanticAnalysisException"
  },
  "status": 400
}
```

**Expectation:**
Able to recognize this as a qualified table name in the `odfe-cluster` catalog.

dai-chen:
I assume `odfe-cluster` is your ES cluster name configured in ES config yml. Need to check if we have this information in plugin code.

jordanw-bq:
Yes, this is the our configured cluster name.

PR #907, opened 2020-12-10
Support show and describe statement

penghuo:
## Related Issue: 
#404

## Description of changes
1. Add the support of SHOW and DESCRIBE.
2. Add orderedTable in the SymbotTable to keep the eleement ordered in the array. The reasons is when executing `SELECT * FROM TABEL` the * should be expanded following the order defined by TABLE schema.

## Breaking Change
In the legacy engine, if the index has alias, the index name will be returned instead of the alias name. e.g.
```
PUT demo_index/alias/demo_alias

SHOW TABLES LIKE demo_alias
```
The demo_index will be returned as table name instead of demo_alias. In the new engine, we change to retrun the alias name, in the example, the demo_alias will be returned.

## Todo
The current implementation doesn't existly follow the semantic definition of LIKE operator. Ideally, the operand of LIKE operator shoud be string literal to represent the match pattern. But the legacy engine use identifier in this case, e.g. `SHOW TABLES LIKE %`. For backward compability reason, we still support it now, but it will be deprecated soon.

## To Reviewer
I briefly describe the changed files in each module.
1. core. add the orderedTable.
2. docs. no breaking change. the fields order chagned.
3. elasticsearch. the major change to implement show and describe.
4. integ-test. only one breaking change in MetaDataQueriesIT which has been reimplemented in AdminIT which only run with new engine.
5. protocol. remove the legecy type convert to type class.
6. sql. front end change for show/describe.


chloe-zh (approved):

dai-chen (approved):
LGTM. Thanks! Could you check if any change needed to avoid this SHOW/DESCRIBE issue in old engine? https://github.com/opendistro-for-elasticsearch/sql/issues/404

And could you also check if these 2 can be supported now?

1. https://github.com/opendistro-for-elasticsearch/sql/issues/324: Filter by TABLE_CAT, TABLE_SCHEM and TABLE_TYPE
2. https://github.com/opendistro-for-elasticsearch/sql/issues/238: Support for `<catalog>.<table>`

penghuo:
> LGTM. Thanks! Could you check if any change needed to avoid this SHOW/DESCRIBE explain issue in new engine? #404
> 
> And could you also check if these 2 can be supported now?
> 
> 1. #324: Filter by TABLE_CAT, TABLE_SCHEM and TABLE_TYPE
> 2. #238: Support for `<catalog>.<table>`

* The #404 has been addressed.
* The #324, #238, we didn't support catalog now.

Issue #899, opened 2020-12-08
select item check issue for aggregate queries


rupal-bq:
For queries with HAVING clause only (without GROUP BY), the select list fails to handle functions applied to literal (it only allows literal in this case which is wrong)

Query:
```
SELECT (TO_DAYS(DATE('2050-01-01')) - 693961)
FROM `calcs`
HAVING (COUNT(1) > 0)
```

Error message:
```
"Explicit GROUP BY clause is required because expression [-(TO_DAYS(DATE(2050-01-01)), 693961)] contains non- aggregated column"
```

PR #913, opened 2020-12-10
add fix to handle functions applied to literal 

rupal-bq:
*Issue #899*

*Description of changes:*
- added filter for functions with literal in select clause
- added unit test
- added manual IT because the comparison test didn't work for the date function.

### DIFF
--- a/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/DateTimeFunctionIT.java
+++ b/integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/DateTimeFunctionIT.java
@@ -61,6 +61,15 @@ public class DateTimeFunctionIT extends SQLIntegTestCase {
   }
 
   @Test
+  public void testDateWithHavingClauseOnly() throws IOException {
+    JSONObject result =
+            executeQuery(String.format("SELECT (TO_DAYS(DATE('2050-01-01')) - 693961) FROM %s HAVING (COUNT(1) > 0)",TEST_INDEX_BANK) );
+    verifySchema(result,
+            schema("(TO_DAYS(DATE('2050-01-01')) - 693961)", null, "long"));
+    verifyDataRows(result, rows(54787));
+  }
+
+  @Test
   public void testAddDate() throws IOException {
     JSONObject result =
         executeQuery("select adddate(timestamp('2020-09-16 17:30:00'), interval 1 day)");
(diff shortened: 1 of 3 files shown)


sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/AstAggregationBuilder.java
```diff
@@ -137,6 +139,15 @@ private boolean isNonLiteral(UnresolvedExpression expr) {
     return !(expr instanceof Literal);
   }
 
+  private boolean isNonLiteralFunction(UnresolvedExpression expr) {
+    if (expr instanceof Function) {
+      List<? extends Node> children = expr.getChild();
+      return children.stream()
+              .allMatch(child -> isNonLiteral((UnresolvedExpression) child));
```
dai-chen, line 146: I think we need to check this recursively, ex. `ABS(ABS(age))`. Could we merge `isNonLiteral()` to this method as base case for recursion?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/DateTimeFunctionIT.java
```diff
@@ -60,6 +60,15 @@ public void testDateInGroupBy() throws IOException{
             rows("2018-08-11"));
   }
 
+  @Test
+  public void testDateWithHavingClauseOnly() throws IOException {
```
dai-chen, line 64: Could you add some more test queries as comparison test for this issue in https://github.com/opendistro-for-elasticsearch/sql/tree/develop/integ-test/src/test/resources/correctness/bugfixes

rupal-bq: Sure, I will add queries without the date function.

sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/AstAggregationBuilder.java
rupal-bq: right, I will add.

rupal-bq: added this change.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/DateTimeFunctionIT.java
rupal-bq: I added a few tests in comparison but while testing function without input PI() the results returned but another database was different (3.15 instead of 3.141592653589793) so I added manual IT for this.

dai-chen (approved):
Thanks for the fix! Just a minor request, could you remove `isNonLiteral()`? I think the current `isNonLiteralFunction()` takes care of it.

rupal-bq:
> Thanks for the fix! Just a minor request, could you remove `isNonLiteral()`? I think the current `isNonLiteralFunction()` takes care of it.

Sure, will remove it.

chloe-zh (approved):

Issue #916, opened 2020-12-11
Count() function not working when input is Date type


harold-wang:
Count() function not working when input is Date type including:
DATE
DATETIME
TIMESTAMP

PR #918, opened 2020-12-11
Enable Date type input in function Count() 

harold-wang:
Issue #916  , if available:

Description of changes:

* Add three input type (DATE, DATETIME, TIMESTAMP) in Count()
* Add UT (DATE, DATETIME or TIMESTAMP) input type


chloe-zh:
Could you add comparison test?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AggregatorFunction.java
```diff
@@ -92,6 +94,12 @@ private static FunctionResolver count() {
                 arguments -> new CountAggregator(arguments, INTEGER))
             .put(new FunctionSignature(functionName, Collections.singletonList(BOOLEAN)),
                 arguments -> new CountAggregator(arguments, INTEGER))
+            .put(new FunctionSignature(functionName, Collections.singletonList(DATE)),
+                arguments -> new CountAggregator(arguments, INTEGER))
+            .put(new FunctionSignature(functionName, Collections.singletonList(DATETIME)),
+                arguments -> new CountAggregator(arguments, INTEGER))
+            .put(new FunctionSignature(functionName, Collections.singletonList(TIMESTAMP)),
+                arguments -> new CountAggregator(arguments, INTEGER))
```
dai-chen, line 102: I assume SQL `COUNT()` should work with all types. If so, is it more maintainable to generate function signatures from all core types like this way: https://github.com/opendistro-for-elasticsearch/sql/blob/2aeb80b937788d5a7cb4ccdff21f082064898454/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/BinaryPredicateOperator.java#L172

dai-chen (commented):
Please add comparison test in https://github.com/opendistro-for-elasticsearch/sql/tree/develop/integ-test/src/test/resources/correctness/bugfixes. You need to create a txt file using issue number as name. This is mostly to avoid regression by future code changes.

harold-wang (commented):
> Could you add comparison test?

sure, thanks

Issue #925, opened 2020-12-11
Support cast(expr as text) should return CLOB


penghuo:
In H2 database, the cast(expr as text) return value in CLOB type. we need to more use case to understand why they have such implementation.

PR #926, opened 2020-12-12
Support cast function in SQL

penghuo:
*Issue #, if available:* N/A

*Description of changes:*
1. Implement cast function in SQL.
2. Update cast function doc. https://github.com/penghuo/sql/blob/support-cast/docs/user/dql/functions.rst#cast

*Todo:* #924, #925 


chloe-zh (approved):

dai-chen (approved):
Thanks for the changes!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/expression/Cast.java
```diff
@@ -0,0 +1,98 @@
...
+import lombok.ToString;
+
+/**
+ * AST node that represents Cast clause.
+ */
+@AllArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+@Getter
+@ToString
+public class Cast extends UnresolvedExpression {
```
dai-chen, line 50: Just curious, without this new node, can we handle all overloading signatures of `CAST` function in function repository?

penghuo: The second paramater is literaly string to indicate the target type, which is not variable/value.

Issue #928, opened 2020-12-12
Support null ordering for SQL window functions


dai-chen:
Currently `ORDER BY` in `OVER` clause only supports sort ordering `ASC/DESC`. As null ordering like `NULLS FIRST/LAST` is already supported in other order by clause, we need to support it for window functions too.

PR #929, opened 2020-12-12
Support NULLS FIRST/LAST ordering for window functions

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/928

*Description of changes:* Previously window function only parses sorting order `ASC/DESC`. This PR is to support null ordering `NULLS FIRST/LAST` for window function `OVER` clause too.

The basic logic is same as `ORDER BY` clause handling:

1. SQL frontend just pass sorting and null ordering given in queries.
2. The final ordering is determined in analyzer which has the knowledge of default ordering option.

*Testing*: Because both H2 and SQLite has problem with null ordering in OVER clause, a new IT class is added manually.

*Documentation*: https://github.com/dai-chen/sql/blob/support-null-first-last-for-window-function/docs/user/dql/window.rst#row-number


chloe-zh (approved):

penghuo (approved):
LGTM.

PR #938, opened 2020-12-15
add string function RIGHT

rupal-bq:
*Description of changes:*
- Added string function RIGHT().
- Added unit test.
- Added manual IT because the comparison test didn't work.
- Updated doc.


dai-chen (approved):
Thanks for the changes! Could you check if we can add comparison test now? https://github.com/opendistro-for-elasticsearch/sql/tree/develop/integ-test/src/test/resources/correctness/expressions

I recall there was some issue adding text function. I think it's doable now?

penghuo:
could you also add the doc for PPL? https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/experiment/ppl/functions/string.rst

chloe-zh (approved):

rupal-bq:
> Thanks for the changes! Could you check if we can add comparison test now? https://github.com/opendistro-for-elasticsearch/sql/tree/develop/integ-test/src/test/resources/correctness/expressions
> 
> I recall there was some issue adding text function. I think it's doable now?

Thanks. added comparison test.

rupal-bq:
> could you also add the doc for PPL? https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/experiment/ppl/functions/string.rst

Thanks. added this change.

Issue #941, opened 2020-12-16
Support Alias field or Throw proper error message for user.


penghuo:
The following query run null value, because we use @timestamp in the source filter. 
```
POST /_opendistro/_ppl
{
  "query": """
  source=kibana_sample_data_logs | fields @timestamp
  """
}
```

PR #942, opened 2020-12-16
Throw exception when access unknow field type

penghuo:
*Issue #, if available:* #941

*Description of changes:*
1. Remove the unknow type field from symbol table.
2. update the limitation doc to callout alias field type can't be used as identifier.
a. SQL, https://github.com/penghuo/sql/blob/issue-941/docs/user/limitations/limitations.rst
b. PPL, https://github.com/penghuo/sql/blob/issue-941/docs/experiment/ppl/limitations/limitations.rst


dai-chen (approved):
Thanks for the fix!

chloe-zh (approved):

PR #944, opened 2020-12-16
Fill hyphen strings to workbench table cells that have null and missing values

chloe-zh:
*Issue #, if available:*

*Description of changes:*

- Filled hyphen strings to indicate null and missing values
- Added unit test and snapshot
- Upgraded ts-jest and lockfile
Note: empty strings in result will not be filled

**Before**
image

**After**
image

### DIFF
--- a/workbench/package.json
+++ b/workbench/package.json
@@ -39,7 +39,7 @@
     "jest-raw-loader": "^1.0.1",
     "lint-staged": "^10.2.0",
     "mutationobserver-shim": "^0.3.3",
-    "ts-jest": "^25.4.0",
+    "ts-jest": "^26.1.0",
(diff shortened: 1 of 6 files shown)


joshuali925 (approved):
would this be confusing if the index actually has `null` string as values? Discover uses ` - ` to indicate null, do we want to be consistent with that?

also what happened to yarn.lock?

chloe-zh:
> would this be confusing if the index actually has `null` string as values? Discover uses `-` to indicate null, do we want to be consistent with that?
> 
> also what happened to yarn.lock?

That sounds good to use "-" to indicate null values. Will change it.

penghuo (approved):

Issue #770, opened 2020-10-07
Support aggregate window functions


dai-chen:
Add support for aggregate window functions: MIN, MAX, AVG, SUM, COUNT.

PR #946, opened 2020-12-18
Support aggregate window functions

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/770

*Description of changes:*

To reviewers, most file changes are due to 1 and 2a renaming. You can focus on core logic in `PeerRowsWindowFrame` and `AggregateWindowFunction`.

1. Use `NamedExpression` in logical and physical window operator: this is required because after expression analysis window function lost info in `OVER` clause, ex. `RANK() OVER(expr1), RANK(expr2)`
2. Refactor window frame
 a. Rename existing frame to `CurrentRowWindowFrame` which only maintains current row (and previous) for ranking window functions.
 b. Add new `PeerRowsWindowFrame` which preloads all peer rows for aggregate window function **(Please check the design doc for more details)**
3. Add `AggregateWindowFunction` adapter class to reuse existing aggregator

*Documentation*:

1. Design doc: https://github.com/dai-chen/sql/blob/support-aggregate-window-functions-2/docs/dev/AggregateWindowFunction.md
2. User manual: https://github.com/dai-chen/sql/blob/support-aggregate-window-functions-2/docs/user/dql/window.rst#aggregate-functions
3. Limitations: https://github.com/dai-chen/sql/blob/support-aggregate-window-functions-2/docs/user/limitations/limitations.rst#limitations-on-window-functions

*TODOs:* (covered in limitation doc)

1. Optimize reference to other expression in SELECT clause, ex. `SELECT SUM(a), SUM(SUM(a)) OVER(...) ...`
2. [Low] window function nested in other expression, ex. `CASE WHEN SUM(a) OVER(...) ...`
3. Enable circuit break for window operator in case of worst case happen (elaborated in design doc above).

*Testing:* Added UT and comparison test.


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/window/ranking/DenseRankFunction.java
```diff
@@ -30,7 +30,7 @@ public DenseRankFunction() {
   }
 
   @Override
-  protected int rank(CumulativeWindowFrame frame) {
+  protected int rank(CurrentRowWindowFrame frame) {
```
penghuo, line 33: Could we use PeerRowsWindowFrame in here?

chloe-zh (approved):

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/window/ranking/DenseRankFunction.java
dai-chen: I just dived into the code and it looks doable. Conceptually, `RANK` and `DENSE_RANK` share same meaning of "current row" and thus can use peer window frame. But I think we'd better use this minimal window frame for now to avoid potential data preload. We can revisit and refactor this later if necessary. Thanks!

penghuo (approved):

PR #960, opened 2021-01-05
Support filter clause in aggregations

chloe-zh:
*Issue #, if available:*
#964 

*Description of changes:*
- Added filter option for aggregate function in grammar
- Applied filter in aggregators
- Push down filter aggregation to ES DSL
- User doc: https://github.com/chloe-zh/sql/blob/adql/filter/docs/user/dql/aggregations.rst#filter-clause

### DIFF
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
@@ -43,13 +43,9 @@ public class AvgAggregator extends Aggregator<AvgAggregator.AvgState> {
   }
 
   @Override
-  public AvgState iterate(BindingTuple tuple, AvgState state) {
-    Expression expression = getArguments().get(0);
-    ExprValue value = expression.valueOf(tuple);
-    if (!(value.isNull() || value.isMissing()) && conditionValue(tuple)) {
-      state.count++;
-      state.total += ExprValueUtils.getDoubleValue(value);
-    }
+  protected AvgState iterate(ExprValue value, AvgState state) {
+    state.count++;
+    state.total += ExprValueUtils.getDoubleValue(value);
     return state;
   }
 
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
@@ -46,7 +46,7 @@ public class AvgAggregator extends Aggregator<AvgAggregator.AvgState> {
   public AvgState iterate(BindingTuple tuple, AvgState state) {
     Expression expression = getArguments().get(0);
     ExprValue value = expression.valueOf(tuple);
-    if (!(value.isNull() || value.isMissing())) {
+    if (!(value.isNull() || value.isMissing()) && condition(tuple)) {
       state.count++;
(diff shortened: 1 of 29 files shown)


elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/aggregation/dsl/MetricAggregationBuilder.java
```diff
@@ -62,28 +66,39 @@ public MetricAggregationBuilder(
...
         throw new IllegalStateException(
             String.format("unsupported aggregator %s", node.getFunctionName().getFunctionName()));
     }
   }
 
-  private ValuesSourceAggregationBuilder<?> make(ValuesSourceAggregationBuilder<?> builder,
-                                                  Expression expression) {
-    return helper.build(expression, builder::field, builder::script);
+  private QueryBuilder makeFilter(Expression condition) {
+    return new FilterQueryBuilder(serializer).build(condition);
```
penghuo, line 91: could we make FilterQueryBuilder as depedency?

elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/storage/script/aggregation/dsl/MetricAggregationBuilder.java
```diff
@@ -62,28 +66,39 @@ public MetricAggregationBuilder(
...
+  }
+
+  private AggregationBuilder make(ValuesSourceAggregationBuilder<?> builder,
+                                  Expression expression, Expression condition, String name) {
+    ValuesSourceAggregationBuilder aggBuilder =
+        helper.build(expression, builder::field, builder::script);
+    if (condition == null) {
+      return aggBuilder;
+    }
+    return AggregationBuilders.filter(name, makeFilter(condition)).subAggregation(aggBuilder);
```
penghuo, line 101: could we move these logic as a decoration to hide the implementation details, e.g.
```
return filter(childAggBuilder, name, condition)
```

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/aggregation/AvgAggregator.java
```diff
@@ -46,7 +46,7 @@ public AvgState create() {
   public AvgState iterate(BindingTuple tuple, AvgState state) {
     Expression expression = getArguments().get(0);
     ExprValue value = expression.valueOf(tuple);
-    if (!(value.isNull() || value.isMissing())) {
+    if (!(value.isNull() || value.isMissing()) && conditionValue(tuple)) {
```
dai-chen, line 49: Any chance to move this template code in each subclass to the parent class `Aggregator`? As what I see, each aggregator impl class only differs on how to accumulate state when this condition is true.

dai-chen (commented):
Please create an issue for this new feature. Thanks!

integ-test/src/test/resources/correctness/queries/filter.txt
```diff
@@ -0,0 +1,5 @@
+SELECT AVG(AvgTicketPrice) FILTER(WHERE Carrier = 'Kibana Airlines') AS filtered FROM kibana_sample_data_flights
```
dai-chen, line 1: Could you add test case to check if we can support function call in `WHERE`? Just want to check if pushdown works in this case.

dai-chen (approved):
LGTM. Thanks!

penghuo (approved):

Issue #950, opened 2020-12-23
Add flow control functions ifNull, nullIf and  isNull


harold-wang:
Add flow control Functions ifNull, nullIf and  isNull

PR #962, opened 2021-01-06
Enable sql function ifnull, nullif and isnull

harold-wang:
Issue #950 , if available:

Description of changes:
Add flow control functions ifNull, nullIf and isNull (#950)

* Init
* UT
* IT

### DIFF
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/BuiltinFunctionName.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/BuiltinFunctionName.java
@@ -137,10 +137,10 @@ public enum BuiltinFunctionName {
    * NULL Test.
    */
   IS_NULL(FunctionName.of("is null")),
-  ISNULL(FunctionName.of("isnull")),
   IS_NOT_NULL(FunctionName.of("is not null")),
   IFNULL(FunctionName.of("ifnull")),
   NULLIF(FunctionName.of("nullif")),
+  ISNULL(FunctionName.of("isnull")),
 
   ROW_NUMBER(FunctionName.of("row_number")),
   RANK(FunctionName.of("rank")),
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/BuiltinFunctionName.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/BuiltinFunctionName.java
@@ -137,7 +137,10 @@ public enum BuiltinFunctionName {
    * NULL Test.
    */
   IS_NULL(FunctionName.of("is null")),
+  ISNULL(FunctionName.of("isnull")),
   IS_NOT_NULL(FunctionName.of("is not null")),
+  IFNULL(FunctionName.of("ifnull")),
+  NULLIF(FunctionName.of("nullif")),
 
   ROW_NUMBER(FunctionName.of("row_number")),
   RANK(FunctionName.of("rank")),
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
@@ -15,16 +15,25 @@
 
 package com.amazon.opendistroforelasticsearch.sql.expression.operator.predicate;
 
+import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.LITERAL_NULL;
+import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.nullValue;
 import static com.amazon.opendistroforelasticsearch.sql.data.type.ExprCoreType.BOOLEAN;
+import static com.amazon.opendistroforelasticsearch.sql.data.type.ExprCoreType.UNKNOWN;
+import static com.amazon.opendistroforelasticsearch.sql.expression.function.FunctionDSL.impl;
 
 import com.amazon.opendistroforelasticsearch.sql.data.model.ExprBooleanValue;
 import com.amazon.opendistroforelasticsearch.sql.data.model.ExprValue;
 import com.amazon.opendistroforelasticsearch.sql.data.type.ExprCoreType;
+import com.amazon.opendistroforelasticsearch.sql.data.type.ExprType;
 import com.amazon.opendistroforelasticsearch.sql.expression.function.BuiltinFunctionName;
 import com.amazon.opendistroforelasticsearch.sql.expression.function.BuiltinFunctionRepository;
 import com.amazon.opendistroforelasticsearch.sql.expression.function.FunctionDSL;
+import com.amazon.opendistroforelasticsearch.sql.expression.function.FunctionName;
 import com.amazon.opendistroforelasticsearch.sql.expression.function.FunctionResolver;
+import com.amazon.opendistroforelasticsearch.sql.utils.OperatorUtils;
+
 import java.util.Arrays;
+import java.util.List;
 import java.util.stream.Collectors;
 import lombok.experimental.UtilityClass;
 
@@ -41,6 +50,8 @@ public class UnaryPredicateOperator {
     repository.register(not());
     repository.register(isNull());
     repository.register(isNotNull());
+    repository.register(nullIf());
+    repository.register(ifNull());
   }
 
   private static FunctionResolver not() {
@@ -65,7 +76,6 @@ public class UnaryPredicateOperator {
   }
 
   private static FunctionResolver isNull() {
-
     return FunctionDSL
         .define(BuiltinFunctionName.IS_NULL.getName(), Arrays.stream(ExprCoreType.values())
             .map(type -> FunctionDSL
@@ -82,4 +92,66 @@ public class UnaryPredicateOperator {
             .collect(
                 Collectors.toList()));
   }
+
+  private static FunctionResolver ifNull() {
+    FunctionName functionName = BuiltinFunctionName.IFNULL.getName();
+    List<ExprType> typeList = ExprCoreType.coreTypes();
+    typeList.add(UNKNOWN);
+    FunctionResolver functionResolver =
+        FunctionDSL.define(functionName,
+            typeList.stream().map(v ->
+              impl((UnaryPredicateOperator::exprIfNull), v, v, v))
+              .collect(Collectors.toList())
+        );
+    return functionResolver;
+  }
+
+  private static FunctionResolver nullIf() {
+    FunctionName functionName = BuiltinFunctionName.NULLIF.getName();
+    List<ExprType> typeList = ExprCoreType.coreTypes();
+    typeList.add(UNKNOWN);
+    FunctionResolver functionResolver =
+        FunctionDSL.define(functionName,
+            typeList.stream().map(v ->
+              impl((UnaryPredicateOperator::exprNullIf), v, v, v))
+              .collect(Collectors.toList()));
+    return functionResolver;
+  }
+
+  /** v2 if v1 is null.
+   * @param v1 varable 1
+   * @param v2 varable 2
+   * @return v2 if v1 is null
+   */
+  public static ExprValue exprIfNull(ExprValue v1, ExprValue v2) {
+    if (v1.isNull()) {
+      return v2;
+    } else if (v1.isMissing()) {
+      return v2;
+    } else {
+      return v1;
+    }
+  }
+
+  /** null if v1 equls to v2.
+   * @param v1 varable 1
+   * @param v2 varable 2
+   * @return null if v1 equls to v2
+   */
+  public static ExprValue exprNullIf(ExprValue v1, ExprValue v2) {
+    if (v1.isNull()) {
+      return v1;
+    } else if (v1.isMissing()) {
+      return v1;
+    } else if (v2.isNull()) {
+      return v1;
+    } else if (v2.isMissing()) {
+      return v1;
+    } else if (v1.value().equals(v2.value())) {
+      return LITERAL_NULL;
+    } else {
+      return v1;
+    }
+  }
+
 }
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
@@ -90,11 +90,11 @@ public class UnaryPredicateOperator {
     List<ExprType> typeList = ExprCoreType.coreTypes();
     typeList.add(UNKNOWN);
     return FunctionDSL
-            .define(functionName, typeList.stream()
-                    .map(type -> FunctionDSL
-                            .impl((v) -> ExprBooleanValue.of(v.isNull()), BOOLEAN, type))
-                    .collect(
-                            Collectors.toList()));
+        .define(functionName, typeList.stream()
+            .map(type -> FunctionDSL
+                .impl((v) -> ExprBooleanValue.of(v.isNull()), BOOLEAN, type))
+            .collect(
+                Collectors.toList()));
   }
 
   private static FunctionResolver isNotNull() {
@@ -137,13 +137,7 @@ public class UnaryPredicateOperator {
    * @return v2 if v1 is null
    */
   public static ExprValue exprIfNull(ExprValue v1, ExprValue v2) {
-    if (v1.isNull()) {
-      return v2;
-    } else if (v1.isMissing()) {
-      return v2;
-    } else {
-      return v1;
-    }
+    return (v1.isNull() || v1.isMissing()) ? v2 : v1;
   }
 
   /** null if v1 equls to v2.
@@ -152,13 +146,7 @@ public class UnaryPredicateOperator {
    * @return null if v1 equls to v2
    */
   public static ExprValue exprNullIf(ExprValue v1, ExprValue v2) {
-    if (v1.isNull()) {
-      return v1;
-    } else if (v1.isMissing()) {
-      return v1;
-    } else if (v2.isNull()) {
-      return v1;
-    } else if (v2.isMissing()) {
+    if (v1.isNull() || v1.isMissing() || v2.isNull() || v2.isMissing()) {
       return v1;
     } else if (v1.value().equals(v2.value())) {
       return LITERAL_NULL;
--- a/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
+++ b/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
@@ -16,6 +16,7 @@
 package com.amazon.opendistroforelasticsearch.sql.expression.operator.predicate;
 
 import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.LITERAL_FALSE;
+import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.LITERAL_MISSING;
 import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.LITERAL_NULL;
 import static com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.LITERAL_TRUE;
 
@@ -53,11 +54,11 @@ public class UnaryPredicateOperator {
    */
   public static void register(BuiltinFunctionRepository repository) {
     repository.register(not());
-    repository.register(isNull());
     repository.register(isNotNull());
     repository.register(ifNull());
     repository.register(nullIf());
-    repository.register(is_Null());
+    repository.register(is_Null(BuiltinFunctionName.IS_NULL));
+    repository.register(is_Null(BuiltinFunctionName.ISNULL));
   }
 
   private static FunctionResolver not() {
@@ -81,27 +82,13 @@ public class UnaryPredicateOperator {
     }
   }
 
-  private static FunctionResolver is_Null() {
+  private static FunctionResolver is_Null(BuiltinFunctionName funcName) {
     return FunctionDSL
-        .define(BuiltinFunctionName.IS_NULL.getName(), Arrays.stream(ExprCoreType.values())
+        .define(funcName.getName(), Arrays.stream(ExprCoreType.values())
             .map(type -> FunctionDSL
                 .impl((v) -> ExprBooleanValue.of(v.isNull()), BOOLEAN, type))
             .collect(
                 Collectors.toList()));
-  }
-
-  private static FunctionResolver isNull() {
-    FunctionName functionName = BuiltinFunctionName.ISNULL.getName();
-    List<ExprType> typeList = ExprCoreType.coreTypes();
-    typeList.add(UNKNOWN);
-
-    List<SerializableFunction<FunctionName, org.apache.commons.lang3.tuple.Pair<FunctionSignature,
-            FunctionBuilder>>> functionsOne = typeList.stream().map(v ->
-            impl((UnaryPredicateOperator::exprIsNull), BOOLEAN, v))
-            .collect(Collectors.toList());
-
-    FunctionResolver functionResolver = FunctionDSL.define(functionName, functionsOne);
-    return functionResolver;
   }
 
   private static FunctionResolver isNotNull() {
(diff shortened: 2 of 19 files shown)


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -82,4 +102,66 @@ private static FunctionResolver isNotNull() {
...
+   * @return v2 if v1 is null
+   */
+  public static ExprValue exprIfNull(ExprValue v1, ExprValue v2) {
+    if (v1.isNull()) {
+      return v2;
+    } else if (v1.isMissing()) {
+      return v2;
+    } else {
+      return v1;
+    }
```
chloe-zh, line 143: Not a problem: combine the isNull and isMissing conditions?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -82,4 +102,66 @@ private static FunctionResolver isNotNull() {
...
+      return v1;
+    } else if (v2.isNull()) {
+      return v1;
+    } else if (v2.isMissing()) {
+      return v1;
+    } else if (v1.value().equals(v2.value())) {
+      return LITERAL_NULL;
+    } else {
+      return v1;
+    }
```
chloe-zh, line 164: combine conditions lines to more readable like above?

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperatorTest.java
```diff
@@ -79,4 +104,93 @@ public void is_not_null_predicate() {
...
+  }
+
+  @Test
+  public void test_null_if_predicate() {
+    Expression v1 = dsl.literal(100);
+    Expression v2 = dsl.literal(200);
+    FunctionExpression result = dsl.nullif(v1, v2);
+    assertEquals(v1.valueOf(valueEnv()), result.valueOf(valueEnv()));
+
+    System.out.println("debug : " + result.valueOf(valueEnv()));
```
chloe-zh, line 138: remove this line?

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperatorTest.java
```diff
@@ -79,4 +104,93 @@ public void is_not_null_predicate() {
...
+
+    v1 = dsl.literal(100);
+    v2 = dsl.literal(100);
+
+    result = dsl.nullif(v1, v2);
```
chloe-zh, line 143: remove the above line; compare the result with expected result?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
```diff
@@ -0,0 +1,121 @@
...
+import org.elasticsearch.search.SearchHits;
+import org.json.JSONObject;
+
+import org.junit.Assume;
+import org.junit.Test;
+
+
+/**
+ * Created by allwefantasy on 8/25/16.
+ */
```
chloe-zh, line 45: remove these lines?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
```diff
@@ -0,0 +1,121 @@
...
+public class FlowConrolFunctionIT extends SQLIntegTestCase {
```
chloe-zh, line 46: Typo: Control

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @Override
+  public void init() throws Exception {
+    super.init();
+    TestUtils.enableNewQueryEngine(client());
+    loadIndex(Index.BANK);
+    loadIndex(Index.ACCOUNT);
+    loadIndex(Index.ONLINE);
+    loadIndex(Index.DATE);
+  }
```
chloe-zh, line 56: It's fine to load only the account index and remove others since only the account index is used in this IT class. Could you also add some test cases over the null/missing values. The index `bank_with_null_values.json` might be helpful since it contains a couple of null or missing values in some fields.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
```diff
@@ -0,0 +1,121 @@
...
+    final XContentParser parser = XContentFactory.xContent(XContentType.JSON).createParser(
+        NamedXContentRegistry.EMPTY,
+        LoggingDeprecationHandler.INSTANCE,
+        rsp);
+    return SearchResponse.fromXContent(parser).getHits();
+  }
+
+  private JSONObject executeJdbcRequest(String query) {
+    return new JSONObject(executeQuery(query, "jdbc"));
+  }
```
chloe-zh, line 120: Could you use the `executeQuery` method inherited from the parent class `SQLIntegTestCase` to execute SQL queries, it will enforce to JDBC format by default. For the assertion, you can check out the `verifySchema`, `verifyDataRows` etc. methods in `MatcherUtils`. For example:
```
JSONObject result = executeQuery(SELECT ISNULL(lastname) AS name ...);
verifySchema(result, schema("ISNULL(lastname)"), "name" , "boolean");
verifyDataRows(...);
```
You can also checkout the IT cases in `DataTimeFunctionIT.java` for reference.

chloe-zh:
Could you also update the user manual here: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/dql/functions.rst#conditional-functions

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
harold-wang: Got it, thanks

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
harold-wang: Thanks, to fix in next commit shortly.

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperatorTest.java
harold-wang: Thanks, to remove it in next commit.

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperatorTest.java
harold-wang: got it, thx

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
harold-wang: Got it, thx.

legacy/src/main/antlr/OpenDistroSqlLexer.g4
```diff
@@ -171,6 +171,7 @@ MONTH:                              'MONTH';
 MONTHNAME:                          'MONTHNAME';
 MULTIPLY:                           'MULTIPLY';
 NOW:                                'NOW';
+NULLIF:                             'NULLIF';
```
dai-chen, line 174: No need to touch this file since this is legacy grammar file?

integ-test/src/test/resources/correctness/queries/flowcontrol.txt
```diff
@@ -0,0 +1,8 @@
+SELECT IFNULL(AvgTicketPrice, AvgTicketPrice * 2.0) FROM kibana_sample_data_flights
```
dai-chen, line 1: To simplify, you can move these tests to https://github.com/opendistro-for-elasticsearch/sql/blob/develop/integ-test/src/test/resources/correctness/expressions/conditionals.txt. In this way, you don't need a `FROM` clause and test index. The only literal not supported is `NULL` but you've figured out the workaround (1/0).

dai-chen (commented):
Please add doctest here: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/dql/functions.rst#conditional-functions

dai-chen:
Btw, I missed `IF` function, could you add it too? You can do it same here or a separate PR. Thanks!

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowConrolFunctionIT.java
harold-wang: Got it, thanks

integ-test/src/test/resources/correctness/queries/flowcontrol.txt
harold-wang: done.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -82,4 +106,62 @@ private static FunctionResolver isNotNull() {
             .collect(
                 Collectors.toList()));
   }
+
+  private static FunctionResolver ifNull() {
+    FunctionName functionName = BuiltinFunctionName.IFNULL.getName();
+    List<ExprType> typeList = ExprCoreType.coreTypes();
+    typeList.add(UNKNOWN);
```
harold-wang, line 113: Chen, we have to add UNKNOWN, otherwise, we will see the "UNKNOWN" type missing error, I think it is NONE type.
est_nullif_predicate()
com.amazon.opendistroforelasticsearch.sql.exception.ExpressionEvaluationException: nullif function expected {[BYTE,BYTE],[SHORT,SHORT],[INTEGER,INTEGER],[LONG,LONG],[FLOAT,FLOAT],[DOUBLE,DOUBLE],[BOOLEAN,BOOLEAN],[STRING,STRING],[TIMESTAMP,TIMESTAMP],[DATE,DATE],[TIME,TIME],[DATETIME,DATETIME],[INTERVAL,INTERVAL],[STRUCT,STRUCT],[ARRAY,ARRAY]}, but get [UNKNOWN,UNKNOWN]
	at com.amazon.opendistroforelasticsearch.sql.expression.function.FunctionResolver.resolve(FunctionResolver.java:46)
	at com.amazon.opendistroforelasticsearch.sql.expression.function.BuiltinFunctionRepository.resolve(BuiltinFunctionRepository.java:44)
	at com.amazon.opendistroforelasticsearch.sql.expression.function.BuiltinFunctionRepository.compile(BuiltinFunctionRepository.java:30)
	at com.amazon.opendistroforelasticsearch.sql.expression.DSL.function(DSL.java:500)
	at com.amazon.opendistroforelasticsearch.sql.expression.DSL.nullif(DSL.java:526)
	at com.amazon.opendistroforelasticsearch.sql.expression.operator.predicate.UnaryPredicateOperatorTest.test_nullif_predicate(UnaryPredicateOperatorTest.java:138)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)


test_isnull_predicate()
com.amazon.opendistroforelasticsearch.sql.exception.ExpressionEvaluationException: isnull function expected {[BYTE],[SHORT],[INTEGER],[LONG],[FLOAT],[DOUBLE],[BOOLEAN],[STRING],[TIMESTAMP],[DATE],[TIME],[DATETIME],[INTERVAL],[STRUCT],[ARRAY]}, but get [UNKNOWN]

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/correctness/CorrectnessIT.java
```diff
@@ -52,7 +52,8 @@
  * Correctness integration test by performing comparison test with other databases.
  */
 @ESIntegTestCase.SuiteScopeTestCase
-@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.SUITE, numDataNodes = 3, supportsDedicatedMasters = false, transportClientRatio = 1)
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.SUITE, numDataNodes = 3,
```
penghuo, line 55: what is the difference?

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowControlFunctionIT.java
```diff
@@ -0,0 +1,212 @@
...
+    loadIndex(Index.ACCOUNT);
+    loadIndex(Index.BANK_WITH_NULL_VALUES);
+  }
+
+  @Test
+  public void ifnullShouldPassJDBC() throws IOException {
+    JSONObject response = executeJdbcRequest(
+        "SELECT IFNULL(lastname, 'unknown') AS name FROM " + TEST_INDEX_ACCOUNT
+            + " GROUP BY name");
+    assertEquals("IFNULL(lastname, \'unknown\')", response.query("/schema/0/name"));
```
penghuo, line 58: could you use verifySchema and verifyDataRows to get rid of jsonpath access?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/BuiltinFunctionName.java
```diff
@@ -138,6 +138,9 @@
    */
   IS_NULL(FunctionName.of("is null")),
   IS_NOT_NULL(FunctionName.of("is not null")),
+  IFNULL(FunctionName.of("ifnull")),
+  NULLIF(FunctionName.of("nullif")),
+  ISNULL(FunctionName.of("isnull")),
```
penghuo, line 143: why we need ISNULL?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
penghuo: The null value doesn't have a type. the problem only related to NULL literal test, e.g. ifnull(NULL), do we need to support NULL literal?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -82,4 +106,62 @@ private static FunctionResolver isNotNull() {
...
+
+    List<SerializableFunction<FunctionName, org.apache.commons.lang3.tuple.Pair<FunctionSignature,
+            FunctionBuilder>>> functionsOne = typeList.stream().map(v ->
+            impl((UnaryPredicateOperator::exprIfNull), v, v, v))
```
penghuo, line 117: do we require the paramater of ifnull have same type? ifnull(expr1, expr2), the type of expr1 and expr2 must be same, right?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -82,4 +106,62 @@ private static FunctionResolver isNotNull() {
...
+
+  /** null if v1 equls to v2.
+   * @param v1 varable 1
+   * @param v2 varable 2
+   * @return null if v1 equls to v2
+   */
+  public static ExprValue exprNullIf(ExprValue v1, ExprValue v2) {
+    if (v1.isNull() || v1.isMissing() || v2.isNull() || v2.isMissing()) {
+      return v1;
+    } else if (v1.value().equals(v2.value())) {
```
penghuo, line 160: could we use the v1.equals(v2) directly?

penghuo:
* Please add the documentation accordingly
* Remove the leading "*" in the PR title

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
harold-wang: yes, thx

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
harold-wang: yes, expr1 and expr2 must be same, otherwise return value is correct, but return type could be wrong.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/function/BuiltinFunctionName.java
harold-wang: isnull(expr1, expr2), is_null is for is null(with space between is and null).

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
harold-wang: public class Main {
  public static void main(String[] args) {
    String myStr1 = null;
    System.out.println("debug : " + myStr1.equals("test"));
  }
}
Exception in thread "main" java.lang.NullPointerException	at Main.main(Main.java:4)
if v1 is null, using the v1.equals(v2) directly will cause exception.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowControlFunctionIT.java
harold-wang: Done, thanks

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
dai-chen: @harold-wang I think you can remove null literal test from your UT and code here. This null literal support is missing for all expressions. We need to handle it in one place.

harold-wang: Let me remove null literal, thanks @penghuo @dai-chen

harold-wang: @dai-chen remove the UNKNOWN(null) type will make isnull() function totally useless, how about to keep it for now and remove it once you can enable it in generic way, thx.

dai-chen: @harold-wang I don't think so. If we don't consider NULL literal, you won't get any UNKNOWN type input. For example, in your IT, the value of 1/0 is null but its type is integer. Could you post an example in which test case this is required?

harold-wang: All IT passed without UNKNOWN type, but UT has some coverage issue: what make sense to me:

  public static ExprValue exprIsNull(ExprValue v) {
    if (!v.isNull() && !v.isMissing()) {
      return LITERAL_FALSE;  // added by Harold, how can we cover here without UNKNOWN type.
    } else {
      return LITERAL_TRUE;
    }
  }

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -74,6 +90,20 @@ private static FunctionResolver isNull() {
                 Collectors.toList()));
   }
 
+  private static FunctionResolver isNull() {
```
chloe-zh, line 93: From my point of view, this method should be identical to the `is_Null()` method above. The setting of `UNKNOWN` type is not necessary

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -64,10 +82,9 @@ public ExprValue not(ExprValue v) {
     }
   }
 
-  private static FunctionResolver isNull() {
-
+  private static FunctionResolver is_Null(BuiltinFunctionName funcName) {
```
dai-chen, line 85: Could you keep original name `isNull`?

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperatorTest.java
```diff
@@ -59,18 +65,37 @@ public void test_not_missing() {
...
     assertEquals(LITERAL_TRUE, expression.valueOf(valueEnv()));
   }
 
   @Test
-  public void is_not_null_predicate() {
+  public void test_isnull_predicate() {
+    ArrayList<Expression> exprValueArrayList = new ArrayList<>();
+    exprValueArrayList.add(dsl.literal("test"));
+    exprValueArrayList.add(dsl.literal(100));
+    exprValueArrayList.add(dsl.literal(""));
```
dai-chen, line 83: np: you can refactor this in your PR by JUnit's parameterized test. Example: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/function/WideningTypeRuleTest.java#L103

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/FlowControlFunctionIT.java
```diff
@@ -0,0 +1,209 @@
...
+import org.elasticsearch.common.xcontent.XContentFactory;
+import org.elasticsearch.common.xcontent.XContentParser;
+import org.elasticsearch.common.xcontent.XContentType;
+import org.elasticsearch.search.SearchHits;
+import org.json.JSONObject;
+
+import org.junit.Assume;
+import org.junit.Test;
+
+public class FlowControlFunctionIT extends SQLIntegTestCase {
```
dai-chen, line 43: np: noticed we use "conditional" in doc and comparison test. could you rename to `ConditionalIT` instead?

docs/experiment/ppl/functions/condition.rst
```diff
@@ -67,3 +67,78 @@ Example, the account 13 doesn't have email field::
...
+------
+
+Description
+>>>>>>>>>>>
+
+Usage: ifnull(field1, field2) return field2 if field1 is null.
+
+Argument type: all the supported data type, (NOTE : if two parametershas different type, return result is correct, but return type could be wrong)
+
+Return type: any
```
dai-chen, line 80: Fo argument type, it should be "if two parameters have different type, you will fail semantic check"?

For return type, to avoid confusion, could you elaborate a little more than "any"?

Same for other function docs in PPL and SQL added.

dai-chen (approved):
Just several minor comments. Thanks for your changes!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
harold-wang: UNKNOWN is removed now.

core/src/test/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperatorTest.java
harold-wang: Done,thx

chloe-zh (approved):

Issue #441, opened 2020-04-25
group by query cannot support offset


zerix:
I used elasticsearch 7.6.1 with SQL 1.6.0

I founded that the count(*) group by query cannot support offset argment.

the query :
{
    "query": "select customer_id, count(*) as label_value from kibana_sample_data_ecommerce group by (customer_id) order by _id limit 10 offset 2"
}

the offset argument not work, it is a bug or not supported future?

abbashus:
The plugin currently does not support offset for `GROUP BY`. We don't have a plan/ETA to add this enhancement soon.

dheerajravindranath:
It would be great if we get this feature soon!

dai-chen:
> It would be great if we get this feature soon!

Sure, we're planning to reimplement `GROUP BY` by ES composite aggregation API. Will evaluate this feature then. Thanks!

dai-chen:
This issue has been fixed in our new SQL engine. Please find more details in PR above and README here: https://github.com/opendistro-for-elasticsearch/sql#experimental

PR #973, opened 2021-01-07
Adding additional logging to ExprMissingValue exception

FreCap:
*Issue #, if available:*

In a simple GROUP BY queries I'm getting the following exception and need more info about the problem:
```
{
  "error": {
    "details": "invalid value operation on MISSING_VALUE",
    "type": "IllegalStateException"
  },
  "status": 500
}
```

FAILS
```
POST _opendistro/_sql
{
  "query": "select ABC.Status, count(*) from my_index group by ABC.Status"
}
```

WORKS with ?format=json
```
POST _opendistro/_sql?format=json
{
  "query": "select ABC.Status, count(*) from my_index group by ABC.Status"
}

```

*Description of changes:*
Adding additional details to the error message

### DIFF
--- a/legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/expression/model/ExprMissingValue.java
+++ b/legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/expression/model/ExprMissingValue.java
@@ -15,10 +15,25 @@
 
 package com.amazon.opendistroforelasticsearch.sql.legacy.expression.model;
 
+import java.util.Set;
+
 /**
  * The definition of the missing value.
  */
 public class ExprMissingValue implements ExprValue {
+    private final String missingValue;
+    private final Set<String> availableValues;
+
+    public ExprMissingValue(String missingValue, Set<String> availableValues) {
+        this.missingValue = missingValue;
+        this.availableValues = availableValues;
+    }
+
+    @Override
+    public Object value() {
+        throw new IllegalStateException("invalid value operation on " + kind() + " (" + this.missingValue + "), available: " + this.availableValues.toString());
+    }
+
     @Override
     public ExprValueKind kind() {
         return ExprValueKind.MISSING_VALUE;
--- a/legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/expression/domain/BindingTuple.java
+++ b/legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/expression/domain/BindingTuple.java
@@ -46,7 +46,7 @@ public class BindingTuple {
      * @return binding value.
      */
     public ExprValue resolve(String bindingName) {
-        return bindingMap.getOrDefault(bindingName, new ExprMissingValue());
+        return bindingMap.getOrDefault(bindingName, new ExprMissingValue(bindingName, bindingMap.keySet()));
     }
 
     @Override


legacy/src/main/java/com/amazon/opendistroforelasticsearch/sql/legacy/expression/model/ExprMissingValue.java
```diff
@@ -15,10 +15,25 @@
 
 package com.amazon.opendistroforelasticsearch.sql.legacy.expression.model;
 
+import java.util.Set;
+
 /**
  * The definition of the missing value.
  */
 public class ExprMissingValue implements ExprValue {
+    private final String missingValue;
```
penghuo, line 24: Is it possible to add log for debug purpose? from my understanding, the constructor of ExprMissingValue making it doesn't have clearly definition.

FreCap: Hi @penghuo, could you add some info on your advice?
Are you suggesting that "missingValue" doesn't belong to the constructor of ExprMissingValue?
Thank you

penghuo: Yes, the missing value should have not value.
No idea in my mind now, just wondering is it possible to do it with bettern log/exception?

FreCap: Should this be considered a _user_ error (which the user could resolve), or a _full server error_ in which only the _developer only_ would see the exception in the logs?

Being ES a product used by other devs, I would assume that for **no query** as a user I should take a look at the internal logs, but the logs where reserved to internal indexing errors/quorum/networking problems etc..

On top of that, I've a cluster with 50+ machines and having to look at the logs, looks less and less of an option since it creates a lot of overhead.

As a user I would love to know which field is missing/unsupported if we know it in code. Otherwise all the users (even just to report the problem), would require server level access to debug the issue further.
On top of that, I've a cluster with 50+ machines and having to look at the logs, looks less and less of an option since it creates a lot of overhead.

I love this project but more than once I've been completely lost when the server errors look fairly generic to the consumer (e.g. engineer of another team) and to the system maintainer (the one who installs it on ES). I've noticed there are a number of exceptions that could be improved by specifying at least the "subject/operator/portion of the query" that caused the exception, and I'm happy to contribute to move this direction.

To take this further we could even think of a debug mode to see all of the exceptions related to a given call directly in the response, a sort of stack trace.

E.g.
```

curl -H 'Content-Type: application/json' -X PUT localhost:9200/_opendistro/_sql/settings -d '{
  "transient" : {
    "opendistro.sql.enabledDebugInResponse" : "true"
  }
}'

{
  "error": {
    "details": "invalid value operation on MISSING_VALUE",
**********************
    "extra details":["ALL", 
                              "THE", 
                              "LOGS",
                              "ASSOCIATED with this call"]
**********************
    "type": "IllegalStateException"
  },
  "status": 500
}
```

penghuo: Hi, 
For this issue, I think it should be considered as bug in our old engine. 
The missing value represent the field value which is missing. e.g. the tuple employee
```
{
  "name": "bob"
}
```
when resolve the employee.age, the result is missing value. Semantically, the missing value in the your example all has meaning which should be handled properly.
My suggestion are:
* For this bug, you could find the solution to fix this bug.
* If customer do want know which value is null/missing, they should use isnull function.

I agree with your idea to extral debug mode which will help developer, thanks for bringing this up.

FreCap: Hi @penghuo, thank you for the answer.

I'm fairly sure that this bug is still in progress since I enabled the new engine
```


PUT _opendistro/_sql/settings
{
  "transient" : {
    "opendistro.sql.engine.new.enabled" : "true"
  }
}
```

I understand that there is a bug or missing feature on the server side, and with isNull I could find out which field is null, but here the goal is to tell to who runs the query what might have gone wrong so he/she can find a workaround meanwhile a fix comes out.

A lot of the excpetions that are thrown could be pure missing features, IMHO the user should be hinted on to which direction to go, and which feature/parameter is the offender in the current state of the implementation before a patch comes out.

The example I gave has only two parameters, but I cannot run N queries to find out which field isnull if I have N parameters, do you agree?

I'd be happy to help solving the bug, but for now I feel it would be a good step towards it.

I'm still not super familiar with the dev of the project, if you can point out which classes should be modified to solve the problem and what would be the expected behavior, I'd be happy to take a look

penghuo: By the way, could u share the test data and mapping. 
I am testing with following data and query with new engine disabled, it works.
```
PUT leaders/_doc/1
{ 
  "region": "US",
  "manager": { 
    "age":     30,
    "name": { 
      "first": "John",
      "last":  "Smith"
    }
  }
}
POST /_opendistro/_sql
{
  "query": """
    SELECT manager.age, count(*) FROM leaders GROUP BY manager.age
  """
}
```

FreCap: Hi @penghuo, unfortunately I cannot share the dataset, and I couldn't make up an example that is failing. 
All examples with small datasets I tried were succeeding, but with a larger dataset was failing.

Adding that logging would allow me getting some further insight in what went wrong

FreCap: Hi @penghuo, 
could you please help me further understanding the concern in adding the additional in-request logging?
Thank you for your time

penghuo: The sample dataset to reproduce the issue. same issue #441 

```
PUT debug/_doc/1
{ 
  "region": "US",
  "manager": { 
    "age":     30,
    "name": "bob"
  }
}

POST /_opendistro/_sql
{
  "query": """
    SELECT manager.name, count(*) FROM debug GROUP BY manager.name
  """
}

{
  "error": {
    "details": "invalid value operation on MISSING_VALUE (manager#name), available: [COUNT_0, manager.name.keyword]",
    "type": "IllegalStateException"
  },
  "status": 500
}
```
In our NewEngine, we already change the behaviour of the [value function](https://github.com/opendistro-for-elasticsearch/sql/blob/develop/core/src/main/java/com/amazon/opendistroforelasticsearch/sql/data/model/ExprMissingValue.java#L36) which will return null value instead of throw exception.

My suggestion is throw exception or log error messge in the resolve place.
```
@RequiredArgsConstructor
public class PhysicalProject implements PhysicalOperator<BindingTuple> {

    private static final Logger LOG = LogManager.getLogger(PhysicalProject.class);

    private final PhysicalOperator<BindingTuple> next;
    private final List<ColumnNode> fields;

    @Override
    public Cost estimate() {
        return null;
    }

    @Override
    public PlanNode[] children() {
        return new PlanNode[]{next};
    }

    @Override
    public boolean hasNext() {
        return next.hasNext();
    }

    @Override
    public Row<BindingTuple> next() {
        BindingTuple input = next.next().data();
        BindingTuple.BindingTupleBuilder outputBindingTupleBuilder = BindingTuple.builder();
        fields.forEach(field -> {
            final ExprValue exprValue = field.getExpr().valueOf(input);
            if (exprValue instanceof ExprMissingValue) {
                LOG.error(String.format("field:%s resolved as missing in context: %s", field, input));
            }
            outputBindingTupleBuilder.binding(field.getName(), exprValue);
        });
        return new BindingTupleRow(outputBindingTupleBuilder.build());
    }
}
```

FreCap: When running

```
PUT debug/_doc/1
{ 
  "region": "US",
  "manager": { 
    "age":     30,
    "name": "bob"
  }
}


PUT _opendistro/_sql/settings
{
  "transient" : {
    "opendistro.sql.engine.new.enabled" : "true"
  }
}


POST /_opendistro/_sql
{
  "query": """
    SELECT manager.name, count(*) FROM debug GROUP BY manager.name
  """
}
```

I get 
```
{
  "error": {
    "details": "invalid value operation on MISSING_VALUE",
    "type": "IllegalStateException"
  },
  "status": 500
}
```
and not the ideal:
```
{
  "error": {
    "details": "invalid value operation on MISSING_VALUE (manager#name), available: [COUNT_0, manager.name.keyword]",
    "type": "IllegalStateException"
  },
  "status": 500
}
```
How did you get `manager#name), available: [COUNT_0, manager.name.keyword]` with the current codebase? If I can get it, it'd be perfect and we could close the ticket



```
if (exprValue instanceof ExprMissingValue) {
                LOG.error(String.format("field:%s resolved as missing in context: %s", field, input));
            }
```
If we just log on disk but now throw, isn't it like swallowing the exception and have the user figure out that some data is missing?

I didn't fully get the answer why we should prefer logging on disk compared to adding the additional in-request logging.

penghuo: I apply your code change, then get the response, then I remove it now.

penghuo: @RequiredArgsConstructor
public class PhysicalProject implements PhysicalOperator<BindingTuple> {

    private static final Logger LOG = LogManager.getLogger(PhysicalProject.class);

    private final PhysicalOperator<BindingTuple> next;
    private final List<ColumnNode> fields;

    @Override
    public Cost estimate() {
        return null;
    }

    @Override
    public PlanNode[] children() {
        return new PlanNode[]{next};
    }

    @Override
    public boolean hasNext() {
        return next.hasNext();
    }

    @Override
    public Row<BindingTuple> next() {
        BindingTuple input = next.next().data();
        BindingTuple.BindingTupleBuilder outputBindingTupleBuilder = BindingTuple.builder();
        fields.forEach(field -> {
            final ExprValue exprValue = field.getExpr().valueOf(input);
            if (exprValue instanceof ExprMissingValue) {
//                LOG.error(String.format("field:%s resolved as missing in context: %s", field, input));
                throw new RuntimeException(String.format("field:%s resolved as missing in "
                    + "context: %s", field, input));
            }
            outputBindingTupleBuilder.binding(field.getName(), exprValue);
        });
        return new BindingTupleRow(outputBindingTupleBuilder.build());
    }
}

// output
{
  "error": {
    "details": "field:ColumnNode(name=manager.name, alias=null, type=TEXT, expr=manager#name) resolved as missing in context: <COUNT_0:1.0,manager.name.keyword:bob>",
    "type": "RuntimeException"
  },
  "status": 500
}

FreCap: ok, do you confirm I can move ahead adding the snippet you sent to the PR? 
`throw new RuntimeException(String.format("field:%s resolved as missing in "
                    + "context: %s", field, input));
`

penghuo: sure, thanks!

anirudha:
@FreCap any update from your side, thanks

chloe-zh (approved):

PR #974, opened 2021-01-07
Double quoted as string literal instead of identifier

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/975

*Description of changes:* Change grammar, AST builder and documentation to make double quoted treated as a string literal rather than an identifier.

*Documentation*:

1. https://github.com/dai-chen/sql/blob/use-double-quoted-as-string/docs/user/general/datatypes.rst#string-data-types
2. https://github.com/dai-chen/sql/blob/use-double-quoted-as-string/docs/user/general/identifiers.rst#id1

*Testing*: UT and IT changed accordingly.


penghuo (approved):
LGTM

chloe-zh (approved):

PR #983, opened 2021-01-11
[PPL] Support index name with date suffix

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/980

*Description of changes:*

1. Support index name with date suffix separated by dash or dot: Add an alternative for index name rule to avoid change on qualified name and thus impact on other identifiers, such as alias and field name.
2. Disable nested field rule in grammar: because it is not supported in core engine.

With the first change, the following index names are supported now. Please find more details in user manual doc below.

1. Date suffix: `filebeat-2020.01.11`
2. Multi date suffix separated by dash: `filebeat-7.10-2020.01.11`
3. Wildcard: `filebeat-2020.*` or `filebeat-2020.*.11`

*Documentation*: Add a new paragraph for this support for index name only. https://github.com/dai-chen/sql/blob/support-index-name-with-dots-in-ppl/docs/experiment/ppl/general/identifiers.rst#description

*Testing*: All old UT/IT passed. Add new UT/IT to verify different index names.

### DIFF
--- a/docs/experiment/ppl/general/identifiers.rst
+++ b/docs/experiment/ppl/general/identifiers.rst
@@ -30,6 +30,8 @@ For Elasticsearch, the following identifiers are supported extensionally:
 3. Identifiers with ``-`` in the middle: this is mostly the case for index name with date information.
 4. Identifiers with star ``*`` present: this is mostly an index pattern for wildcard match.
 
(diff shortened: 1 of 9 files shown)


chloe-zh (approved):
LGTM, thanks for the changes!

penghuo (approved):
Thanks for the change.

PR #760, opened 2020-09-28
Support security plugin in comparison test

dai-chen:
*Issue #, if available:*

*Description of changes:* Comparison test for new engine doesn't support testing against ODFE cluster with security plugin enabled. The following changes are made to fix this:

 1. Pass `https`, `user` and `password` parameter value to SQL JDBC driver to support security plugin.
 2. Change `CorrectnessTestBase` to inherit from `ODFERestTestCase`.
 3. Change `wipeAllIndices` to `wipeAllODFEIndices` to avoid permission issue: `SQLIntegTestCase` already have this fix in legacy module, however it's missing in the cloned `RestIntegTestCase`.

*Testing*:

Run ODFE 1.10.1 docker with security plugin enabled:

```
# Run docker with security plugin against OD version
# https://github.com/opendistro-for-elasticsearch/opendistro-build/blob/master/.github/scripts/setup_runners_service.sh#L122
➜ echo "FROM opendistroforelasticsearch/opendistroforelasticsearch:1.10.1" >> Dockerfile
➜ echo "RUN echo ''  >> /usr/share/elasticsearch/config/elasticsearch.yml" >> Dockerfile
➜ echo "RUN echo \"path.repo: [\\\"/usr/share/elasticsearch\\\"]\" >> /usr/share/elasticsearch/config/elasticsearch.yml" >> Dockerfile
➜ echo "RUN echo \"script.context.field.max_compilations_rate: 1000/1m\" >> /usr/share/elasticsearch/config/elasticsearch.yml" >> Dockerfile

➜ docker build -t odfe-http:security -f Dockerfile .
Sending build context to Docker daemon  2.048kB
Step 1/4 : FROM opendistroforelasticsearch/opendistroforelasticsearch:1.10.1
 ---> b3e80722f0c5
Step 2/4 : RUN echo ''  >> /usr/share/elasticsearch/config/elasticsearch.yml
 ---> Running in 0240221db08c
Removing intermediate container 0240221db08c
 ---> 4c4e94ead724
Step 3/4 : RUN echo "path.repo: [\"/usr/share/elasticsearch\"]" >> /usr/share/elasticsearch/config/elasticsearch.yml
 ---> Running in 09ed4624d11b
Removing intermediate container 09ed4624d11b
 ---> 7e0945ada901
Step 4/4 : RUN echo "script.context.field.max_compilations_rate: 1000/1m" >> /usr/share/elasticsearch/config/elasticsearch.yml
 ---> Running in 65b17fe267fe
[...]

PR #984, opened 2021-01-12
Bump jackson-databind version to 2.10.5.1

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
* Bump jackson-databind version to 2.10.5.1 to address CVE for sql
* Bump jackson-core version to 2.10.5
* Force integ-test to use jackson-core version 2.10.5
* Exclude jackson-core from :plugin 

### DIFF
--- a/plugin/build.gradle
+++ b/plugin/build.gradle
@@ -31,6 +31,7 @@ thirdPartyAudit.enabled = false
 configurations.all {
     // conflict with spring-jcl
     exclude group: "commons-logging", module: "commons-logging"
+    exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'
     // enforce 1.1.3, https://www.whitesourcesoftware.com/vulnerability-database/WS-2019-0379
     resolutionStrategy.force 'commons-codec:commons-codec:1.13'
     resolutionStrategy.force 'com.google.guava:guava:29.0-jre'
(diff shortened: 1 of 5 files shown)


plugin/build.gradle
```diff
@@ -31,6 +31,7 @@ thirdPartyAudit.enabled = false
 configurations.all {
     // conflict with spring-jcl
     exclude group: "commons-logging", module: "commons-logging"
+    exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'
```
dai-chen, line 34: I assume this is supposed to be same `resolutionStrategy.force` as above? Otherwise, although all IT passed, our plugin will have problem at runtime. Please confirm and do sanity test by running all ITs with remote ES cluster with our plugin built from your branch.

dai-chen: Reference: `./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername=es-integrationtest -Dtests.class="*SQLCorrectnessIT" -Dhttps=true -Duser=admin -Dpassword=admin` in https://github.com/opendistro-for-elasticsearch/sql/pull/760

chloe-zh (approved):

dai-chen (approved):
Thanks for the fix!

Issue #927, opened 2020-12-12
Handle the null literal in function


penghuo:
The null literal could as the arguments of the function. e.g.
```
SELECT 1 + null, cast(null as int)
```

dai-chen:
Related: https://github.com/opendistro-for-elasticsearch/sql/issues/715

Issue #715, opened 2020-08-26
[New Engine] Null literal uses unknown as value type


dai-chen:
Currently `NULL` literal is treated as unknown type which causes the following error.

```
POST _opendistro/_sql
{
  "query": "SELECT ABS(NULL)"
}
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "abs function expected {[INTEGER],[LONG],[FLOAT],[DOUBLE]}, but get [UNKNOWN]",
    "type": "ExpressionEvaluationException"
  },
  "status": 503
}
```

MySQL:

```
mysql> SELECT ABS(NULL);

+-----------+
| ABS(NULL) |
+-----------+
|      NULL |
+-----------+
1 row in set (0.12 sec)
```

dai-chen:
It seems `NULL` as function argument can be supported by simply adding it a new type (ex. `UNDEFINED` to distinguish from `UNKNOWN`) and assign it a bigger distance when function signature matching. Will confirm if this idea works or not.

PR #985, opened 2021-01-12
Support NULL literal as function argument

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/927, https://github.com/opendistro-for-elasticsearch/sql/issues/715

*Description of changes:* 

1. Add undefined type for NULL literal
2. Change JDBC driver to map undefined type to JDBC NULL type

*Documentation:* https://github.com/dai-chen/sql/blob/support-null-literal-as-function-argument/docs/user/general/datatypes.rst#undefined-data-type

*Design:* A new undefined data type is added as narrowest and root of data type tree.

1. `UNDEFINED` represents only the type of NULL literal for now (probably represent some other special literals in future)
2. `UNKNOWN` represents type not supported yet. For example, some new Elasticsearch type is given as function argument which yet to be added to core engine.

With this new type, we're able to support NULL literal. For example, given `ABS(NULL)`, the distance between NULL and INTEGER and DOUBLE are 1 and 2 respectively, so function signature ABS(INTEGER) is used to evaluate its value. Previously NULL literal's type is unknown which caused exception thrown.

null-literal

*Testing:* Add UT and IT. Because SQLite and H2 has different result type for expression with NULL literal, a `NullLiteralIT` is added manually.

### DIFF
--- a/docs/user/general/values.rst
+++ b/docs/user/general/values.rst
@@ -11,7 +11,7 @@ Data Types
 
 NULL and MISSING Values
 =======================
-ODFE SQL has two ways to represent missing information. (1) The presence of the field with a NULL for its value. and (2) the absence of the field.
+ODFE SQL has two ways to represent missing information. (1) The presence of the field with a NULL for its value. and (2) the absence of the filed.
 
 Please note, when response is in table format, the MISSING value is translate to NULL value.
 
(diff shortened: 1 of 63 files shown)


docs/user/general/values.rst
```diff
@@ -46,6 +46,21 @@ Here is an example::
...
+
+NULL Literal Handling
+---------------------
+
+Because the type of a null literal is unknown, a special data type ``UNDEFINED`` is reserved to allow it be accepted as a valid function argument::
+
+    od> SELECT NULL, NULL = NULL, 1 + NULL, LENGTH(NULL);
+    fetched rows / total rows = 1/1
+    +--------+---------------+------------+----------------+
+    | NULL   | NULL = NULL   | 1 + NULL   | LENGTH(NULL)   |
```
harold-wang, line 58: 1 + NULL should 1 or NULL?
LENGTH(NULL) should be NULL?

dai-chen: yes, it should be what is documented here. Line 34 explains this. Unless the operator/function is special, expression involving NULL will end with NULL.

chloe-zh (approved):
LGTM, thanks for the changes!

harold-wang (approved):
Thanks for the change!

Issue #861, opened 2020-11-27
The IF function with nested locate function doesn't work in Query Workbench


nicolaslledo:
The query 

`SELECT 
IF(LOCATE('o', 'hello')>0, 'found', 'no found') AS ifTrue
FROM  siouxsso* limit 10;`

results in ": no response" the result pane

chloe-zh:
Hi @nicolaslledo , thanks for reporting this issue! The old engine does not support nested functions in many cases, neither string functions nor conditional functions, even though both `locate` and `if` work fine separately. This issue results from the implementation of the functions in the old engine is to push the hard coded painless script to the es engine, but it is ineligible when the functions are nested. The good thing is our new engine is eligible fo nested functions, and the case you mentioned would also be supported once the `IF` and `LOCATE` are supported. We will keep you updated when the features are good to release. Thanks!

dai-chen:
IF function is already added in new query engine. Locate function is pending.

PR #990, opened 2021-01-14
Add Flow control function IF(expr1, expr2, expr3)

harold-wang:
*Issue #861 , if available:*

*Description of changes:*

Add flow control function IF(expr1, expr2, expr3)


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -149,4 +163,8 @@ public static ExprValue exprNullIf(ExprValue v1, ExprValue v2) {
     return v1.equals(v2) ? LITERAL_NULL : v1;
   }
 
+  public static ExprValue exprIf(ExprValue v1, ExprValue v2, ExprValue v3) {
+    return v1.value().equals(LITERAL_TRUE.value()) ? v2 : v3;
```
dai-chen, line 167: For `v1`, I think we need to do null/missing check? Please check MySQL behavior and if ours is the same.

And `ExprBooleanValue` already implements `equals()` so just do `LITERAL_TRUE.equals(v1)`?

docs/user/dql/functions.rst
```diff
@@ -1982,6 +1982,40 @@ Example::
...
+------
+
+Description
+>>>>>>>>>>>
+
+Specifications:
+
+1. IF(ES_TYPE0, ES_TYPE1, ES_TYPE2) -> ES_TYPE1 or ES_TYPE2
+
+Usage: return first parameter is true, return second parameter, otherwise return third one.
```
dai-chen, line 1995: np: "`if` first parameter is true ..."

dai-chen:
A question: should the function signature register with `UNKNOWN` be removed too? https://github.com/opendistro-for-elasticsearch/sql/pull/990/files#diff-ee3435d38230141befb55d9dce13674513b55496848176905e88e44ecaeffea5R126-R129

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
harold-wang: Thanks @dai-chen, you are correct, we need to do null/missing check. let me update both.

https://www.w3resource.com/mysql/control-flow-functions/if-function.php
IF() function
MySQL IF() takes three expressions and if the first expression is true, not zero and not NULL, it returns the second expression. Otherwise, it returns the third expression.

Depending on the context in which it is used, it returns either numeric or string value.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
```diff
@@ -100,6 +99,25 @@ private static FunctionResolver isNotNull() {
...
+
+    List<SerializableFunction<FunctionName, org.apache.commons.lang3.tuple.Pair<FunctionSignature,
+            FunctionBuilder>>> functionsOne = typeList.stream().map(v ->
+            impl((UnaryPredicateOperator::exprIf), v, BOOLEAN, v, v))
+            .collect(Collectors.toList());
+
+    List<SerializableFunction<FunctionName, org.apache.commons.lang3.tuple.Pair<FunctionSignature,
+            FunctionBuilder>>> functionsTwo = typeList.stream().map(v ->
+            impl((UnaryPredicateOperator::exprIf), v, UNKNOWN, v, v))
+            .collect(Collectors.toList());
```
dai-chen, line 114: Same as my previous question for line 130-133, why is this required?

docs/user/dql/functions.rst
```diff
@@ -1982,6 +1982,40 @@ Example::
     | True          | False         |
     +---------------+---------------+
 
+IF
+------
```
dai-chen, line 1986: could you update PPL function doc too?

docs/user/dql/functions.rst
```diff
@@ -1982,6 +1982,40 @@ Example::
...
+
+Description
+>>>>>>>>>>>
+
+Specifications:
+
+1. IF(ES_TYPE0, ES_TYPE1, ES_TYPE2) -> ES_TYPE1 or ES_TYPE2
```
penghuo, line 1993: Is there any reason we use ES_TYPE0/1/2? In my opinion, IF(expr, expr1, expr2) is more clear definition?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/operator/predicate/UnaryPredicateOperator.java
penghuo: could we consider the solution which could for case...when and if. i think they both control flow statement which may not fit into our function defintion.

harold-wang: Good idea.

dai-chen (approved):

penghuo (approved):

harold-wang:
Todo issue: merge if function with case when function.

PR #998, opened 2021-01-21
Add RAW format in new engine

harold-wang:
*Issue #997 , if available:*

*Description of changes:*
Added raw formatter FlatResponseFormatter in new engine. The formatter formats the QueryResult to raw format and by default it sanitizes the result by (1) inserting single-quote in front of header/data if it starts with special characters including +, -, = and @; (2) quoting the header/data if one or more pipe charater (|) in the header/data. The sanitizing can be skipped by adding a sanitize param and set it to false value in the request.

User manual updates: https://github.com/harold-wang/sql/blob/raw/docs/user/interfaces/protocol.rst

Combine the CsvResponseFormatter and raw format into FlatResponseFormatter


protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FlatResponseFormatter.java
```diff
@@ -26,44 +26,44 @@
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 
-@RequiredArgsConstructor
-public class CsvResponseFormatter implements ResponseFormatter<QueryResult> {
-  private static final String INLINE_SEPARATOR = ",";
+//@RequiredArgsConstructor
+public class FlatResponseFormatter implements ResponseFormatter<QueryResult> {
```
penghuo: Could you add some java doc? majorly what is Flat means?

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FlatResponseFormatter.java
```diff
@@ -26,44 +26,44 @@
...
+  private static String separator = ",";
```
penghuo: why static? does it need default value ","?

penghuo:
Could also refine the docs?
1. add protocl doc for PPL.
2. differenate the RAW and CSV format? what is common? what is special?

plugin/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -196,7 +196,9 @@ public void onFailure(Exception e) {
     Format format = pplRequest.format();
     ResponseFormatter<QueryResult> formatter;
     if (format.equals(Format.CSV)) {
-      formatter = new CsvResponseFormatter(pplRequest.sanitize());
+      formatter = new FlatResponseFormatter(",", pplRequest.sanitize());
+    } else if (format.equals(Format.RAW)) {
+      formatter = new FlatResponseFormatter("|", pplRequest.sanitize());
```
dai-chen: Could you check if raw data need sanitize? Should it be either always false or at least false by default? Because I think it would be a surprise to users if we sanitize raw data by default.

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FlatResponseFormatter.java
```diff
@@ -100,12 +109,12 @@ private String getHeaderLine(QueryResult response, boolean sanitize) {
...
 
     /**
-     * Sanitize CSV headers because Elasticsearch allows special character present in field names.
+     * Sanitize headers because Elasticsearch allows special character present in field names.
      */
     private List<String> sanitizeHeaders(List<String> headers) {
       return headers.stream()
           .map(this::sanitizeCell)
-          .map(cell -> quoteIfRequired(INLINE_SEPARATOR, cell))
+          .map(cell -> quoteIfRequired(separator, cell))
```
dai-chen: For now both sanitize and escape logic is controlled by `sanitize` field, is this why we want `escape=true` by default even for RAW formatter? Should we always do escape despite of `escape`?

plugin/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/rest/RestPPLQueryAction.java
harold-wang: Feel that comma, tab, colon, or vertical bar(pipe) are just all column delimiter, that give me believe that sanitize should be applied to raw data, but need to confirm from document. Tested in excel that sanitize is implemented, but could not test raw format. Let me search around more.

1. https://en.wikipedia.org/wiki/Delimiter-separated_values
2. https://docs.microsoft.com/en-us/sql/integration-services/import-export-data/configure-flat-file-destination-sql-server-import-and-export-wizard?view=sql-server-ver15

harold-wang: For RAW format: character '\' is put before the '|' character, as same as csv I think
=============Data Base===============
mysql> select * from tutorials_tbl;
+-------------+----------------+-----------------+-----------------+
| tutorial_id | tutorial_title | tutorial_author | submission_date |
+-------------+----------------+-----------------+-----------------+
|           1 | Learn PHP      | John Poul       | 2021-01-25      |
|           2 | Learn ,,,,PHP  | John||| Poul    | 2021-01-25      |
+-------------+----------------+-----------------+-----------------+
2 rows in set (0.00 sec)

=============export to text file limited by vertical bar ===============
mysql> SELECT * FROM tutorials_tbl
    -> INTO OUTFILE '/tmp/cancelled_orders2.csv' 
    -> FIELDS ENCLOSED BY '|' 
    -> LINES TERMINATED BY '\r\n';
Query OK, 2 rows affected (0.00 sec)

=============\ is put before the | character in field ===============
f8ffc257e573:tmp harowang$ sudo more  cancelled_orders2.csv
|1|     |Learn PHP|     |John Poul|     |2021-01-25|
|2|     |Learn ,,,,PHP| |John\|\|\| Poul|       |2021-01-25|

============= \ is put before the , character in field ===============
f8ffc257e573:tmp harowang$ sudo more  cancelled_orders3.csv
Password:
,1,     ,Learn PHP,     ,John Poul,     ,2021-01-25,
,2,     ,Learn \,\,\,\,PHP,     ,John||| Poul,  ,2021-01-25,

docs/user/interfaces/protocol.rst
```diff
@@ -355,9 +356,43 @@ SQL query::
...
 	Dale|Adams|33
 	Hattie|Bond|36
-	
 
+
+The formatter sanitizes the raw result with the following rules:
+
+1. If a header cell or data cell is starting with special character including '+', '-', '=' , '@', the sanitizer will insert a single-quote at the start of the cell.
+
+2. If there exists one or more commas ('|'), the sanitizer will quote the cell with double quotes.
```
chloe-zh: commas -> pipes ?

docs/user/interfaces/protocol.rst
```diff
@@ -355,9 +356,43 @@ SQL query::
...
+For example::
+
+    >> curl -H 'Content-Type: application/json' -X PUT localhost:9200/userdata/_doc/1?refresh=true -d '{
+      "+firstname": "-Hattie",
+      "=lastname": "@Bond",
+      "address": "671 Bristol Street|, Dente, TN"
+    }'
+	>> curl -H 'Content-Type: application/json' -X POST localhost:9200/_opendistro/_sql?format=csv -d '{
+	  "query" : "SELECT firstname, lastname, address FROM userdata"
+	}'
```
chloe-zh, line 379: np: align the lines?

plugin/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/rest/RestPPLQueryAction.java
```diff
@@ -197,6 +198,8 @@ public void onFailure(Exception e) {
     ResponseFormatter<QueryResult> formatter;
     if (format.equals(Format.CSV)) {
       formatter = new CsvResponseFormatter(pplRequest.sanitize());
+    } else if (format.equals(Format.RAW)) {
+      formatter = new RawResponseFormatter(pplRequest.sanitize());
```
chloe-zh: Should we let the raw format always be sanitized instead?

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/RawResponseFormatter.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.stream.Collectors;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Response formatter to format response to csv or raw format.
+ */
+//@RequiredArgsConstructor
+public class RawResponseFormatter extends FlatResponseFormatter {
+  private boolean sanitize = false;
```
chloe-zh: Unnecessary local variable `sanitize`, remove?

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/RawResponseFormatter.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Response formatter to format response to csv or raw format.
+ */
+//@RequiredArgsConstructor
```
chloe-zh, line 32: remove?

protocol/src/test/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FormatTest.java
```diff
@@ -48,6 +48,15 @@ void csv() {
...
 
+  /*
+  @Test
+  void raw() {
+    Optional<Format> format = Format.of("raw");
+    assertTrue(format.isPresent());
+    assertEquals(Format.RAW, format.get());
+  }
+
+   */
```
chloe-zh: uncomment the test?

chloe-zh (approved):
Some minor changes, the rest looks good to me. Thanks!

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FlatResponseFormatter.java
dai-chen: Make the base class abstract if possible

plugin/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/rest/RestPPLQueryAction.java
dai-chen: I recall as discussed we need to separate sanitize and escape logic in `FlatResponseFormatter`.

1. For CSV, always do escape for `,`, sanitize if parameter `sanitize` given
2. For RAW, always do escape for `|`, but **never sanitize**. Because: a) that sanitize logic only make sense for Excel; b) assumption is that users want the data exported as much original as possible.

docs/user/interfaces/protocol.rst
harold-wang: correct.

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/RawResponseFormatter.java
harold-wang: yes

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/RawResponseFormatter.java
harold-wang: agree, removed

plugin/src/main/java/com/amazon/opendistroforelasticsearch/sql/plugin/rest/RestPPLQueryAction.java
harold-wang: @dai-chen implemented "For RAW, always do escape for |, but never sanitize. "

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FlatResponseFormatter.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public abstract class FlatResponseFormatter implements ResponseFormatter<QueryResult> {
```
dai-chen, line 30: Could you make `CSVResponseFormatter` subclass? Otherwise there are many duplicated code in both class.

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/RawFormatIT.java
```diff
@@ -0,0 +1,78 @@
...
+
+public class RawFormatIT extends SQLIntegTestCase {
+
+  @Override
+  public void init() throws IOException {
+    loadIndex(Index.BANK_RAW_SANITIZE);
+  }
+
+  @Test
+  public void sanitizeTest() {
```
dai-chen: Could you rename it? because RAW formatter doesn't do sanitize. To avoid confusion, probably you just need a test index with pipe in field value.

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/FlatResponseFormatter.java
harold-wang: @dai-chen let me make CSVResponseFormatter and address your other comments as well, thanks

integ-test/src/test/java/com/amazon/opendistroforelasticsearch/sql/sql/RawFormatIT.java
harold-wang: Thanks, done

docs/user/interfaces/protocol.rst
```diff
@@ -355,9 +356,41 @@ SQL query::
...
+
+Result set::
+
+    '+firstname|'=lastname|address
+    'Hattie|'@Bond|"671 Bristol Street|, Dente, TN"
```
dai-chen: Since we don't do sanitize for RAW, the 2nd column is supposed to be `@Bond` rather than `'@Bond`?

And the rest of the doc should be removed?

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/CsvResponseFormatter.java
```diff
@@ -26,119 +26,14 @@
...
-      return cell.contains(separator)
-          ? quote + cell.replaceAll("\"", "\"\"") + quote : cell;
-    }
-
-    private boolean isStartWithSensitiveChar(String cell) {
-      return SENSITIVE_CHAR.stream().anyMatch(cell::startsWith);
-    }
+  public CsvResponseFormatter(boolean sanitize) {
+    super(",", sanitize);
+    System.out.println("Harold sanitize : " + sanitize);
```
dai-chen: Could you remove all the sysout? There is another one at line 44 in `FlatResponseFormatter`

dai-chen (approved):
Just 2 minor comments for doc and sysout. Thanks for the changes!

protocol/src/main/java/com/amazon/opendistroforelasticsearch/sql/protocol/response/format/CsvResponseFormatter.java
harold-wang: Sorry, removed.

dai-chen: Thanks! 👍

docs/user/interfaces/protocol.rst
harold-wang: Done, thanks for quality control!

PR #1001, opened 2021-01-22
Fetch error message in root cause for new default formatter

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/1002

*Description of changes:* Reuse existing `ErrorMessage` and `ErrorMessageFactory` to extract root cause info from Elasticsearch search. After changed, the error response is exactly same as that of old engine and PPL.

*Testing*: UT and manual verification as below

Without the fix:

```
PUT _settings
{
  "index.max_result_window": 100
}

POST _opendistro/_sql/
{
  "query": """
    SELECT * FROM accounts
  """
}
{
  "error": {
    "type": "SearchPhaseExecutionException",
    "reason": "all shards failed",
    "details": "all shards failed"
  },
  "status": 400
}
```

After the fix:

```
{
  "error": {
    "type": "SearchPhaseExecutionException",
    "reason": "Error occurred in Elasticsearch engine: all shards failed",
    "details": "Shard[0]: java.lang.IllegalArgumentException: Result window is too large, from + size must be less than or equal to: [100] but was [200]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.\n\nFor more details, please send request for Json format to see the raw response from elasticsearch engine."
  },
  "status": 400
}
```

### DIFF
--- a/elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/response/error/ErrorMessage.java
+++ b/elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/response/error/ErrorMessage.java
@@ -28,17 +28,16 @@ public class ErrorMessage {
 
   protected Throwable exception;
 
-  @Getter
-  private int status;
+  private final int status;
 
   @Getter
-  private String type;
+  private final String type;
 
   @Getter
-  private String reason;
+  private final String reason;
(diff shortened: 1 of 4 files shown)


penghuo (approved):
LGTM

chloe-zh (approved):

PR #1010, opened 2021-01-26
Rename odbc release artifacts

joshuali925:
*Issue #, if available:*

*Description of changes:*
rename odbc artifacts before uploading to S3 from
```
Open Distro for Elasticsearch SQL ODBC Driver 64-bit-1.12.0.0-Darwin.pkg
Open Distro for Elasticsearch SQL ODBC Driver 64-bit-1.12.0.0-Windows.msi
Open Distro for Elasticsearch SQL ODBC Driver 32-bit-1.12.0.0-Windows.msi
```
to
```
opendistro-sql-odbc-1.12.0.x-macos-x64.pkg
opendistro-sql-odbc-1.12.0.x-windows-x64.msi
opendistro-sql-odbc-1.12.0.x-windows-x86.msi
```

### DIFF
--- a/.github/workflows/sql-odbc-release-workflow.yml
+++ b/.github/workflows/sql-odbc-release-workflow.yml
@@ -11,6 +11,8 @@ env:
   ODBC_BIN_PATH: "./build/odbc/bin"
   ODBC_BUILD_PATH: "./build/odbc/build"
   AWS_SDK_INSTALL_PATH: "./build/aws-sdk/install"
+  PLUGIN_NAME: opendistro-sql-odbc
+  OD_VERSION: 1.12.0.0
 
 jobs:
   build-mac:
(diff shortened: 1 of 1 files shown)


davidcui-amzn (approved):

dai-chen (approved):
I was thinking generate the artifact with the expected name directly when ODBC build. But I found many reference to old name in Github action and documentation. Is that the reason we'd rather rename it at the final step?

joshuali925:
> I was thinking generate the artifact with the expected name directly when ODBC build. But I found many reference to old name in Github action and documentation. Is that the reason we'd rather rename it at the final step?

Yes, peter suggested it could be easier to rename after it builds than changing build params, since I'm not very familiar with cmake

Issue #991, opened 2021-01-14
Interval expression check and evaluation issue


dai-chen:
After undefined type introduced in PR https://github.com/opendistro-for-elasticsearch/sql/pull/985, two interval expression issues are found in the following query.

```
POST _opendistro/_sql
{
  "query": """
   SELECT INTERVAL 60 * 60 * 24 * (NULL - FLOOR(NULL)) SECOND AS `TEMP(Test)(2744314424)(0)` FROM `calcs` HAVING (COUNT(1) > 0)
  """
}
```

1. Non-aggregate SELECT item check failed:

```
{
  "error": {
    "reason": "Invalid SQL query",
    "details": "Explicit GROUP BY clause is required because expression [Interval(value=*(*(*(60, 60), 24), -(null, FLOOR(null))), unit=SECOND)] contains non-aggregated column",
    "type": "SemanticCheckException"
  },
  "status": 400
}
```

2. If we bypass the check above, exception is thrown when interval value evaluation because input is undefined type.

```
[2021-01-14T14:56:18,909][INFO ][c.a.o.s.l.p.RestSqlAction] [186590df9563.ant.amazon.com] [06ce0988-5b76-4924-8429-8ce2f1c00966] Request is handled by new SQL query engine
[2021-01-14T14:56:18,929][ERROR][c.a.o.s.l.p.RestSQLQueryAction] [186590df9563.ant.amazon.com] Error happened during query handling
com.amazon.opendistroforelasticsearch.sql.exception.ExpressionEvaluationException: invalid to get longValue from value of type UNDEFINED
	at com.amazon.opendistroforelasticsearch.sql.data.model.ExprValue.longValue(ExprValue.java:109) ~[core-1.12.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.data.model.ExprValueUtils.getLongValue(ExprValueUtils.java:164) ~[core-1.12.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.expression.datetime.IntervalClause.second(IntervalClause.java:92) ~[core-1.12.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.expression.datetime.IntervalClause.interval(IntervalClause.java:66) ~[core-1.12.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.expression.function.FunctionDSL$3.valueOf(FunctionDSL.java:172) ~[core-1.12.0.0.jar:?]
[...]

PR #1011, opened 2021-01-26
Fixed interval type null/missing check failure

chloe-zh:
*Issue #, if available:*
#991 

*Description of changes:*
- Interval type failed semantic check in aggregations with implicit group by, so added interval type as literal type to be non-aggregation select items to pass the check
- Handled null and missing value check for interval type
- Added corresponding tests. Note: didn't add comparison test because missing `INTERVAL` in JDBC types

### DIFF
--- a/sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/AstAggregationBuilder.java
+++ b/sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/AstAggregationBuilder.java
@@ -22,6 +22,7 @@ import com.amazon.opendistroforelasticsearch.sql.ast.Node;
 import com.amazon.opendistroforelasticsearch.sql.ast.expression.AggregateFunction;
 import com.amazon.opendistroforelasticsearch.sql.ast.expression.Alias;
 import com.amazon.opendistroforelasticsearch.sql.ast.expression.Function;
+import com.amazon.opendistroforelasticsearch.sql.ast.expression.Interval;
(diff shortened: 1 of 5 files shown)


sql/src/main/java/com/amazon/opendistroforelasticsearch/sql/sql/parser/AstAggregationBuilder.java
```diff
@@ -136,7 +137,8 @@ private boolean isAggregatorNotFoundAnywhere() {
 
   private boolean isNonLiteralFunction(UnresolvedExpression expr) {
     // The base case for recursion
-    if (expr instanceof Literal) {
+    // Interval is included since it is a special case of literal
+    if (expr instanceof Literal || expr instanceof Interval) {
```
dai-chen, line 141: Could you try out the workaround here? https://github.com/opendistro-for-elasticsearch/sql/issues/991#issuecomment-760542246

Because I realized what we really want to check here is **if any qualified name (field) not in aggregate function call**. But here we're checking it in the opposite way which may miss some cases. For example, CAST is not covered.

chloe-zh: I see. Changed to the `isNonAggregateOrLiteralExpression` recursion, and added `cast` and `case when` cases to unit test

dai-chen (approved):
Thanks for the fix!

penghuo (approved):
LGTM

Issue #1004, opened 2021-01-25
DELETE command: an stack trace is logged when executing a DELETE command


fbaligand:
If I run this command:
`DELETE index1 WHERE field1='value1';`

The DELETE command is well executed, and expected documents are deleted.

But I get this error (with stack trace) in the logs:
```
[2021-01-25T09:12:34,669][ERROR][c.a.o.s.l.u.QueryDataAnonymizer] [DESKTOP-8QK15CF] Caught an exception when removing sensitive data
com.alibaba.druid.sql.parser.ParserException: ERROR. token : DELETE, pos : 8
	at com.alibaba.druid.sql.parser.SQLExprParser.primary(SQLExprParser.java:555) ~[druid-1.0.15.jar:1.0.15]
	at com.amazon.opendistroforelasticsearch.sql.legacy.parser.ElasticSqlExprParser.primary2(ElasticSqlExprParser.java:305) ~[legacy-1.8.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.legacy.parser.ElasticSqlExprParser.primary(ElasticSqlExprParser.java:199) ~[legacy-1.8.0.0.jar:?]
	at com.alibaba.druid.sql.parser.SQLExprParser.expr(SQLExprParser.java:121) ~[druid-1.0.15.jar:1.0.15]
	at com.amazon.opendistroforelasticsearch.sql.legacy.utils.Util.toSqlExpr(Util.java:270) ~[legacy-1.8.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.legacy.utils.QueryDataAnonymizer.anonymizeData(QueryDataAnonymizer.java:44) [legacy-1.8.0.0.jar:?]
	at com.amazon.opendistroforelasticsearch.sql.legacy.plugin.RestSqlAction.prepareRequest(RestSqlAction.java:147) [legacy-1.8.0.0.jar:?]
	at org.elasticsearch.rest.BaseRestHandler.handleRequest(BaseRestHandler.java:94) [elasticsearch-7.9.3.jar:7.9.3]
	at tech.beshu.ror.es.ReadonlyRestPlugin.$anonfun$getRestHandlerWrapper$2(ReadonlyRestPlugin.scala:228) [readonlyrest-1.26.1_es7.9.3.jar:?]
	at org.elasticsearch.rest.RestController.dispatchRequest(RestController.java:236) [elasticsearch-7.9.3.jar:7.9.3]
[...]

Issue #431, opened 2020-04-14
Exception while masking/anonymizing SQL queries


abbashus:
Both `SHOW` and `DESCRIBE` are throwing exceptions.

SQL Plugin: commit `bf4810dc26cc1782ab433bc97d63a6a0b326da11`

elasticserach.log
```
[2020-04-14T11:00:17,731][ERROR][c.a.o.s.u.QueryDataAnonymizer] [f45c898a2ba7.ant.amazon.com] Caught an exception when removing sensitive data
com.alibaba.druid.sql.parser.ParserException: ERROR. token : SHOW, pos : 4
	at com.alibaba.druid.sql.parser.SQLExprParser.primary(SQLExprParser.java:555) ~[druid-1.0.15.jar:1.0.15]
	at com.amazon.opendistroforelasticsearch.sql.parser.ElasticSqlExprParser.primary2(ElasticSqlExprParser.java:305) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.parser.ElasticSqlExprParser.primary(ElasticSqlExprParser.java:199) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.alibaba.druid.sql.parser.SQLExprParser.expr(SQLExprParser.java:121) ~[druid-1.0.15.jar:1.0.15]
	at com.amazon.opendistroforelasticsearch.sql.utils.Util.toSqlExpr(Util.java:270) ~[opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.utils.QueryDataAnonymizer.anonymizeData(QueryDataAnonymizer.java:44) [opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at com.amazon.opendistroforelasticsearch.sql.plugin.RestSqlAction.prepareRequest(RestSqlAction.java:130) [opendistro_sql-1.6.0.0.jar:1.6.0.0]
	at org.elasticsearch.rest.BaseRestHandler.handleRequest(BaseRestHandler.java:87) [elasticsearch-7.6.1.jar:7.6.1]
	at org.elasticsearch.rest.RestController.dispatchRequest(RestController.java:227) [elasticsearch-7.6.1.jar:7.6.1]
	at org.elasticsearch.rest.RestController.tryAllHandlers(RestController.java:309) [elasticsearch-7.6.1.jar:7.6.1]
[...]

PR #1014, opened 2021-01-27
Removed stack trace when catches anonymizing data error

chloe-zh:
*Issue #, if available:*
#1004 #431 

*Description of changes:*
Removed the stack trace for query data anonymizing error since it causes confusion for users who execute metadata queries and dml queries. Now it only prints a warning in log with error message "Caught an exception when anonymizing sensitive data" as follows:

Giving query `show tables like '%'`
```
[2021-01-27T14:31:29,304][WARN ][c.a.o.s.l.u.QueryDataAnonymizer] [147dda5ddf8a.ant.amazon.com] Caught an exception when anonymizing sensitive data
[2021-01-27T14:31:29,305][INFO ][c.a.o.s.l.p.RestSqlAction] [147dda5ddf8a.ant.amazon.com] [a8a76a90-8b90-449e-b50c-44cd954ee851] Incoming request /_opendistro/_sql?pretty=true: 
  show tables like '%'
```


dai-chen (approved):
Thanks for the fix!

penghuo (approved):

PR #1015, opened 2021-01-27
Rename sql-cli wheel to use dashes instead of underscore

joshuali925:
*Issue #, if available:*

*Description of changes:*
change `odfe_sql_cli-1.12.0.0-py3-none-any.whl` to `odfe-sql-cli-1.12.0.0-py3-none-any.whl` when uploading to s3 for infra team's script

### DIFF
--- a/.github/workflows/sql-cli-release-workflow.yml
+++ b/.github/workflows/sql-cli-release-workflow.yml
@@ -42,9 +42,11 @@ jobs:
           python setup.py sdist bdist_wheel
           artifact=`ls ./dist/*.tar.gz`
           wheel_artifact=`ls ./dist/*.whl`
+          renamed_wheel_artifact=`echo $wheel_artifact | sed 's/_/-/g'`
+          mv "$wheel_artifact" "$renamed_wheel_artifact"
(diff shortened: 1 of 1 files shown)


zhongnansu:
When we upload this artifact to PyPI, what name will we use? Did you try installing from the wheel file? `pip install --use-wheel`, is there any error? 
Ref： https://github.com/pypa/pip/issues/1150#issue-18278500

joshuali925:
> When we upload this artifact to PyPI, what name will we use? Did you try installing from the wheel file? `pip install --use-wheel`, is there any error?

I think `--use-wheel` is deprecated. I've tried `pip install odfe-sql-cli-1.12.0.0-py3-none-any.whl` and did some sanity tests, didn't see errors. Also this doesn't affect release as we won't be pushing to PyPI from S3. When we push it will still be `odfe_sql_cli-1.12.0.0-py3-none-any.whl` locally. This is only for infra team's scripts to work.

zhongnansu (approved):

davidcui-amzn (approved):

Issue #1017, opened 2021-01-28
[Enhancement] Handle the multivalue field properly


penghuo:
Currently, we the field has [multiple value](https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html), SQL/PPL can't handle it properly.
```
PUT objectsit/_doc/1
{
  "name": ["test1", "test2"],
  "ids": [1, 2],
  "accounts": [
    {
      "id": 1
    },
    {
      "id": 2
    }
  ]
}
POST /_opendistro/_sql
{
  "query": """
    SELECT name, ids, accounts
    FROM objectsit
  """
}

{
  "schema": [
    {
      "name": "name",
      "type": "text"
    },
    {
      "name": "ids",
      "type": "long"
    },
    {
      "name": "accounts",
      "type": "object"
    }
  ],
  "datarows": [
    [
      "",
      0,
      {}
    ]
  ],
  "total": 1,
  "size": 1,
  "status": 200
}
```

Issue #795, opened 2020-10-22
myIndex.field.subFieldA > myIndex.field.subFieldB generates the wrong painless (doc['subFieldA'].value > doc['subFieldB'].value)


FreCap:
```
POST _opendistro/_sql/_explain
{
  "query": """SELECT (i.field1.subFieldA) FROM
  myIndex as i 
  WHERE 
  i.field1.subFieldA >i.field1.subFieldB
    LIMIT 50;"""
}
```

generates

```
{
  "from" : 0,
  "size" : 50,
  "query" : {
    "bool" : {
      "filter" : [
        {
          "bool" : {
            "must" : [
              {
                "script" : {
                  "script" : {
                    "source" : "doc['subFieldA'].value > doc['subFieldB'].value", ### HERE IS WRONG! I suppose it should be field field1.subFieldA instead of just subFieldA
                    "lang" : "painless"
                  },
                  "boost" : 1.0
                }
              }
            ],
            "adjust_pure_negative" : true,
            "boost" : 1.0
          }
        }
      ],
      "adjust_pure_negative" : true,
      "boost" : 1.0
    }
  },
  "_source" : {
    "includes" : [
      "Alpha.ID"
    ],
    "excludes" : [ ]
  }
}
```

The source of painless is wrong, since it doesn't reference anywhere `field1`, and therefore it doesn't find the field and fails

```
{
  "error": {
    "reason": "Error occurred in Elasticsearch engine: all shards failed",
    "details": """Shard[0]: ScriptException[runtime error]; nested: IllegalArgumentException[No field found for [subFieldA] in mapping with types []];
Shard[1]: ScriptException[runtime error]; nested: IllegalArgumentException[No field found for [subFieldA] in mapping with types []];
Shard[2]: ScriptException[runtime error]; nested: IllegalArgumentException[No field found for [subFieldA] in mapping with types []];
Shard[3]: ScriptException[runtime....
```

Can anyone point me to the code that generates this?

EDIT: opendistro version: 1.10.1.1
```
 wget  https://d3g5vo6xdbdb9a.cloudfront.net/downloads/elasticsearch-plugins/opendistro-sql/opendistro_sql-1.10.1.1.zip  
 bin/elasticsearch-plugin install ./openSQL/opendistro_sql-1.10.1.1.jar 
 ```

dai-chen:
Hi @FreCap , thanks for reporting the issue! Could you post your index mapping for our investigation?

[...]

PR #1018, opened 2021-01-28
Support Struct Data Query In SQL/PPL

penghuo:
*Issue #, if available:* #795

*Description of changes:*
1. Support Struct Data Query In SQL/PPL. https://github.com/penghuo/sql/blob/object-query-pr/docs/experiment/ppl/general/datatypes.rst#query-struct-data-types
2. Add the Paths in ReferenceExpression.

## To Reviewer
There are 2 ITs in legacy engine have been disabled.
1. ObjectFieldSelectIT.testSelectObjectFieldOfArrayValuesItself, the issue is tracking by #1017 
2. PrettyFormatResponseIT. the issue is tracking by #1019


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/ReferenceExpression.java
```diff
@@ -49,4 +69,44 @@ public ExprType type() {
   public String toString() {
     return attr;
   }
+
+  /**
+   * Resolve the ExprValue from {@link ExprTupleValue} using paths.*
```
dai-chen, line 74: np: is `paths.*` intentional or a typo?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/ReferenceExpression.java
```diff
@@ -49,4 +69,44 @@ public ExprType type() {
...
+   * 2. Resolve the path recursively through ExprValue.
+   *
+   * @param value {@link ExprTupleValue}.
+   * @return {@link ExprTupleValue}.
+   */
+  public ExprValue resolve(ExprTupleValue value) {
+    return resolve(value, paths);
+  }
+
+  private ExprValue resolve(ExprValue value, List<String> paths) {
```
dai-chen, line 104: Can this / do we need to handle `project.year` is another object field? ex. `project.year`.month

dai-chen (approved):
Just minor comments, thanks for the changes!

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/ReferenceExpression.java
penghuo: typo. removed.

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/expression/ReferenceExpression.java
penghuo: Good point. add more cases in doc and UT.
We don't expect object filed name include "." which is not supported now.

chloe-zh (approved):

Issue #930, opened 2020-12-14
Refactoring PPL plans in Head and RareTopN commands


chloe-zh:
Extract `Limit` from `Head` and `RareTopN`

PR #1022, opened 2021-01-29
Refine PPL head command syntax

chloe-zh:
*Issue #, if available:*
#930 

*Description of changes:*

- Modified `head` command syntax to simply `head [N]` where the optional argument N is the number of results to return
- Removed head operator and replaced the original use of head operator with limit operator
- Updated user manual

**Breaking Changes Note:** We have made breaking changes in the syntax of ``head`` command from previous ``head [keeplast = (true | false)] [while "("<boolean-expression>")"] [N]`` to current ``head [N]``. We are no longer supporting the ``keeplast`` and ``while`` options.


core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/tree/Head.java
```diff
@@ -38,7 +38,7 @@
 public class Head extends UnresolvedPlan {
```
penghuo, line 38: Merge with Limit?

ppl/src/main/antlr/OpenDistroPPLParser.g4
```diff
@@ -77,8 +77,6 @@ evalCommand
 
 headCommand
     : HEAD
-    (KEEPLAST EQUAL keeplast=booleanLiteral)?
-    (WHILE LT_PRTHS whileExpr=logicalExpression RT_PRTHS)?
```
dai-chen, line 81: remove `KEEPLAST` and `WHILE` token as well if no use?

core/src/main/java/com/amazon/opendistroforelasticsearch/sql/ast/tree/Head.java
chloe-zh: The PPL logging will not be able to show correct query since the ppl query data anonymizer is using built AST directly:
```
  @Override
  public String visitHead(Head node, String context) {
    String child = node.getChild().get(0).accept(this, context);
    Integer size = node.getSize();
    return StringUtils.format("%s | head %d", child, size);
  }
```

ppl/src/main/antlr/OpenDistroPPLParser.g4
chloe-zh: Done, thanks!

docs/experiment/ppl/cmd/head.rst
```diff
@@ -16,10 +16,8 @@ Description
 
 Syntax
 ============
-head [keeplast = (true | false)] [while "("<boolean-expression>")"] [N]
```
dai-chen, line 19: I'm thinking should we document this breaking change here or elsewhere

dai-chen (approved):
Just a minor comment. Thanks for the refactoring!

penghuo (approved):
Thanks for the change!

docs/experiment/ppl/cmd/head.rst
penghuo: Could we 
1. add the breaking change description in this PR.
2. add the breaking change label for the PR.
Then, during the release stage, 
we can pull the breaking change and describe it in the release notes?

PR #1026, opened 2021-02-02
Only keep the first element of multivalue field response.

penghuo:
*Issue #, if available:* #1017

*Description of changes:*
1. Only keep the first element of multivalue field response. https://github.com/penghuo/sql/blob/issue-1017/docs/experiment/ppl/general/datatypes.rst#example-3-selecting-field-of-array-value


elasticsearch/src/main/java/com/amazon/opendistroforelasticsearch/sql/elasticsearch/data/utils/Content.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+package com.amazon.opendistroforelasticsearch.sql.elasticsearch.data.utils;
+
+import java.util.Iterator;
+import java.util.Map;
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * Content represent data.
```
dai-chen, line 24: Could you add a little more details here? ex. abstraction for storage data ...?

dai-chen (approved):
Thanks for the changes!

chloe-zh (approved):

PR #1030, opened 2021-02-05
Workflow Fix for JDBC & Workbench

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Fix workflow credentials for JDBC and s3 bucket name for workbench

### DIFF
--- a/.github/workflows/sql-jdbc-release-workflow.yml
+++ b/.github/workflows/sql-jdbc-release-workflow.yml
@@ -24,8 +24,8 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.AWS_STAGING_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_STAGING_SECRET_ACCESS_KEY }}
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
 
       - name: Setup Java ${{ matrix.java }}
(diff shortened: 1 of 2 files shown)


.github/workflows/sql-jdbc-release-workflow.yml
```diff
@@ -37,6 +37,13 @@ jobs:
         run: |
           ./gradlew publishShadowPublicationToInternal-releasesRepository '-Dorg.gradle.jvmargs=--add-modules java.xml.bind' -Dbuild.snapshot=false
 
+      - name: Configure Staging AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
joshuali925, line 41: jdbc needs `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to build?

joshuali925 (approved):

.github/workflows/sql-jdbc-release-workflow.yml
davidcui-amzn: Currently the build is failing with an `Access denied` error, after a short working session with Sreekar from infra team he suggested using this workaround

dai-chen (approved):

joshuali925:
got it, could you also add back `cd kibana/plugins/workbench` in here above line 62? I think it was accidentally removed in staging bucket change, without `cd` the working directory is not right and `ls` doesn't work https://github.com/opendistro-for-elasticsearch/sql/blob/2f2d8c31f08f9a0a3073c4754a6755264366dec3/.github/workflows/sql-workbench-release-workflow.yml#L61-L62

davidcui-amzn:
> got it, could you also add back `cd kibana/plugins/workbench` in here above line 62? I think it was accidentally removed in staging bucket change, without `cd` the working directory is not right and `ls` doesn't work
> 
> https://github.com/opendistro-for-elasticsearch/sql/blob/2f2d8c31f08f9a0a3073c4754a6755264366dec3/.github/workflows/sql-workbench-release-workflow.yml#L61-L62

Will do, maybe that is why the workbench is failing too

Issue #1043, opened 2021-02-10
[BUG] Support empty array value for none nested field.


penghuo:
Expectation

```
PUT test/_doc/1
{
  "intv": []
}
PUT test/_doc/2
{
  "intv": [1,2,3]
}
PUT test/_doc/3
{
  "intv": [4]
}

POST _opendistro/_sql
{
  "query": """
    select * from test
  """
}

{
  "schema": [
    {
      "name": "intv",
      "type": "long"
    }
  ],
  "datarows": [
    [
      null
    ],
    [
      1
    ],
    [
      4
    ]
  ],
  "total": 3,
  "size": 3,
  "status": 200
}
```

PR #1044, opened 2021-02-10
The empty array values should be treated as null value when parsing json content

penghuo:
*Issue #, if available:* #1043

*Description of changes:*
The empty array values should be treated as null value when parsing json content


davidcui1225 (approved):
LGTM- just to confirm, this will fix the issue for the nested fields from new engine? e.g queries like `select * from employee_nested`

chloe-zh (approved):

Issue #915, opened 2020-12-11
Support cast function in aggregation function


penghuo:
E.g.
```
SELECT AVG(CAST(Cancelled AS INT)) FROM kibana_sample_data_flights
```

dai-chen:
Need to support type name `INTEGER` too.

PR #1053, opened 2021-02-17
Support type name INTEGER for CAST statement

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/915

*Description of changes:* Because MySQL spec is complex for this part, we decided to follow PostgreSQL which is clear and standard. In PostgreSQL, type name `INT` or `INT4` is just alias of `INTEGER`, so this PR is to map both to the same internal function `cast_to_int`.

*Documentation*: https://github.com/dai-chen/sql/blob/support-cast-integer/docs/user/dql/functions.rst#cast

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
have the right to submit it under the open source license
indicated in the file; or

(b) The contribution is based upon previous work that, to the best
of my knowledge, is covered under an appropriate open source
license and I have the right under that license to submit that
work with modifications, whether created in whole or in part
by me, under the same open source license (unless I am
permitted to submit under a different license), as indicated
in the file; or

(c) The contribution was provided directly to me by some other
person who certified (a), (b) or (c) and I have not modified
it.

(d) I understand and agree that this project and the contribution
are public and that a record of the contribution (including all
personal information I submit with it, including my sign-off) is
maintained indefinitely and may be redistributed consistent with
this project or the open source license(s) involved.


penghuo (approved):

chloe-zh (approved):
Np of this PR, but for the conversions between timestamp and other date and time types, I think one is missing: date -> timestamp by `timestamp()`, and the result is to attach the time of `00:00:00` and the timezone of the system default timezone to the date value. Related docs for reference: 
Our user manual: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/dql/functions.rst#timestamp 
And MySQL reference manual: https://dev.mysql.com/doc/refman/8.0/en/date-and-time-type-conversion.html (refer to the first conversion rule "Conversion of DATE values")
So could you also update the data type conversion table in the cast function in the doc? The rest looks good to me!

Issue #1058, opened 2021-02-22
Using IN ('some string') in combination to GROUP BY fails


FreCap:
Version 1.13

Given a doc
```
POST my_test_ndex2/_doc/
{
  "field1": {
    "subFieldA": 1,
    "subFieldB": 2
  },
  "field3": {
    "subFieldA": "test"
  }
}
```

This query using GROUP BY and IN with STRINGS, fails:
```
POST _opendistro/_sql
{
  "query": "select field3.subFieldA, Count(*) from my_test_ndex2 where field3.subFieldA IN ('test','test2') group by field3.subFieldA order by Count(*) desc"
}


---> Response
{
  "error": {
    "reason": "There was internal problem at backend",
    "details": "invalid value operation on MISSING_VALUE",
    "type": "IllegalStateException"
  },
  "status": 500
}

```

**Same query without group by SUCCEEDS**
```
POST _opendistro/_sql
{
  "query": "select field3.subFieldA from my_test_ndex2 where field3.subFieldA IN ('test','test2') "
}

---> Response
{
  "schema": [{
    "name": "field3.subFieldA",
    "type": "text"
  }],
  "total": 2,
  "datarows": [
    ["test"],
    ["test"]
  ],
  "size": 2,
  "status": 200
}
```

**Same query WITH group by, using IN with INTS (and not strings) SUCCEEDS**
```


POST _opendistro/_sql
{
  "query": "select field1.subFieldA, Count(*) from my_test_ndex2 where field1.subFieldA IN (1,2) group by field1.subFieldA order by Count(*) desc"
}

---> Response
{
  "schema": [
    {
      "name": "field1.subFieldA",
      "type": "long"
    },
    {
      "name": "COUNT(*)",
      "type": "integer"
    }
  ],
  "total": 1,
  "datarows": [[
    1,
    1
  ]],
  "size": 1,
  "status": 200
}
```

penghuo:
Thanks for reporting the issue.

The IN operator is not supported in our [new engine](https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/dev/NewSQLEngine.md), at runtime, it will [fallback](https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/dev/NewSQLEngine.md#32-fallback-mechanism) to the old engine.

We will prioritize this feature in new engine.

PR #1067, opened 2021-03-06
Support between/not between operator

chloe-zh:
*Issue #, if available:*
#1058 

*Description of changes:*
- Supported between / not between operator in SQL following the PostgreSQL syntax and return type (boolean): https://www.postgresqltutorial.com/postgresql-between/
- Added tests
- Update user manual: https://github.com/chloe-zh/sql/blob/between-operator/docs/user/dql/expressions.rst#between-and
- Pushdown to range query:
`(NOT) BETWEEN ... AND ...` operator is pushdown to range query when it's in a filter expression that can be pushed down. The explanation of optimized DSL is as follows:
Query `SELECT lastname, age FROM accounts WHERE age BETWEEN 20 AND 30`:
```
{
  "from":0,
  "size":200,
  "timeout":"1m",
  "query":{
    "range":{
      "age":{
        "from":20,
        "to":30,
        "include_lower":true,
        "include_upper":true,
        "boost":1.0
      }
    }
  },
  "_source":{
    "includes":[
      "age",
      "lastname"
    ],
    "excludes":[
      
    ]
  },
  "sort":[
    {
      "_doc":{
        "order":"asc"
      }
    }
  ]
}
```

And query `SELECT lastname, age FROM accounts WHERE age NOT BETWEEN 20 AND 30`:
```
{
  "from":0,
  "size":200,
  "timeout":"1m",
  "query":{
    "bool":{
      "must_not":[
        {
          "range":{
            "age":{
              "from":20,
              "to":30,
              "include_lower":true,
              "include_upper":true,
              "boost":1.0
            }
          }
        }
      ],
      "adjust_pure_negative":true,
      "boost":1.0
    }
  },
  "_source":{
    "includes":[
      "age",
      "lastname"
    ],
    "excludes":[
      
    ]
  },
  "sort":[
    {
      "_doc":{
        "order":"asc"
      }
    }
  ]
}
```

Note: this PR does not include comparison test because of different behaviors for the testing DBs, they are returning integer type (1 as true and 0 as false).


kaioken:
@chloe-zh this is perfect we are needed this specific feature for one of our components, any idea when this will get a release?

Thanks

kaioken:
@chloe-zh any update on this pull?

Issue #1076, opened 2021-03-17
SQL - Like operator behavior using curl or jdbc


Ncoati:
Executing a query using the like operator with a space, dash and possibly other characters leads to an empty resultset when using curl or a jdbc connection.

`Select * from islands where name like '%Isle %`

`>> curl -XPOST https://localhost:9200/_opendistro/_sql -u 'admin:admin' -k -H 'Content-Type: application/json' -d
{"query":"Select * from islands where name like ?", "parameters": [{"value":"%Isle %", "type":"keyword"}], "fetch_size":0 }
`

Executing the query using the Query Workbench is returning the correct results.

I am using the latest versions available.

dai-chen:
@Ncoati Thanks for reporting the issue! Could you share your index data for our testing? And you mean your LIKE query worked in Query Workbench but returned nothing using curl/JDBC driver?

Ncoati:
Here is some sample data:
`curl -XPUT https://localhost:9200/islands/_doc/1 -u 'admin:admin' -k -H 'Content-Type: application/json' -d '{"name":"Isle of Man"}'`
`curl -XPUT https://localhost:9200/islands/_doc/2 -u 'admin:admin' -k -H 'Content-Type: application/json' -d '{"name":"Galapagos Islands"}'`
`curl -XPUT https://localhost:9200/islands/_doc/3 -u 'admin:admin' -k -H 'Content-Type: application/json' -d '{"name":"Tasmania"}'`
`curl -XPUT https://localhost:9200/islands/_doc/4 -u 'admin:admin' -k -H 'Content-Type: application/json' -d '{"name":"Test-Island"}'`

Yes, you are correct. Executing the following query returns no data:
`curl -XPOST https://localhost:9200/_opendistro/_sql -u 'admin:admin' -k -H 'Content-Type: application/json' -d '{"query":"Select * from islands where name like ? ","parameters":[{"value":"%Isle %","type":"keyword"}],"fetch_size":0}'`

As soon as there is a space or dash in the parameter value, no data is being returned.

dai-chen:
Thanks for the info! I can reproduce the issue from my side:

```
PUT islands/_doc/1
{"name":"Isle of Man"}

POST _opendistro/_sql/_explain
{
  "query": "SELECT * FROM islands WHERE name LIKE ? ",
  "parameters":[{"value":"%Isle %","type":"keyword"}]
}
{
  "from" : 0,
  "size" : 200,
[...]

Issue #1070, opened 2021-03-09
Using GROUP BY in JDBC with new sql engine failed 


boicehuang:
Version 1.13

As mentioned in https://github.com/opendistro-for-elasticsearch/sql/issues/401,   query using GROUB BY is correct for NOT JSON formatting POST _opendistro/_sql/?sql= by using new SQL engine,  but still failed for JDBC query.

1. Open the new SQL engine: 
```
PUT /_opendistro/_sql/settings { "transient" : { "opendistro.sql.engine.new.enabled" : "true" } }
```

2. This query using GROUP BY with POST  succeed:
```
POST localhost:9200/_opendistro/_sql
{
    "query" : "SELECT title FROM movies GROUP BY title"
}
```

3. But it still failed for JDBC query 

java code
```
public static void queryTest() throws Exception {
        Class.forName("com.amazon.opendistroforelasticsearch.jdbc.Driver");
        String url = "jdbc:elasticsearch://localhost:9200";
        Properties properties = new Properties();
        String sql = "SELECT title FROM movies GROUP BY title";
        try (Connection connection = DriverManager.getConnection(url, properties);
             PreparedStatement statement = connection.prepareStatement(sql)) {
            ResultSet rs = statement.executeQuery();
            ResultSetMetaData metaData = rs.getMetaData();
            int numCols = metaData.getColumnCount();
            int numRows = 0;
            while (rs.next()) {
                System.out.print(++numRows + ":");
                for (int i = 1; i <= numCols; i++) {
                    System.out.print(metaData.getColumnLabel(i) + ":" + rs.getString(i) + "\t");
                }
                System.out.println();
            }
        }
    }
```
response:  
```
Exception in thread "main" Internal Server Error. Reason: There was internal problem at backend. Type: IllegalStateException. Details: invalid value operation on MISSING_VALUE
        at com.amazon.opendistroforelasticsearch.jdbc.StatementImpl.executeQueryRequest(StatementImpl.java:80)
        at com.amazon.opendistroforelasticsearch.jdbc.PreparedStatementImpl.executeQueryX(PreparedStatementImpl.java:87)
        at com.amazon.opendistroforelasticsearch.jdbc.PreparedStatementImpl.executeQuery(PreparedStatementImpl.java:78)
        at Main.queryTest(Main.java:41)
        at Main.main(Main.java:28)
```
error stack in es logs:
```
[...]

PR #1077, opened 2021-03-19
Enable prepared statement support in new engine

dai-chen:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/sql/issues/1076, https://github.com/opendistro-for-elasticsearch/sql/issues/1070

*Description of changes:* Currently the new SQL engine won't handle request with `parameters` field which is mostly from JDBC `PreparedStatement`. Actually parameters are parsed and substituted before hitting new engine. So this PR is just to change fallback criteria a little bit and avoid fallback on request with `parameters` field.

*Documentation*: https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/user/interfaces/protocol.rst#example-2

*Testing*: Added new IT and UT.


chloe-zh (approved):
LGTM, thanks for the fix!

penghuo (approved):

